Issue #8, opened 2024-06-03
Design Doc for SDG


aakankshaduggal:
Add a design doc for the proposed SDG library explaining the motivation, parameters, import calls and architectural design.

russellb:
https://github.com/instructlab/dev-docs/blob/main/docs/sdg/sdg-api-interface.md

PR #9, opened 2024-06-03
Create sdg_design.md

aakankshaduggal:
Addresses #8

### DIFF
--- a/docs/sdg_design.md
+++ b/docs/sdg_design.md
@@ -1,0 +1,99 @@
+# Instructlab SDG Library Design
+
+(This is a draft, seeking reviews and architectural consensus on the design of the ilab sdg library)
+
+## SDG Library
+
+Library called `sdg` that can be called per q&a pair from various clients.
+
+## Motivation
+
+The motivation behind having a specifc SDG library is to focus the functionality around synthetic data generation.
+
+Various modules of InstructLab like the CLI (`ilab`), backend orchestrator, and other internal services access the SDG tool but the expectations are different from each interface.
+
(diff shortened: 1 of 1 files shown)


docs/sdg_design.md
```diff
@@ -0,0 +1,99 @@
+# Instructlab SDG Library Design
+
+(This is a draft, seeking reviews and architectural consensus on the design of the ilab sdg library)
```
russellb, line 3: Thanks for posting this!

I know previously we were thinking about a different starting point. Could you update this doc to reflect the current starting point of the code pulled out of `instructlab/instructlab` ?  It currently exposes a single `generate_data()` function. Are there steps you would envision taking to evolve from that code to what you envision here? 

Are there any significant challenges you expect to need to overcome with this starting point?

Hopefully what I'm saying makes sense!

aakankshaduggal: Thanks @russellb. This absolutely makes sense. I was planning to capture that in issues instead of this document, because there are a lot of small changes that we will have to make on the way. 

But if you suggest, will add those details in here itself

russellb: I just think it would be helpful to put the proposal in context of what we're working from. Clarify the problems the updated design solves with what we're starting from - that sort of thing

russellb:
@aakankshaduggal what do you think of moving this over to the `dev-docs` repository?

We could create a new `docs/sdg` directory to hold design docs associated with this repo.

aakankshaduggal:
> @aakankshaduggal what do you think of moving this over to the `dev-docs` repository?
> 
> We could create a new `docs/sdg` directory to hold design docs associated with this repo.

Sure, will close this for now.

Issue #6, opened 2024-06-02
Remove imports of code from `instructlab`


russellb:
Since this code was pulled from the `instructlab` repo, it still imports some code from there. We do not want a bidirectional dependency between the packages. We need to remove all `instructlab` imports from here.

russellb:
closing since #11 is the same and has more comments

PR #7, opened 2024-06-03
Implement initial CI integrations

nathan-weinberg:
Resolves #4 

This PR implements the following CI integrations:
- **Dependabot** via a GitHub Action configuration
- **Markdown Linting** implemented via `Makefile` as well as a Hardened GitHub Action
- **Spellchecking** implemented via `tox` (with a `Makefile` entrypoint) as well as a Hardened GitHub Action
- **Action Linting** implemented via `Makefile` as well as a Hardened GitHub Action
- **Python Linting and Formatting** via `tox` using `ruff` and `isort` (with a `Makefile` entrypoint) as well as a Hardened GitHub Action

It also addresses Markdown and Python linting errors that are now being detected

PR #12, opened 2024-06-04
ci: add action for python linting and formatting

nathan-weinberg:
Implemented this locally in #7 but missed this action


nathan-weinberg:
Only CI failure here is due to #6

requirements.txt
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+click>=8.1.7,<9.0.0
+httpx>=0.25.0,<1.0.0
+jinja2
+openai>=1.13.3,<2.0.0
+rouge_score
+tqdm>=4.66.2,<5.0.0
```
russellb, line 7: can you add `instructlab` here?  until we resolve #6

bjhargrave:
We should also add the `mypy` step from https://github.com/instructlab/instructlab/pull/1235 once it is completed.

tiran:
You also need to add a `py.typed` marker file to `src/instructlab/sdg`, so the parent package knows that the subpackage uses type annotations. https://peps.python.org/pep-0561/#packaging-type-information

requirements.txt
nathan-weinberg: Yup, added that and a note to remove when the issue is resolved

nathan-weinberg:
> We should also add the `mypy` step from [instructlab/instructlab#1235](https://github.com/instructlab/instructlab/pull/1235) once it is completed.

As part of this PR or as a followup?

> You also need to add a `py.typed` marker file to `src/instructlab/sdg`, so the parent package knows that the subpackage uses type annotations. https://peps.python.org/pep-0561/#packaging-type-information

That path doesn't exist - do you mean `src/instructlab_sdg`? What should this file look like?

nathan-weinberg:
@bjhargrave since the MyPy PR merged in the other repo I went ahead and added it to this PR

bjhargrave (approved):

russellb (approved):

Issue #2, opened 2024-06-01
Reserve package name on PyPI


russellb:
See title.  Register the library on PyPI. We have to publish a package, even if it's empty at first.

- `sdg` is taken
- `instructlab_sdg` could work, and roughly lines up with the repo `instructlab/sdg`
- `ilabsdg` seems a lot nicer to type. We could rename the repo to that, too, so it matches.

Another option from @tiran 

> here is another option. We could turn instructlab into a namespace package. That would allow us to distribute `instructlab.sdg` subpackage as a separate Python package that can be installed with pip install `instructlab.sdg`.
> 
> https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

russellb:
@tiran I'm interested in your feedback on the package name. There's more context here: https://github.com/instructlab/dev-docs/blob/main/docs/sdg-repo.md

tiran:
@russellb There is another option. We could turn `instructlab` into a namespace package. That would allow us to distribute 'instructlab.sdg` subpackage as a separate Python package that can be installed with `pip install instructlab.sdg`. 

 https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

russellb:
> @russellb There is another option. We could turn `instructlab` into a namespace package. That would allow us to distribute `instructlab.sdg` subpackage as a separate Python package that can be installed with `pip install instructlab.sdg`.
> 
> https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

Thanks! I added that to the list of options at the top.

I don't really have a strong opinion. I just want to do something following good Python practices.

dhellmann:
I like the namespace package idea, if the things are coupled in a way that you'd only use them together. If you would use instructlab.sdg without instructlab then I'm not sure I'd go that route.

tiran:
Either option is fine.

Namespace packages require a tiny boiler plate and some coordination between packages. They are a great fit for an organization that controls a top level namespace. The concept is similar to a top level domain in DNS. The concept has been around for 15, 20 years and is used by e.g. Zope and Plone. IMHO it would be a better option for us.

dhellmann:
> I didn’t mean to close this.
> 
> I think the namespace package idea sounds good. I think we need to change the current instructlab package first?

Yes, you'll need to ensure that everything is delivered in a subdirectory of the `instructlab` directory.

[...]

PR #14, opened 2024-06-04
Rename package to `instructlab.sdg` from `instructlab_sdg`

russellb:
Part of issue #2

This changes to use Python namespace packages where `instructlab` is
the namespace used for our libraries.

---

- [x] Depends on https://github.com/instructlab/instructlab/pull/1263


russellb:
The instructlab side went in, so this can go in now once I rebase it.

russellb:
I think I'm down to a linter fight between ruff wanting this import to move to its own first party section

```
+# First Party
+from instructlab.sdg import utils
```

but if I do that, there's another linter complaining that all of the `instructlab` imports aren't grouped together

src/instructlab/__init__.py
```diff
@@ -0,0 +1 @@
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```
bjhargrave, line 1: Is this __init__.py needed here? 

> This will allow some portions of a namespace to be legacy portions while others are migrated to PEP 420.

https://peps.python.org/pep-0420/#migrating-from-legacy-namespace-packages 

indicates (to me) that it is not necessary in the parts of the namespace (like this and `instructlab.schema`) to have an __init__.py.

alimaredia (approved):

hickeyma (commented):
Looks good, waiting on gate to be fixed

russellb:
I fixed the linter issue.

nathan-weinberg (approved):

PR #20, opened 2024-06-13
Drop generate_data function import

russellb:
This is confusing some tests in the CLI repo and is not necessary.


russellb:
This is a change I introduced. It's not from the original code. I'm just going to merge it once it passes CI so I can continue pushing through the first PR for the CLI to consume this library.

PR #21, opened 2024-06-13
Wrap read_taxonomy from instructlab.utils

russellb:
This is a hack to get tests passing back in `instructlab/instructlab`.
There is a test there that tries to mock read_taxonomy, but it's not
working because of how we use it. This hack gives a place the test can
mock.

The real issue here is that the split of code between instructlab and
sdg is a mess and these dependencies in both directions both in code
and in the tests. There's a lot of work to clean this up, but I'm just
trying to get to a functional starting point.


russellb:
Trying to get this done quickly so I'm just merging ... ping me if you see anything I broke

PR #26, opened 2024-06-17
Specify versions for all dependencies

russellb:
Two packages in requirements.txt did not specify versions in a way
consistent with the rest of file. Now they do.


ktam3:
@russellb - is this a GA blocker?

russellb:
this can be closed -- neither requirement is used anymore

russellb:
related follow-up #152

PR #28, opened 2024-06-19
requirements.txt: Move instructlab to requirements-dev.txt

russellb:
70335a1 requirements.txt: instructlab to requirements-dev.txt

commit 70335a140f9036a135fa228165c4964695a37912
Date:   Tue Jun 18 21:26:12 2024 -0400

    requirements.txt: instructlab to requirements-dev.txt
    
    This is still a dependency, but it's causing some problems in
    `instructlab` repo CI. A released version of instructlab gets
    installed first because of this entry, then it has to be uninstalled
    before the development version can be installed.
    
    The currently released version of `instructlab` still has schema
    embedded. The next release (and current `main`) uses the
    `instructlab-schema` package. Moving from one to the other in tox
    is breaking the schema package.
    
    By moving it, it will still be installed for local repo testing
    purposes.
    
    This should solve the CI failures seen here:
    https://github.com/instructlab/instructlab/pull/1244
    
### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
-# TODO: Uncomment below line once requirements.txt is created
 -r requirements.txt
 
+# TODO: remove 'instructlab' once https://github.com/instructlab/sdg/issues/6 is resolved
+instructlab>=0.17.0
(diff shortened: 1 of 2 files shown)


russellb:
This is a trivial change that's blocking the CLI change to use the library, so I'm just going to merge it.

Once we get to a working initial state, I won't do merges like this.

PR #33, opened 2024-06-24
Add e2e test workflow

russellb:
This workflow runs an end-to-end test workflow using `ilab` based on
the latest instructlab-sdg code. This is based on the e2e workflow in
the instructlab repo.

It currently runs against whatever is in `main` for `instructlab`.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,3 @@
+self-hosted-runner:
+  labels:
+  - ubuntu-gpu
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E test
+
+on:
+  push:
+    branches:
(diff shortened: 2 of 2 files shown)


nathan-weinberg (approved):

.github/actionlint.yaml
```diff
@@ -0,0 +1,3 @@
+self-hosted-runner:
```
bjhargrave, line 1: Why is this file named "actionlint" rather than "e2e"?

russellb: because it's a config file for the actionlint workflow. This is needed so that actionlint considers `ubuntu-gpu` a valid runner type

russellb: it's a copy of the same file in `instructlab/instructlab`

bjhargrave: Ah! I did not make that connection. Silly me.

bjhargrave (approved):

russellb:
Thanks for the reviews!

Issue #11, opened 2024-06-04
Import functions from the ilab repository


aakankshaduggal:
Some functions that are being called in the `generate_data.py` file are in this file - https://github.com/instructlab/instructlab/blob/main/src/instructlab/utils.py

---

to-do list:

- [x] `DEFAULT_MULTIPROCESSING_START_METHOD`
  - Used in `lab.py` but also this is just a variable
- [x] `DEFAULT_API_KEY`
  - Used in many places but also this is just a variable
- [x] `DEFAULT_MODEL_OLD`
  - Used in `lab.py` but also this is just a variable
- [x] `get_model_family()`
  - Used by `generate_data.py` but also `server.py`
  - https://github.com/instructlab/sdg/pull/46

**From utils, we import the following:**
- [x] `chunk_document()`
  - Only used by `generate_data.py`
  - #35 
- [x] `max_seed_example_tokens()`
  - Only used by `generate_data.py`
  - #35 
- [x] `num_chars_from_tokens()`
  - Only used by `generate_data.py`
  - #35 
- [x] `read_taxonomy()`
  - Used by `generate_data.py` but also `lab.py`
  - https://github.com/instructlab/sdg/pull/46
- [ ] `get_sysprompt()`
  - Used by `generate_data.py` but also `lab.py`, `chat.py` and `make_data.py`
  - https://github.com/instructlab/sdg/pull/89

nathan-weinberg:
@aakankshaduggal @russellb is this something we want to try and preserve commits for? if not I'm happy to take this (and the other issue I guess, seems one PR can resolve both)

nathan-weinberg:
ack @russellb - fine either way - AFAIK the code in question is only used in the CLI currently

russellb:
> ack @russellb - fine either way - AFAIK the code in question is only used in the CLI currently

question is it called from only code other than what got moved here

nathan-weinberg:
Okay looked into this a bit - we import from two locations within `instructlab` - [config](https://github.com/instructlab/instructlab/blob/main/src/instructlab/config.py) and [utils](https://github.com/instructlab/instructlab/blob/main/src/instructlab/utils.py)

**From config, we import the following:**
- `DEFAULT_MULTIPROCESSING_START_METHOD`
  - Used in `lab.py` but also this is just a variable
- `DEFAULT_API_KEY`
  - Used in many places but also this is just a variable
- `DEFAULT_MODEL_OLD`
[...]

PR #35, opened 2024-06-24
Move some code from instructlab.utils

russellb:
This code was only used by instructlab.sdg, so move it over here
instead of leaving it back in the `instructlab` repo.

Part of issue #11


cdoern (approved):
lgtm

PR #37, opened 2024-06-25
reduce number of arguments in openai_completion and openai_completion

makelinux:
Several functions passes trough six argument verbatim to OpenAI client creation.
It causes to very long argument lists and increases technological debt.

Pass OpenAI client object instead of those six arguments to reduce
duplicated code and increase repeatability.

TODO: perform the same optimization with generate_data and cli too.

Signed-off-by: Costa Shulyupin <costa.shul@redhat.com>


markmc:
Thanks for your contribution. However, I think now that #46 has merged, this code is no longer present?

PR #38, opened 2024-06-25
Restrict version of tenacity

russellb:
A bug was discovered when the related code was in the CLI repo. We
need to restrict the version of `tenacity` pulled in by
`langhchain-text-splitters`.


russellb:
This is a trivial dependency update that has no functional impact in practice since it's identical to what `instructlab` has in its requirements.txt, so I'm going to merge

Issue #41, opened 2024-06-25
Update the sdg library 


aakankshaduggal:
Add the new design and code for the sdg library

russellb:
bulk of the new library code is merged!

PR #42, opened 2024-06-25
Add SDG library code

aakankshaduggal:
This PR adds the code for the SDG library #41 

In order to test the code, you can run the `tests/test_knowledge.py`, specify the model endpoint and run the file from the root like such - `python3 -m tests.test_knowledge.py`

---

Corresponding design doc: https://github.com/instructlab/dev-docs/blob/main/docs/sdg/sdg-api-interface.md


russellb:
> This PR adds the code for the SDG library #41
> 
> In order to test the code, you can run the `tests/test_knowledge.py`, specify the model endpoint and run the file from the root like such - `python3 -m tests.test_knowledge.py`

You'll need to move the test script to a different directory. It's getting picked up by the automation that runs unit tests. If you move it, it shouldn't run.

scripts/test_knowledge.py
```diff
@@ -0,0 +1,51 @@
...
+
+samples = [
+    {
+        "question_1": "what is the location of the tubal tonsils?",
+        "response_1": "The location of the tubal tonsils is the roof of the pharynx.",
+        "question_2": "How long does the adenoid grow?",
+        "task_description": "Teaching about human anatomy, specifically tonsils",
+        "response_2": "The adenoid grows until the age of 5, starts to shrink at the age of 7 and becomes small in adulthood.",
+        "question_3": "What is the immune systems first line of defense against ingested or inhaled foreign pathogens?",
+        "response_3": "The tonsils are the immune systems first line of defense.",
```
russellb, line 34: Why is it numbered like this instead of being a list of q/a pairs?

shivchander: This depends on the prompt template. Essentially not all the seed questions provided in the taxonomy yaml goes in at a time. We will be using a sliding window (of size 3) approach to select the in context questions, then keep moving the window until it covers all the questions provided in the taxonomy yaml

oindrillac: since that was the easiest way to format it into a prompt template, all the keys are directly injected into the prompt for consistency for example see below
```
For this {domain} domain here are some sample questions:
[Question]
{question_1}
[Answer]
{answer_1}
[End]
```

russellb: got it -- so the sliding window bit needs to be handled in `generate_data()` then?  that works

shivchander: yep

src/instructlab/sdg/__init__.py
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: Apache-2.0
+# Third Party
+from datasets import Dataset
+
+# Local
+from .pipeline import Pipeline
+
+
+class SDG:
```
RobotSail, line 9: Should this be in the `__init__.py` file?

src/instructlab/sdg/llmblock.py
```diff
@@ -0,0 +1,107 @@
...
+                pattern = re.escape(start_tag) + r"(.*?)" + re.escape(end_tag)
+                all_matches = re.findall(pattern, generated_string, re.DOTALL)
+                matches[output_col] = (
+                    [match.strip() for match in all_matches] if all_matches else None
+                )
+
+        return matches
+
+    def _generate(self, samples, **gen_kwargs) -> list:
+        prompts = [
```
RobotSail, line 64: You can leave this as-is but in the future I recommend using a simpler for loop vs. list comprehension here so that it's simpler to read, e.g.:

```py
prompts = []
for sample in samples:
  prompt = self.prompt_template.format(**sample).strip()
  prompts.append(self.model_prompt.format(prompt))
```
Is easier to read than 

```py
        prompts = [
            self.model_prompt.format(
                prompt=self.prompt_template.format(**sample).strip()
            )
            for sample in samples
```

RobotSail (approved):
One comment but otherwise LGTM

src/instructlab/sdg/__init__.py
shivchander: Do you have a suggestion for refactor? didnt want to create another sdg class

russellb: ideally not ... i'd probably put it in an sdg.py and then import it here

russellb: though I'll probably put that off for future changes

RobotSail: @shivchander My recommendation is to just have a `classes.py` file and then in this file you'd just add a `from .classes import SDG`

src/instructlab/sdg/llmblock.py
shivchander: I buy your point on readability, but I think list comprehensions are better in terms of performance

scripts/test_knowledge.py
```diff
@@ -0,0 +1,51 @@
...
+openai_api_base = "Add model endpoint here"
+
+client = OpenAI(
+    api_key=openai_api_key,
+    base_url=openai_api_base,
+)
+
+models = client.models.list()
+teacher_model = models.data[0].id
+router_model = models.data[1].id
```
npalaska, line 24: What is the router model? can it be the same as the teacher model? Also if they are the same then the index need to be `0`

russellb:
As of right now, none of this new code is being used, so it will not break the CLI (as confirmed by the e2e CI job).

Since this is so big, we're going to merge and keep working on it in tree.

PR #44, opened 2024-06-25
Update e2e config to optimize pip caching

nathan-weinberg:
Based on convo in https://github.com/instructlab/eval/pull/14


russellb (approved):
lgtm assuming it still passes CI

PR #47, opened 2024-06-26
📚 Adding Knowledge blocks

abhi1092:
* 🧩 Conditional Block for conditionally picking prompt templates and parsing function
* 🎛️ Changes in the Sampling block to handle post_fixes
* 🐞 Bug fix for _parse function in LLMBlock


aakankshaduggal (approved):
Thanks @abhi1092 for the quick fix!

PR #51, opened 2024-06-28
e2e: Fix permissions error

russellb:
Something changed in the underlying image where this command now
caused a failure on a permissions error. The output was helpful at
some point during debugging so I'm leaving it in place, but running it
with sudo to avoid the permissions error.


russellb:
This change fixed the same error in `instructlab/instructlab` and it's a trivial change, so I'm just going to merge it to unblock running this job on other PRs

PR #53, opened 2024-06-28
updates to grounded flow

oindrillac:
Stress testing was done on the SDG pipeline
* added grounded test script that can be used to test
* updates to templates to get full pipeline working

### DIFF
--- a/scripts/test_grounded_skills.py
+++ b/scripts/test_grounded_skills.py
@@ -1,0 +1,107 @@
+# Third Party
+from datasets import Dataset
+from openai import OpenAI
+
+# First Party
+from src.instructlab.sdg import SDG
+from src.instructlab.sdg.default_flows import SynthGroundedSkillsFlow
+from src.instructlab.sdg.pipeline import Pipeline
+
+# for vLLM endpoints, the api_key remains "EMPTY"
+openai_api_key = "EMPTY"
+openai_api_base = "Add model endpoint here"
+
+
+client = OpenAI(
+    api_key=openai_api_key,
+    base_url=openai_api_base,
+)
+
+models = client.models.list()
+teacher_model = models.data[0].id
+
+samples = [
+    {
+        'seed_context': """*Ms. Thompson:* Good morning, everyone. Today, we''re here to discuss
+    our customer journey mapping and analysis. I believe this is crucial to understanding
+    our customers'' experiences and improving our services.
+
+
+    *Mr. Patel:* I agree, Lisa. We should start by identifying all touchpoints in
+    our customer journey, from initial contact to post-sale support.
+
+
+    *Ms. Rodriguez:* Yes, and let''s not forget about the emotional aspect of the
+    journey. How do our customers feel at each stage? What are their pain points?
+
+
+    *Mr. Kim:* We can use data from our CRM system to track the customer journey and
+    gather insights. This will help us create a detailed, data-driven map.
+
+
+    *Ms. Johnson:* Once we have the map, we can analyze it to identify areas for improvement.
+    Perhaps there are steps where customers drop off or express dissatisfaction.
+
+
+    *Mr. Davis:* We should also consider the customer''s perspective. Conducting interviews
+    or surveys can provide valuable insights into their thoughts and feelings.
+
+
+    *Ms. Brown:* Absolutely. And once we''ve identified areas for improvement, we
+    can develop strategies to address them. This might involve redesigning certain
+    touchpoints, enhancing our communication, or streamlining processes.
+
+
+    *Mr. Smith:* And we must remember to measure the impact of any changes we make.
+    By tracking key performance indicators, we can determine whether our efforts are
+    successful.
+
+
+    *Ms. Thompson:* Great suggestions, everyone. Let''s divide into teams to tackle
+    different aspects of the customer journey. Team 1 will focus on pre-sale touchpoints,
+    Team 2 on sales, and Team 3 on post-sale support.
+
+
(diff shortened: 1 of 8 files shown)


scripts/test_grounded_skills.py
```diff
@@ -0,0 +1,107 @@
...
+    \ I look forward to our continued collaboration as we work towards enhancing our\
+    \ services and delivering exceptional customer experiences.\n\nBest regards,\n\
+    \n[Your Full Name]\n[Your Position]\n[Company Name]""",
+    }
+]
+
+
+ds = Dataset.from_list(samples)
+
+skills_flow = SynthGroundedSkillsFlow(client, teacher_model).get_flow()
```
npalaska, line 100: The `Flow` class expect `client, model_family, model_id, num_iters`. You might want to change this to 
```
skills_flow = SynthGroundedSkillsFlow(client, "mixtral", teacher_model, 1).get_flow()
```
Same goes for other `test_*` files

src/instructlab/sdg/default_flows.py
```diff
@@ -225,8 +226,9 @@ def get_flow(self) -> list:
                 "block_config": {
                     "block_name": "filter_relevancy",
                     "filter_column": "score",
-                    "filter_value": "2",
+                    "filter_value": "2.0",
                     "operation": operator.eq,
+                    "convert_dtype": float,
```
npalaska, line 231: This change is somehow breaking the knowledge generation. The `filter_relevancy` block is returning a dataset with 0 rows (probably because it filters out all the responses generated by the previous block). Is this expected? Same goes for `filter_verify_question` block below.

npalaska: Ah this could be because the file `src/instructlab/sdg/configs/knowledge/simple_generate_qa.yaml` still contains 

```
  {question_1}
  {response_1}

  {question_2}
  {response_2}

  {question_3}
  {response_3}
  
```
instead of 

```
  {icl_query_1}
  {icl_response_1}
 
  {icl_query_2}
  {icl_response_2}

  {icl_query_3}
  {icl_response_3}
```

markmc:
The commit log could be much improved. The logical changes I see, with the explanations I'd expect:

1. Convert filter values to a float - does this fix a bug? details of how the bug manifested would help others learn
2. Combine question and context in the grounded skills flow at the end - why? what's the before and after behavior
3. template updates to get full pipeline working - these look like important, but subtle changes, what details can you provide to help a person in future understand what to look out for if they are making further changes?

scripts/test_grounded_skills.py
oindrillac: that should be fixed, thanks for the catch

oindrillac:
here are some explanations for the changes:
> 1. Convert filter values to a float - does this fix a bug? details of how the bug manifested would help others learn

This is addressing an issue that we observed. In some of the cases where the model is being asked to provide a total score, it provides floating point sums like 1.5/2.5 and to accommodate those cases where we want to filter those values out it made sense to keep it float.

> 2. Combine question and context in the grounded skills flow at the end - why? what's the before and after behavior

That is a step towards converting it to the final format (messages) required for training.

> 3. template updates to get full pipeline working - these look like important, but subtle changes, what details can you provide to help a person in future understand what to look out for if they are making further changes?

The template changes stem from model behavior on the templates and stress testing across various leaf nodes. The changes were mainly made to make sure the responses from the model align best with the expected behavior

scripts/test_grounded_skills.py
oindrillac: fixed, PTAL!

markmc:
> here are some explanations for the changes:

It would be great to get those explanations into the commit messages. See also "Information in commit messages" in https://www.berrange.com/tags/commit-message/

> The commit message must contain all the information required to fully understand & review the patch for correctness. Less is not more. More is more.

shivchander:
> > here are some explanations for the changes:
> 
> It would be great to get those explanations into the commit messages. See also "Information in commit messages" in https://www.berrange.com/tags/commit-message/
> 
> > The commit message must contain all the information required to fully understand & review the patch for correctness. Less is not more. More is more.

Thanks for the recommendation! We could start doing this here on, this PR needs to go in by today. We are on a very tight deadline

shivchander (commented):
a few minor changes, otherwise lgtm

oindrillac:
> a few minor changes, otherwise lgtm

done, thank you!

aakankshaduggal (approved):
LGTM, thanks @oindrillac

shivchander (approved):
LGTM, thanks

npalaska (commented):
LGTM

russellb (changes requested):
the last two commits just say "updated to string" but neither are updating something to be a string.

either way it looks like something that should be squashed in to your main commit

oindrillac:
@russellb squashed

russellb:
> @russellb squashed

Thanks. Can you also add your explanations of the changes that you gave here to the commit messages as suggested?

oindrillac:
@russellb @markmc updated the commit messages as per your suggestion, ptal, thanks https://github.com/instructlab/sdg/pull/53

russellb (approved):
I see that e2e passed, though the commit messages still seem the same?  I'm happy for it to merge once that content is there.  Let me know if any help is needed with that. You don't need to wait for e2e to pass again if you only change the commit message.  Just merge away.

russellb:
lgtm!

oindrillac:
Oh for some reason it didn't get synced. Thanks for adding descriptive notes @aakankshaduggal

PR #50, opened 2024-06-27
📚 Adding Knowledge llm blocks

abhi1092:
- 🧩 Conditional Block for conditionally picking prompt templates and parsing function
- 🎛️ Changes in the Sampling block to handle post_fixes
- 🐞 Bug fix for _parse function in LLMBlock

PR #57, opened 2024-06-30
Fix dataset formatting for pipeline differences

### LINKED ISSUE #52: Ensure "simple" pipeline does not regress knowledge document chunking behavior
In the conversion to use new APIs, we added a "simple" pipeline intended to work on small environments (laptops, etc). The document chunking behavior was left as a TODO item and is probably a regression. It needs to be revisited before releasing a new version of the library.

https://github.com/instructlab/sdg/blob/1f71fb67aa46151bc0362e733b06529a1c609e6d/src/instructlab/sdg/generate_data.py#L293-L301

Follow-up to #46

### LINKED ISSUE #55: Fix dataset formats to match each pipeline expectations
`generate_data.py` calls `leaf_node_to_samples()`, which right now assumes that the expected dataset format is the same for all pipeline types. It matched the simple pipelines and the full knowledge pipeline. I didn't notice that the full skills pipeline was different. I saw that #50 changed the name of fields expected in the template, so now we have 3 different dataset formats expected by the default pipelines.

simple pipelines expect `question_N`, `answer_N` for `N` in `(1, 2, 3)`
full knowledge pipeline now expects `icl_query_N` / `icl_reponse_N`for `N` in `(1, 2, 3)`
full skills pipelines expect `seed_question` and `seed_response` (just 1 instead of 3)

Short term fix:

1. Document the dataset format expected by each pipeline.
2. fix the code to generate the dataset in the expected format for each pipeline.

Longer term consideration:

None of this matches the source format for the dataset (taxonomy). Allowing people to specify a custom pipeline implies specifying their expected sample dataset format somehow.

Another idea instead ...

1. Always assume a consistent dataset format.
2. Add a new pipeline capability for dataset transformation -- rename fields if you want, squash the rows into groups of 3 seed questions/answers per row (for the knowledge case)

I think something like this is going to be necessary to allow more configurable custom pipelines, as we'll need a way for a custom pipeline to declare the dataset format it is expecting from a known starting point.

russellb:
This PR

- re-introduces document chunk handling for knowledge docs
- fixes a mismatch between pipeline expectations and the formatting of the
  dataset of seed examples.

Closes #52
Closes #55

8112123 Re-introduce document chunking for knowledge
15ae2b9 Change question/response to icl_query/icl_response
bba13d3 Create a sample per seed example for skills

commit 811212353016e0307ef1399f6dd50ec60e44f09d
Date:   Sun Jun 30 14:02:12 2024 -0400

    Re-introduce document chunking for knowledge
    
    When generating samples for a knowledge pipeline, we have to chunk the
    document down to a size that will fit within the model's context size.
    There was a hack in place that only used a single chunk. The code now
    iterates over all chunks of the document for creating samples to send
    through the pipeline.
    
    The commit also separates the code for knowledge and skills since the
    differences between the formats is growing.
    
    Closes #52
    
commit 15ae2b947369716a27e26ee8cd138e79fce4f45a
Date:   Sun Jun 30 14:08:39 2024 -0400

    Change question/response to icl_query/icl_response
    
    PR #50 changed the format used in the full knowledge pipeline. Change
    the simple pipelines to match.
    
    Part of issue #55.
    
commit bba13d3ae7eb96846ef8aa830a787d49aa693b55
Date:   Sun Jun 30 14:25:25 2024 -0400

    Create a sample per seed example for skills
    
    The full skills pipelines expect a single seed question and response
    in each sample in the dataset. Change the simple skills pipelines to
    match and update the code to generate the samples in the expected
    format.
    
    Closes #55 (the short term needs at least)
    
### DIFF
--- a/src/instructlab/sdg/utils/taxonomy.py
+++ b/src/instructlab/sdg/utils/taxonomy.py
@@ -19,6 +19,7 @@ import yaml
 
 # First Party
 from instructlab.sdg import utils
+from instructlab.sdg.utils import chunking
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


markmc:
Looks good to me as a short term fix

It does all seem arbitrarily divergent from the taxonomy format though (as you say in #52) - e.g. icl vs seed, query/response vs question/answer

Since the taxonomy format is basically 'seed_examples: [context, question, answer]' I think settling on `seed_{context,question,answer}(_N)` for both knowledge and skills would be a nice cleanup, even in the short-term

russellb:
> Looks good to me as a short term fix
> 
> It does all seem arbitrarily divergent from the taxonomy format though (as you say in #52) - e.g. icl vs seed, query/response vs question/answer
> 
> Since the taxonomy format is basically `seed_examples: [context, question, answer]` I think settling on `seed_{context,question,answer}(_N)` for both knowledge and skills would be a nice cleanup, even in the short-term

Totally agree. I was thinking of splitting that part out of #52 into a new issue to handle as another PR.

russellb:
I filed #59 to follow-up on naming and structure of the data

src/instructlab/sdg/utils/taxonomy.py
```diff
@@ -415,42 +416,81 @@ def read_taxonomy_leaf_nodes(taxonomy, taxonomy_base, yaml_rules):
     return leaf_nodes
 
 
-def leaf_node_to_samples(leaf_node):
+def _knowledge_leaf_node_to_samples(leaf_node, server_ctx_size, chunk_word_count):
     samples = [{}]
 
+    # document is the same for the whole leaf node
```
aakankshaduggal, line 422: Need a bit of clarification here --
Are we expecting just one document at a time? Because in a leaf node, we could have multiple documents as well.

russellb: `chunk_document()` handles multiple documents.  take a look here: https://github.com/instructlab/sdg/blob/1f71fb67aa46151bc0362e733b06529a1c609e6d/src/instructlab/sdg/utils/chunking.py#L46-L48

russellb: @aakankshaduggal i'm going too hold off merging until you confirm this makes sense to you -- code may not be super clear, but I do think it's handling multiple documents properly

aakankshaduggal (approved):

Issue #61, opened 2024-07-01
Consider updating API design for clarity


markmc:
Consider the following:

```
ds = Dataset.from_list(samples)

skills_flow = SynthGroundedSkillsFlow(client, "mixtral", teacher_model).get_flow()
skills_pipe = Pipeline(skills_flow)

sdg = SDG([skills_pipe])
gen_data = sdg.generate(ds)
```

or:

```
ds = Dataset.from_list(samples)

mmlu_flow = MMLUBenchFlow(client, teacher_model).get_flow()
mmlu_pipe = Pipeline(mmlu_flow)
knowledge_flow = SynthKnowledgeFlow(client, teacher_model).get_flow()
knowledge_pipe = Pipeline(knowledge_flow)

sdg = SDG([mmlu_pipe, knowledge_pipe])
gen_data = sdg.generate(ds)
```

Consider the nouns:

- **Dataset** - this is [from Hugging Face's datasets library](https://huggingface.co/docs/datasets/main/en/package_reference/main_classes#datasets.Dataset)
- **Block** - not shown in the code above, but required to understand a flow - a block provides a `generate()` method transforms an input dataset and returns an output dataset
-  **Block config** - a description of how to instantiate and invoke a block
- **Flow** - a class which describes how to render a sequence of block configs from a template
- **Pipeline** - a pipeline is created from a sequence of block configs, and provides a `generate()` method in which it instantiates and invokes blocks in turn, passing the input dataset and collecting the output
- **SDG** - an SDG is created from a list of pipelines, and its `generate()` method calls pipelines in turn

Proposals:

1. Remove `SDG` - we don't need both `SDG` and `Pipeline` since `Pipeline` can already do everything `SDG` can do
2. Model `Flow` as a block config template - it would be more clear if we reinforced the idea that a "flow" is a template of a block config sequence - a `render()` method make sense to me, and an extensible `params` object for the common case of instantiating multiple flows
3. Create a pipeline from a sequence of flows - add a `Pipeline.from_flows()` convenience class method to Pipeline that knows how to render block configs from a sequence of flows

So we could have e.g.

```
ds = Dataset.from_list(samples)

flow_params = FlowParams(client, "mixtral", teacher_model)

block_configs = SynthGroundedSkillsFlow(flow_params).render(params)

skills_pipe = Pipeline(block_configs)

gen_data = skills_pipe.generate(ds)
```

or:

```
ds = Dataset.from_list(samples)

[...]

PR #64, opened 2024-07-02
Remove the unnecessary SDG class

markmc:
Part of #61 and instructlab/dev-docs#113

A pipeline chains together a sequence of blocks, and an SDG chains together a sequence of pipelines. There is no need for this additional layer - we can construct a pipeline with the full sequence of blocks, and not chain pipelines together.

Check out how the API is used in `generate_data.py` to get a sense of the improvement


russellb:
Assuming `e2e` passes, I am definitely supportive of this general direction. There's some discussion on #61 with @shivchander, so I'm not going to approve/merge just yet to give time for his input.

markmc:
> Assuming `e2e` passes, I am definitely supportive of this general direction. There's some discussion on #61 with @shivchander, so I'm not going to approve/merge just yet to give time for his input.

The discussion has stalled for now

I've reduced the scope of this PR to just the `SDG` thing, since the rest is now incorporated into #86

russellb (approved):
I kicked off the CI job that runs the full pipeline.

Otherwise, this looks like a nice simplification to me.

hickeyma (commented):
@shivchander you raised reasons for maintaining the SDG class in https://github.com/instructlab/sdg/issues/61#issuecomment-2203700941. What do you think of the changes in this PR?

markmc:
See also instructlab/dev-docs#109

hickeyma:
> See also https://github.com/instructlab/dev-docs/pull/109

@markmc do you mean https://github.com/instructlab/dev-docs/pull/113 instead?

markmc:
> > See also [instructlab/dev-docs#109](https://github.com/instructlab/dev-docs/pull/109)
> 
> @markmc do you mean [instructlab/dev-docs#113](https://github.com/instructlab/dev-docs/pull/113) instead?

Doh, yes - thank you

hickeyma (commented):
@markmc Is this blocked waiting on the design proposal to progress?

markmc:
> @markmc Is this blocked waiting on the design proposal to progress?

Thanks for the reminder. I've rebased

I think the only blocker is that @shivchander has not yet acked [my YAGNI argument](https://github.com/instructlab/sdg/issues/61#issuecomment-2205460470)

shivchander:
On the research side, we are actively working on developing complex flows that chain pipelines in a DAG fashion. Our goal with the SDG class wass to serve as an abstract wrapper capable of managing these complex flows.

Could this be renamed to something better? Yes. But I and the research team are not in favor of removing this, but I will not be blocker on this PR, If there's a consensus on moving forward. Just wanted to give a heads-up that at the pace research moves, we might need something which can handle this sooner.

markmc:
> On the research side, we are actively working on developing complex flows that chain pipelines in a DAG fashion. Our goal with the SDG class wass to serve as an abstract wrapper capable of managing these complex flows.
> 
> Could this be renamed to something better? Yes. But I and the research team are not in favor of removing this, but I will not be a blocker on this PR, If there's a consensus on moving forward. Just wanted to give a heads-up that at the pace research moves, we might need something which can handle this sooner.

It's very difficult to remove an API once a library is starting to be used in the wild - and renaming an API is effectively removing an API. Library API maintainers need to be very careful about breaking downstream users. Removing an API generally involves a long deprecation period, to make sure users have time to migrate away from it.

I do not believe something called "SDG" will live on as the name of any abstraction in this library long-term, and I do not want to make it difficult to remove for no reason

Adding whatever the new abstraction should be is super easy, and made no more difficult by removing SDG

markmc:
> It's very difficult to remove an API once a library is starting to be used in the wild - and renaming an API is effectively removing an API. Library API maintainers need to be very careful about breaking downstream users. Removing an API generally involves a long deprecation period, to make sure users have time to migrate away from it.

We will release 0.2.2 or 0.2.3 any day now, which will be used by instructlab 0.18.0, and from that point on I expect we'll be much more cautious about incompatible API changes. For that reason, I'm going to go ahead with this change despite objections, on the basis that it is trivial to restore this abstraction when we need it.

PR #65, opened 2024-07-02
Updated chunking_document.

PalmPalm7:
#### Updates:

1. Applied document-specific test splitter from Langchain in replace of original naive version.
2. Made heuristics changes to markdown file, especially using regex to trim markdown tables in attempt to fit in the whole table with limited context window.
3. For updated chunk_document() function, see [Chunking_Demo.ipynb](https://colab.research.google.com/drive/1lYHhhqQaOkKQCoFkhx0M7AT7N5UUvdRa?usp=sharing) on chunking with server_ctx_size=4096, chunk_word_count=1024). Granite 7b has 4k context windows.

#### Saved for later:

>0. "Soft dependency" freeze and introducing Magika should wait.

For further PR, an [efficient library](https://github.com/google/magika/tree/main) to detect file type could be added for document-specific chunking and heuristics.

> 1. How do you handle chunks that are larger than the provided server_ctx_size ?

No logic on larger chunk size updated this PR. "CharacterTextSplitter will only split on separator (which is '\n\n' by default). chunk_size is the maximum chunk size that will be split if splitting is possible." See this [StackOverflow Post](https://stackoverflow.com/questions/76633836/what-does-langchain-charactertextsplitters-chunk-size-param-even-do).

> 2. Is it assumed that if the file is not any of code file it always markdown?

My Justification is there is I assume our most common use cases, as we discussed, is PDF into markdown formats, therefore default case should be markdown. Furthermore, by specifying the language param in RecursiveCharacterTextSplitter, it uses these following separators:
```
RecursiveCharacterTextSplitter.get_separators_for_language(Language.MARKDOWN)

Output:
['\n#{1,6} ',
 '```\n',
 '\n\\*\\*\\*+\n',
 '\n---+\n',
 '\n___+\n',
 '\n\n',
 '\n',
 ' ',
 '']
```

Instead of [these separators](https://github.com/instructlab/sdg/blob/1f71fb67aa46151bc0362e733b06529a1c609e6d/src/instructlab/sdg/utils/chunking.py#L41C1-L42C1).
 ```
["\n\n", "\n", " "]
```

Original PR, see: https://github.com/instructlab/sdg/pull/45

### DIFF
--- a/src/instructlab/sdg/utils/chunking.py
+++ b/src/instructlab/sdg/utils/chunking.py
@@ -2,11 +2,15 @@
 
 # Standard
 from typing import List
+import logging
+import re
 
 # Third Party
-from langchain_text_splitters import RecursiveCharacterTextSplitter
+from langchain_text_splitters import Language, RecursiveCharacterTextSplitter
 
 _DEFAULT_CHUNK_OVERLAP = 100
+
+logger = logging.getLogger(__name__)
 
 
 def _num_tokens_from_words(num_words) -> int:
@@ -21,12 +25,24 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
     """
     Iterates over the documents and splits them into chunks based on the word count provided by the user.
     Args:
-        documents (dict): List of documents retrieved from git (can also consist of a single document).
+        documents (list): List of documents retrieved from git (can also consist of a single document).
         server_ctx_size (int): Context window size of server.
         chunk_word_count (int): Maximum number of words to chunk a document.
     Returns:
          List[str]: List of chunked documents.
     """
+
+    # Checks for input type error
+    if isinstance(documents, str):
+        documents = [documents]
+        logger.info(
+            "Converted single string into a list of string. Assumed the string passed in is the document. Normally, chunk_document() should take a list as input."
+        )
+    elif not isinstance(documents, list):
+        raise TypeError(
+            "Expected: documents to be a list, but got {}".format(type(documents))
+        )
+
     no_tokens_per_doc = _num_tokens_from_words(chunk_word_count)
     if no_tokens_per_doc > int(server_ctx_size - 1024):
         raise ValueError(
@@ -36,15 +52,24 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
                 )
             )
         )
+    # Placeholder for params
     content = []
-    text_splitter = RecursiveCharacterTextSplitter(
-        separators=["\n\n", "\n", " "],
-        chunk_size=_num_chars_from_tokens(no_tokens_per_doc),
-        chunk_overlap=_DEFAULT_CHUNK_OVERLAP,
+    chunk_size = _num_chars_from_tokens(no_tokens_per_doc)
+    chunk_overlap = _DEFAULT_CHUNK_OVERLAP
+
+    # Using Markdown as default, document-specific chunking will be implemented in seperate pr.
+    text_splitter = RecursiveCharacterTextSplitter.from_language(
+        language=Language.MARKDOWN,
+        chunk_size=chunk_size,
+        chunk_overlap=chunk_overlap,
     )
 
+    # Determine file type for heuristics, default with markdown
     for docs in documents:
+        # Use regex to remove unnecessary dashes in front of pipe characters in a markdown table.
+        docs = re.sub(r"-{2,}\|", "-|", docs)
+        # Remove unnecessary spaces in front of pipe characters in a markdown table.
+        docs = re.sub(r"\  +\|", " |", docs)
         temp = text_splitter.create_documents([docs])
         content.extend([item.page_content for item in temp])
-
     return content

### CODE REVIEW

**Comment by abhi1092** [comment]
Tested this PR. Chunking seems to be working as expected. @PalmPalm7 as discussed can you add the handling of document if its not passed as list?

**Comment by PalmPalm7** [comment]
> Tested this PR. Chunking seems to be working as expected. @PalmPalm7 as discussed can you add the handling of document if its not passed as list?

Updated chunking method with this logic. Ready to merge @abhi1092 .

**Review by russellb — CHANGES_REQUESTED** [review]
Can you please squash your 3 commits into 1 and add all of the description you have on the PR into the commit message?  Let me know if you need any assistance doing this!

**Comment by PalmPalm7** [comment]
> Can you please squash your 3 commits into 1 and add all of the description you have on the PR into the commit message? Let me know if you need any assistance doing this!

Thank you Russel! I have squashed and amended my commits. Ready to merge @russellb .

File: src/instructlab/sdg/utils/chunking.py [hunk 1]
```diff
@@ -36,15 +47,80 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
+                # Remove unnecessary spaces in front of pipe characters in a markdown table.
+                docs = re.sub(r"\  +\|", " |", docs)
+                temp = text_splitter.create_documents([docs])
+                content.extend([item.page_content for item in temp])
 
+        except UnicodeDecodeError as e:
+            for docs in documents:
+                temp = text_splitter.create_documents([docs])
+                content.extend([item.page_content for item in temp])
+            print("Using naive method instead. UnicodeDecode Error: {}".format(e))
```
russellb, line 115:
```suggestion
            logger.debug("Using naive method instead. UnicodeDecode Error: {}".format(e))
```

russellb (changes requested):
looks good -- passing e2e CI, and I like that it falls back to the old method in case of an error.

Instead of a raw `print()` I would prefer using a logger. Can you add a new parameter for `logger`, update consumers to pass in a logger, and then use that for the messages you're printing?  Thanks!

PalmPalm7:
> looks good -- passing e2e CI, and I like that it falls back to the old method in case of an error.
> 
> Instead of a raw `print()` I would prefer using a logger. Can you add a new parameter for `logger`, update consumers to pass in a logger, and then use that for the messages you're printing? Thanks!

Done! @russellb 

I've added this parameter:
`logger = logging.getLogger(__name__)` and replaced all existing print functions.

abhi1092 (approved):

src/instructlab/sdg/utils/chunking.py
```diff
@@ -21,12 +26,24 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
-        documents (dict): List of documents retrieved from git (can also consist of a single document).
+        documents (list): List of documents retrieved from git (can also consist of a single document).
         server_ctx_size (int): Context window size of server.
         chunk_word_count (int): Maximum number of words to chunk a document.
     Returns:
          List[str]: List of chunked documents.
     """
+
+    # Checks for input type error
+    if isinstance(documents, str):
```
RobotSail, line 37: What you have here is correct and will behave as expected, and to make it more concise you could rewrite it as:
```py
if not (isinstance(documents, str) or isinstance(documents, list)):
       raise TypeError(
            "Expected: documents to be a list or str, but got {}".format(type(documents))
        )

documents = [documents] if isinstance(documents, str) else documents
```

Since you've already eliminated the case where it's neither a list or str, you can proceed assuming everything is correct.

src/instructlab/sdg/utils/chunking.py
```diff
@@ -21,12 +26,24 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
+        documents = [documents]
+        logger.info(
```
RobotSail, line 39: I would change this to `logger.debug` since you're providing details that only programmers would understand. Imagine `INFO` as the level of information you'd want to see when you're running a program and you don't know how it works internally.

src/instructlab/sdg/utils/chunking.py
```diff
@@ -36,15 +53,82 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
+                text_splitter = RecursiveCharacterTextSplitter.from_language(
+                    language=Language.MARKDOWN,
+                    chunk_size=chunk_size,
+                    chunk_overlap=chunk_overlap,
+                )
+                # Use regex to remove unnecessary dashes in front of pipe characters in a markdown table.
+                docs = re.sub(r"-{2,}\|", "-|", docs)
+                # Remove unnecessary spaces in front of pipe characters in a markdown table.
+                docs = re.sub(r"\  +\|", " |", docs)
+                temp = text_splitter.create_documents([docs])
```
RobotSail, line 114: It looks like these two lines where `temp` is assigned and `content.extend` is called is common to both conditional branches. Could this be something we can move out from these statements?

src/instructlab/sdg/utils/chunking.py
```diff
@@ -36,15 +53,82 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
+                "rust": Language.RUST,
+                "latex": Language.LATEX,
+                "html": Language.HTML,
+                "cs": Language.CSHARP,
+                "c": Language.C,
+                "perl": Language.PERL,
+            }
+            if file_type in file_type_to_language:
+                language = file_type_to_language[file_type]
+                text_splitter = RecursiveCharacterTextSplitter.from_language(
```
RobotSail, line 94: I would rename this so that we don't override the top-level `text_splitter` object.

src/instructlab/sdg/utils/chunking.py
russellb: ah, yeah, that's what I had suggested originally. I can submit a follow-up PR for it

src/instructlab/sdg/utils/chunking.py
```diff
@@ -36,15 +53,82 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
+                )
+                # Use regex to remove unnecessary dashes in front of pipe characters in a markdown table.
+                docs = re.sub(r"-{2,}\|", "-|", docs)
+                # Remove unnecessary spaces in front of pipe characters in a markdown table.
+                docs = re.sub(r"\  +\|", " |", docs)
+                temp = text_splitter.create_documents([docs])
+                content.extend([item.page_content for item in temp])
+
+        except UnicodeDecodeError as e:
+            for docs in documents:
```
RobotSail, line 118: I would be careful about how we're handling these exceptions. For instance, if `UnicodeDecodeError` was thrown, then it's possible we weren't able to parse the documents correctly when performing language detection. 

If the intention here is to fallback to the default case, then we would need to apply the same logic as in the `else` block above. 

My overall recommendation is to either cut down on the scope of our try/catch here, or to move the `content.extend` logic to happen within a `finally` block.

RobotSail (approved):
Some minor comments but LGTM overall, great work!

src/instructlab/sdg/utils/chunking.py
PalmPalm7: I insist to use my original method for clarity. Thanks for the suggestions. 

Chunk_document() only intends to take a list as input; however, taking a string is a common mistake and should be avoided.

src/instructlab/sdg/utils/chunking.py
PalmPalm7: As mentioned above.

src/instructlab/sdg/utils/chunking.py
PalmPalm7: Fixed in commit.

src/instructlab/sdg/utils/chunking.py
PalmPalm7: This is valid concern. Fixed.

src/instructlab/sdg/utils/chunking.py
PalmPalm7: This is correct! Originally my thought process was to indicate a naive text_splitter as the base case, in `text_splitter_naive`, but my CoT got lost somewhere in the PR.

src/instructlab/sdg/utils/chunking.py
PalmPalm7: You are absolutely correct. I would say ValueError, superclass of UnicodeDecodeError shouldn't be thrown into the try except logic. However, UnicodeDecodeError is a common error when Magika, the file type light weight detection model, fails to take a proper input; thus catching it. I've made amends to the logger for proper error messages.

markmc:
> #### Updates:
> 1. Used a [efficient library](https://github.com/google/magika/tree/main) to detect file type.

This new dependency is coming at a time when we are in a "soft dependency" freeze - i.e. we are avoiding adding new dependencies, unless there is some exceptional reason to justify it

If we did not add this library now (i.e. waited until the next milestone release has been completed), I guess we would just use `language=Language.MARKDOWN` and have no intelligent support for code files (go, java, etc.), latex, and html?

IMO changing from the current naive list of separators to `language=Language.MARKDOWN` is a good change to make now, but adding the new dependency can wait since support for those other file formats isn't a major priority right now

russellb:
> > #### Updates:
> > 
> > 1. Used a [efficient library](https://github.com/google/magika/tree/main) to detect file type.
> 
> This new dependency is coming at a time when we are in a "soft dependency" freeze - i.e. we are avoiding adding new dependencies, unless there is some exceptional reason to justify it
> 
> If we did not add this library now (i.e. waited until the next milestone release has been completed), I guess we would just use `language=Language.MARKDOWN` and have no intelligent support for code files (go, java, etc.), latex, and html?
> 
> IMO changing from the current naive list of separators to `language=Language.MARKDOWN` is a good change to make now, but adding the new dependency can wait since support for those other file formats isn't a major priority right now

That makes a ton of sense. Thanks for thinking this through, Mark.

Markdown is the only knowledge doc format supported via the CLI right now, so it's only format we need to support at the moment.

russellb (changes requested):
Changing review based on Mark's feedback. I think he's right about the dependency and what features are required right now vs. can come later.

markmc:
@PalmPalm7 if you split the magika part into a separate PR, we could merge the rest as a super useful improvement to the markdown support :+1:

dhellmann:
> @PalmPalm7 if you split the magika part into a separate PR, we could merge the rest as a super useful improvement to the markdown support 👍

+1 -- magika introduces a dependency on onnxruntime, which is another package that [doesn't publish standard source code packages](https://pypi.org/project/onnxruntime/1.18.1/#files) so we'll have to figure out how to build it to prepare a release downstream. It would be nice if we could wait to bring that in until this first major release is settled.

PalmPalm7:
Thank you for taking the time to review and foresee potential risks @dhellmann @markmc @russellb , I'll split it into to two PR.

PalmPalm7:
> > > #### Updates:
> > > 
> > > 1. Used a [efficient library](https://github.com/google/magika/tree/main) to detect file type.
> > 
> > 
> > This new dependency is coming at a time when we are in a "soft dependency" freeze - i.e. we are avoiding adding new dependencies, unless there is some exceptional reason to justify it
> > If we did not add this library now (i.e. waited until the next milestone release has been completed), I guess we would just use `language=Language.MARKDOWN` and have no intelligent support for code files (go, java, etc.), latex, and html?
> > IMO changing from the current naive list of separators to `language=Language.MARKDOWN` is a good change to make now, but adding the new dependency can wait since support for those other file formats isn't a major priority right now
> 
> That makes a ton of sense. Thanks for thinking this through, Mark.
> 
> Markdown is the only knowledge doc format supported via the CLI right now, so it's only format we need to support at the moment.

Updated logic according to Mark and Doug's suggestions above. Ready to merge @russellb .

dhellmann:
> Thank you for taking the time to review and foresee potential risks @dhellmann @markmc @russellb , I'll split it into to two PR.

Thank you for being flexible!

oindrillac:
@russellb can you please check if the changes addressed your concern above?

src/instructlab/sdg/utils/chunking.py
```diff
@@ -36,15 +52,24 @@ def chunk_document(documents: List, server_ctx_size, chunk_word_count) -> List[s
...
+    chunk_overlap = _DEFAULT_CHUNK_OVERLAP
+
+    # Using Markdown as default, document-specific chunking will be implemented in seperate pr.
+    text_splitter = RecursiveCharacterTextSplitter.from_language(
+        language=Language.MARKDOWN,
+        chunk_size=chunk_size,
+        chunk_overlap=chunk_overlap,
     )
 
+    # Determine file type for heuristics, default with markdown
```
russellb, line 67: looks like this comment may be out of date, but not a big deal

russellb (approved):

PR #67, opened 2024-07-02
Fix prompt file paths for an installed library

russellb:
2debc77 default_flows: Move sdg_base into base Flow class
3e01b18 Fix remaining prompt file paths

commit 2debc7792a3db75ed90ed59eb42678ac0592d32c
Date:   Tue Jul 2 16:36:28 2024 -0400

    default_flows: Move sdg_base into base Flow class
    
    Each subclass was calculating the root directory of the python package
    for finding prompt templates that are embedded in the package. Move
    this to the base `Flow` class so we're only doing it in one place.
    
commit 3e01b18112fbabd418f028697288366d1c3f8ae5
Date:   Tue Jul 2 16:42:21 2024 -0400

    Fix remaining prompt file paths
    
    Special handling is required to find these prompt files from an
    installed version of the instructlab-sdg package. This updates the
    rest of this file to make use of `self.sdg_base` from the base `Flow`
    class which will work from an installed library.
    
### DIFF
--- a/src/instructlab/sdg/default_flows.py
+++ b/src/instructlab/sdg/default_flows.py
@@ -36,6 +36,7 @@ class Flow(ABC):
         self.model_id = model_id
         self.num_iters = num_iters
         self.batched = batched
+        self.sdg_base = resources.files(__package__)
 
     @abstractmethod
     def get_flow(self) -> list:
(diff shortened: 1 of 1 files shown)


aakankshaduggal (approved):
Thanks @russellb

markmc:
Nice catch, thanks for separating the no-op refactor and bugfix into separate commits - lgtm

oindrillac (approved):

PR #68, opened 2024-07-02
e2e: Only run one job at a time for a given PR

russellb:
If you push updates to a PR, currently we're leaving `e2e` jobs for
all past revisions to keep running. If you do a few quick updates,
that ends up wasting a lot of resources.

This change ensures that for a given PR, only one of these jobs runs
at a time. A previous run will get canceled and replaced with one
against the latest version of the PR.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -22,6 +22,10 @@ on:
       - 'requirements*.txt'
       - '.github/workflows/e2e.yml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


markmc:
Nice catch, same change as instructlab/instructlab#1567 - lgtm

oindrillac (approved):
Thanks!

russellb:
thanks for the reviews!

russellb:
> Nice catch, same change as [instructlab/instructlab#1567](https://github.com/instructlab/instructlab/pull/1567) - lgtm

... unfortunately it was our bill going up a lot faster than I predicted that made me catch it!

PR #71, opened 2024-07-03
generate_data: Fix check for `output` in results

russellb:
This code would hit an unepxected traceback if the result had the
`output` column, but its value in a given row was `None`. This check
detects that case and raises the Exception as intended.


markmc:
lgtm

danmcp (approved):

aakankshaduggal (approved):
lgtm

PR #72, opened 2024-07-03
Hotfix for Filterblocks 

shivchander:
### Major changes:
- Introduced a new parameter `valid_values` in the `FilterByValueBlock` class to specify a list of valid values for filtering.
- Added a private method `_fiter_invalid_values` to filter out samples with invalid values based on the `valid_values` list.
- Added a private method `_convert_dtype` to handle potential `ValueError` during data type conversion, setting the value to `None` if an error occurs.
- Modified the `generate` method to include a call to `_fiter_invalid_values` after the data type conversion (if applicable), ensuring that only valid samples are processed further.

### Minor changes: 
- Slightly better logging added to `pipeline.py` - added input and output logging and some new lines to make it more readable
- Updated test script imports to use instructlab.sdg instead of importing from src
- Fixed default flows to use ints and not floats as expected in the template

### DIFF
--- a/src/instructlab/sdg/default_flows.py
+++ b/src/instructlab/sdg/default_flows.py
@@ -194,6 +194,7 @@ class SynthKnowledgeFlow(Flow):
                     "block_name": "filter_faithfulness",
                     "filter_column": "judgment",
                     "filter_value": "YES",
+                    "valid_values": ["YES", "NO"],
                     "operation": operator.eq,
                     "batch_kwargs": {
                         "num_procs": 8,
@@ -226,9 +227,10 @@ class SynthKnowledgeFlow(Flow):
                 "block_config": {
                     "block_name": "filter_relevancy",
                     "filter_column": "score",
-                    "filter_value": 2.0,
+                    "filter_value": 2,
+                    "valid_values": [0, 1, 2],
                     "operation": operator.eq,
-                    "convert_dtype": float,
+                    "convert_dtype": int,
                     "batch_kwargs": {
                         "num_procs": 8,
                     },
@@ -260,7 +262,8 @@ class SynthKnowledgeFlow(Flow):
                 "block_config": {
                     "block_name": "filter_verify_question",
                     "filter_column": "rating",
-                    "filter_value": 1.0,
+                    "filter_value": 1,
+                    "valid_values": [0, 1],
                     "operation": operator.eq,
                     "convert_dtype": float,
                     "batch_kwargs": {
@@ -312,9 +315,10 @@ class SynthSkillsFlow(Flow):
                 "block_config": {
                     "block_name": "filter_questions",
                     "filter_column": "score",
-                    "filter_value": 1.0,
+                    "filter_value": 1,
+                    "valid_values": [0, 1],
                     "operation": operator.eq,
-                    "convert_dtype": float,
+                    "convert_dtype": int,
                     "batch_kwargs": {
                         "num_procs": 8,
                     },
@@ -356,9 +360,10 @@ class SynthSkillsFlow(Flow):
                 "block_config": {
                     "block_name": "filter_qa_pair",
                     "filter_column": "score",
-                    "filter_value": 2.0,
+                    "filter_value": 2,
+                    "valid_values": [1, 2, 3],
                     "operation": operator.ge,
-                    "convert_dtype": float,
+                    "convert_dtype": int,
                     "batch_kwargs": {
                         "num_procs": 8,
                     },
@@ -432,9 +437,10 @@ class SynthGroundedSkillsFlow(Flow):
                 "block_config": {
                     "block_name": "filter_grounded_questions",
                     "filter_column": "score",
-                    "filter_value": 1.0,
+                    "filter_value": 1,
+                    "valid_values": [0, 1],
                     "operation": operator.eq,
-                    "convert_dtype": float,
+                    "convert_dtype": int,
                     "batch_kwargs": {
                         "num_procs": 8,
                     },
@@ -476,9 +482,10 @@ class SynthGroundedSkillsFlow(Flow):
                 "block_config": {
                     "block_name": "filter_grounded_qa_pair",
                     "filter_column": "score",
-                    "filter_value": 2.0,
+                    "filter_value": 2,
+                    "valid_values": [1, 2, 3],
                     "operation": operator.ge,
-                    "convert_dtype": float,
+                    "convert_dtype": int,
                     "batch_kwargs": {
                         "num_procs": 8,
                     },
--- a/src/instructlab/sdg/default_flows.py
+++ b/src/instructlab/sdg/default_flows.py
@@ -411,6 +411,7 @@ class SynthGroundedSkillsFlow(Flow):
                     "output_cols": ["question"],
                     "batch_kwargs": {
                         "num_procs": 8,
+                        "num_samples": 3,
                         "batched": self.batched,
                     },
                 },
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -20,14 +20,14 @@ class FilterByValueBlock(Block):
         self.valid_values = valid_values
         self.convert_dtype = convert_dtype
         self.num_procs = batch_kwargs.get("num_procs", 1)
-    
+
     def _fiter_invalid_values(self, samples):
         samples = samples.filter(
             lambda x: x[self.column_name] in self.valid_values,
             num_proc=self.num_procs,
         )
         return samples
-    
+
     def _convert_dtype(self, sample):
         try:
             sample[self.column_name] = self.convert_dtype(sample[self.column_name])
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -31,7 +31,8 @@ class FilterByValueBlock(Block):
     def _convert_dtype(self, sample):
         try:
             sample[self.column_name] = self.convert_dtype(sample[self.column_name])
-        except ValueError:
+        except ValueError as e:
+            logger.error("Error converting dtype: %s, filling with None to be filtered later", e)
             sample[self.column_name] = None
         return sample
 
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -11,24 +11,38 @@ logger = setup_logger(__name__)
 
 class FilterByValueBlock(Block):
     def __init__(
-        self, filter_column, filter_value, operation, convert_dtype=None, **batch_kwargs
+        self, filter_column, filter_value, operation, valid_values, convert_dtype=None, **batch_kwargs
     ) -> None:
         super().__init__(block_name=self.__class__.__name__)
         self.value = filter_value
         self.column_name = filter_column
         self.operation = operation
+        self.valid_values = valid_values
         self.convert_dtype = convert_dtype
         self.num_procs = batch_kwargs.get("num_procs", 1)
+    
+    def _fiter_invalid_values(self, samples):
+        samples = samples.filter(
+            lambda x: x[self.column_name] in self.valid_values,
+            num_proc=self.num_procs,
+        )
+        return samples
+    
+    def _convert_dtype(self, sample):
+        try:
+            sample[self.column_name] = self.convert_dtype(sample[self.column_name])
+        except ValueError:
+            sample[self.column_name] = None
+        return sample
 
     def generate(self, samples) -> Dataset:
         if self.convert_dtype:
             samples = samples.map(
-                lambda x: {
-                    **x,
-                    self.column_name: self.convert_dtype(x[self.column_name]),
-                },
+                self._convert_dtype,
                 num_proc=self.num_procs,
             )
+
+        samples = self._fiter_invalid_values(samples)
 
         return samples.filter(
             lambda x: self.operation(x[self.column_name], self.value),
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -11,7 +11,13 @@ logger = setup_logger(__name__)
 
 class FilterByValueBlock(Block):
     def __init__(
-        self, filter_column, filter_value, operation, valid_values, convert_dtype=None, **batch_kwargs
+        self,
+        filter_column,
+        filter_value,
+        operation,
+        valid_values,
+        convert_dtype=None,
+        **batch_kwargs,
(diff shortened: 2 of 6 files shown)


src/instructlab/sdg/default_flows.py
```diff
@@ -194,6 +194,7 @@ def get_flow(self) -> list:
                     "block_name": "filter_faithfulness",
                     "filter_column": "judgment",
                     "filter_value": "YES",
+                    "valid_values": ["YES", "NO"],
```
markmc, line 197: The filter currently only accepts `judgement=YES` ... so adding is a no-op?

src/instructlab/sdg/default_flows.py
```diff
@@ -226,9 +227,10 @@ def get_flow(self) -> list:
                 "block_config": {
                     "block_name": "filter_relevancy",
                     "filter_column": "score",
-                    "filter_value": 2.0,
+                    "filter_value": 2,
+                    "valid_values": [0, 1, 2],
```
markmc, line 231: Same here - if we're only accepting `score=2` then why is `valid_values` needed?

markmc:
Thanks for the explanations - please include these in the commit messages though, so they show up in `git log` and `git blame`

> ### Major changes:
> * Introduced a new parameter `valid_values` in the `FilterByValueBlock` class to specify a list of valid values for filtering.

Every instance of using this appears to be a no-op since it filters a subset of what `filter_value` is already filtering?

> * Added a private method `_convert_dtype` to handle potential `ValueError` during data type conversion, setting the value to `None` if an error occurs.

We shouldn't be suppressing errors IMO - it will make things really hard to debug

src/instructlab/sdg/default_flows.py
aakankshaduggal: We saw that the judgement was sometimes getting populated with values besides "YES" and "NO" and that was breaking the code. Hence it was important to add a set of valid values or expected values from the evaluation.

src/instructlab/sdg/filterblock.py
```diff
@@ -11,25 +11,39 @@
...
+            lambda x: x[self.column_name] in self.valid_values,
+            num_proc=self.num_procs,
+        )
+        return samples
+
+    def _convert_dtype(self, sample):
+        try:
+            sample[self.column_name] = self.convert_dtype(sample[self.column_name])
+        except ValueError:
+            sample[self.column_name] = None
```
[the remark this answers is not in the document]
aakankshaduggal: Thanks for the suggestion Mark, we can add a log here.

src/instructlab/sdg/default_flows.py
```diff
@@ -312,9 +315,10 @@ def get_flow(self) -> list:
                 "block_config": {
                     "block_name": "filter_questions",
                     "filter_column": "score",
-                    "filter_value": 1.0,
+                    "filter_value": 1,
+                    "valid_values": [0, 1],
                     "operation": operator.eq,
-                    "convert_dtype": float,
+                    "convert_dtype": int,
```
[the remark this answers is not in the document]
shivchander: This is technically a bug fix - if you look at the prompt template for this block, it actually expects ints

https://github.com/instructlab/sdg/blob/main/src/instructlab/sdg/configs/skills/evaluate_freeform_questions.yaml#L28

will update the description to note this

src/instructlab/sdg/default_flows.py
shivchander: This change for introducing valid_values is not shortsighted for the current scope and flows. This is useful in making the filter blocks more generic. For instance if you need to filter the outputs from a block using the `in` operation to check if the generated output belongs in a list of strings. 

As a general note and practice, language model output are not deterministic, restricting with these explicit valid_values makes it more generic and robust

src/instructlab/sdg/default_flows.py
shivchander: Refer to above reply, same case

src/instructlab/sdg/default_flows.py
markmc: Ok, if this is an "we don't need it now, but we think we'll need it later" argument - I'll refer again to the YAGNI principle :)

But my "we don't need it now" conclusion might be a misunderstanding. Here's pseudo-code showing my understanding:

```
    def filter_sample(sample):
        value = sample["judgment"]
        if value not in ["YES", "NO"]:
            return False
        if value != "YES":
            return False
        return True
```

Am I missing something?

(If I'm understanding correctly, I think it's important we avoid adding additional unnecessary complexity like this, because it's confusing enough already)

markmc:
> > * Introduced a new parameter `valid_values` in the `FilterByValueBlock` class to specify a list of valid values for filtering.
> 
> Every instance of using this appears to be a no-op since it filters a subset of what `filter_value` is already filtering?

A discussion on this point continues in a [resolved comment](https://github.com/instructlab/sdg/pull/72#discussion_r1664534233)

markmc:
> > > * Introduced a new parameter `valid_values` in the `FilterByValueBlock` class to specify a list of valid values for filtering.
> > 
> > 
> > Every instance of using this appears to be a no-op since it filters a subset of what `filter_value` is already filtering?
> 
> A discussion on this point continues in a [resolved comment](https://github.com/instructlab/sdg/pull/72#discussion_r1664534233)

I gather the point is that `filter_value` is actually required for downstream custom flows (see instructlab/dev-docs#109). That's an entirely different situation!

Please go ahead and add `filter_value` for the custom flow use case, but do not use it in the current flows where it is not required and just adds complexity

src/instructlab/sdg/filterblock.py
```diff
@@ -11,25 +11,48 @@
 
 class FilterByValueBlock(Block):
     def __init__(
-        self, filter_column, filter_value, operation, convert_dtype=None, **batch_kwargs
+        self,
+        filter_column,
+        filter_value,
+        operation,
+        valid_values,
```
russellb, line 18: Something that would help a lot here is adding some documentation of these parameters that explains how they're all to be used. That would, I hope, help some of the discussions we're having trying to understand the changes. Can you add a docstring to this method?

src/instructlab/sdg/default_flows.py
russellb: I'm with @markmc -- that pseudocode matches my understanding, and so I have the same "can you help me understand why we need it?" question

russellb: (I also unresolved the comment thread since it's still active)

src/instructlab/sdg/filterblock.py
oindrillac: https://github.com/instructlab/sdg/issues/76 Sounds good, lets address this in a follow up

markmc: Yeah, one thing that helps is ... `filter_column`, `filter_value`, and `operation` go together

`filter_column` and `valid_values` go together too

so, it's like:

- filter on filter_column,
- first (and optionally) by checking the column contains one of valid_values
- second by comparing the value in the column matches filter_value using the given operation

russellb: and the confusion is "why do we need the second bullet again?"  right?

is it that actually you really only need one or the other?

or what's the example for when you need them both together?

russellb: ... or put another way, two types of filtering?

could they be collapsed into one, where it's always a set, and you might just have only one element in that set?

russellb (changes requested):
just making my review explicit while the active discussion continues in comment threads

src/instructlab/sdg/default_flows.py
mairin:
```
"filter_value": "YES",
"valid_values": ["YES", "NO"],
```

OK so filter_value and valid_values are not related.

filter_value == what the model response must be in order for it to move to the next step of the workflow.

valid_values is simply a way to discard any invalid responses from the model *before* we do any operations. 

They are completely unrelated. So the pseudo code logic isn't really applicable.

```

def validate_responses(sample):
  value = sample["judgement"]
  if value not in ["YES", "NO"]:
    return False
  return True

def promote_next_step(sample):
  value = sample["judgement"]
  if value != "YES":
    return False
  return True

filter_sample(sample)
promote_next_step(sample)
```

markmc: My talking about pseudo code probably didn't help.

The [real code](https://github.com/instructlab/sdg/blob/ed7d74bf318cabe8b9994f9aec76201f1639736e/src/instructlab/sdg/filterblock.py#L54-L58) is more like:

```
samples = samples.filter(lambda x: x['judgement'] in ['YES', 'NO'])
samples = samples.filter(lambda x: operator.eq(x['judgement'], 'YES'))
```

The second line (i.e. the existing functionality) is clearly sufficient on its own

Or indeed, we could simply do (just the new `valid_values` functionality, replacing the old):

```
samples = samples.filter(lambda x: x['judgement'] in ['YES'])
```

What's puzzling is the notion that we need both lines?

(I apologize profusely in advance if I've missed something!)

aakankshaduggal (approved):
Thanks @shivchander 
looks good to me 🚢 
Tested this using the vllm endpoint and works fine 💯 

Approving this but can we please merge this after https://github.com/instructlab/sdg/pull/77/

src/instructlab/sdg/default_flows.py
mairin: @markmc Hey. So the point here is the end user building the config for their dataset may want to configure what is the valid response. And it doesnt make sense to check the valid response against the full dataset when some of it may be garbage / invalid. 

Like for example:

In one case, the user may configure (in the pseudocode example) that they want the YES answer to pass through. They may then decide later on actually they're getting better results w the NO answer and only want to pass thru NO. 

The full dataset with garbage may be (just making numbers up) 1000 things. The dataset with only valid things may be 500 things.

Why run "is it NO?" against 1000 things when you could run it against 500? Maybe doesnt seem like a practical optimization for a one time run. But when you're creating a custom sdg configuration you're running thru these things multiple times over and over to get to a workflow that works for you, so its like

1. filter large dataset
2. see results, tweak config
3. filter large dataset
4. see results, tweak config
5. filter large dataset
6. see results, tweak config
7. *n

vs

1. clean up larger data set
2. operate on smaller dataset
3. see results, tweak
4. operate on smaller dataset
5. see results, tweak
6. operate on small dataset
7.  *n

(Does that make sense? Drafting this on a mobile on the train 😅)

Let's take a more complex example. Valid responses are blue, green, red, yellow. Full dataset is blue, green, red, yellow, and garbage. 

User decides I only want red to pass forward. Do the data cleanse step, then only pass forward the red. Then user understands they need red and blue. They would either have to change config to red,blue and rerun against whole dataset or simply rerun against clean dataset. 

The first phase is to clean the dataset down to a minimal usable set.

The second is to operate on the dataset.

The operation may be more complex than simply matching a value. It's customizable. 

Does this make sense? So I would actually suggest (if the python pseudocode is reflective of the actual code) to make the filter_value more generic to support a more complex match that a single value ==. 

@aakankshaduggal @oindrillac Let me know if I misunderstood or misrepresented anything

aakankshaduggal: Thanks @mairin for this detailed explanation. Makes sense 💯

russellb: The real code isn't more complicated than the psuedo code at this point, ha.

https://github.com/shivchander/sdg/blob/ed7d74bf318cabe8b9994f9aec76201f1639736e/src/instructlab/sdg/filterblock.py#L47-L59

https://github.com/shivchander/sdg/blob/ed7d74bf318cabe8b9994f9aec76201f1639736e/src/instructlab/sdg/filterblock.py#L30-L35

So both steps are operating on the same dataset, back-to-back in the same code path. The two operations are nearly identical, but use slightly different lambdas.  The intermediary dataset in between is not exposed anywhere because it goes straight from one filter to the next. 

@mairin your description kind of makes it sound like an optimization, but it's hard for me to see where there's an efficiency gain by doing two back-to-back filter operations instead of one. What am I missing about this as an optimization?  or it also sounds like a feature ... but it's also an urgent bug fix, right?  What is the broken thing?

I'm trying to guess what error may have inspired this. Reading the PR again, I can see a bug fix implemented in the new `_convert_dtype()` method.

https://github.com/shivchander/sdg/blob/ed7d74bf318cabe8b9994f9aec76201f1639736e/src/instructlab/sdg/filterblock.py#L37-L45

The previous code did not handle the `ValueError` exception, so it would have broken on bogus data. That fix makes sense!

... but the new filtering on valid_values doesn't seem related to fixing that problem. It seems like a new feature (that we're caught up in a long discussion on its merits)

After all that ... is this PR the combination of two things:

1. a bug fix (the dtype fix I quoted above)
2. a new feature, (the valid values filter)

?

If so, I would suggest submitting the bug fix on its own, so its not held up by the feature discussion.

russellb: If that analysis is correct, I pulled that bug fix out here and included test code that will run in CI as well: https://github.com/instructlab/sdg/pull/78

shivchander: I will close this PR and reintroduce the retaining valid values as its own block (since we still need it). Thanks for isolating the bug fix

shivchander:
Closing this PR in favor of #78 and will create a new one to handle the valid values as its own Block

src/instructlab/sdg/default_flows.py
mairin: @russellb it's a bugfix as Aakanksha already explained. Without having a valid_values filter, the code was breaking. I don't know the specifics but clearly it's offered as a bug fix, and Aakanksha tested it independently.

russellb: I'm sorry, @mairin, but I've looked at this long enough that that just isn't correct. I fully believe there may be a use case for it and assume further discussion will help clarify that, but at least as used in the existing pipelines within this PR, the filter part is not fixing anything.

russellb: @shivchander 

> I will close this PR and reintroduce the retaining valid values as its own block (since we still need it).

Sounds good. It looks like you can keep it one block, but just change the value the current block takes to be a set instead of a single value. Then both use cases can be served by the same block.

russellb: @shivchander here's what I had in mind to just augment the existing block -- https://github.com/instructlab/sdg/pull/81/commits/6c1f3eee12d47bf4be7fa23282732d3c8a143028

though note I've only tested via that unit test so far

Issue #63, opened 2024-07-02
Re-organize API exports


markmc:
Below is a list of out current API exports and a proposal for how they should be exposed

Top-level package:

- `sdg.generate_data()` - This is the main interface used by the CLI. Long-term I question whether this belongs in the library (see #instructlab/instructlab#1519)
- `sdg.SDG` -> I propose removing it in #61
- `sdg.pipeline.Pipeline` -> `sdg.Pipeline` (same for `PipelineContext`)
- `sdg.block.Block` -> `sdg.Block`

Blocks:

- `sdg.filterblock.FilterByValueBlock` -> `sdg.FilterByValueBlock`
- `sdg.importblock.FilterByValueBlock` -> `sdg.ImportBlock`
- `sdg.llmblock.LLMBlock` -> `sdg.LLMBlock`
- `sdg.llmblock.ConditionalLLMBlock` -> `sdg.ConditionalLLMBlock`
- `sdg.utilblocks.SamplePopulatorBlock` -> `sdg.SamplePopulatorBlock`
- `sdg.utilblocks.SelectorBlock` -> `sdg.SelectorBlock`
- `sdg.utilblocks.CombineColumnsBlock` -> `sdg.CombineColumnsBlock`

Default pipelines:

- `sdg.SIMPLE_PIPELINES_PACKAGE` and `sdg.FULL_PIPELINES_PACKAGE`

Exceptions:

- `FilterByValueBlockError`
- `EmptyDatasetError`
- `PipelineConfigParserError`
- `GenerateException`
- `TaxonomyReadingException`

russellb:
related issue: #30

PR #73, opened 2024-07-03
Export public APIs in top-level package

tiran:
Export `SDG`, `GenerateException`, and `generate_data()` in the top-level package name. Add notes to `generate_data` and `GenerateException` that the function and class are part of the public API.


markmc:
See #63

src/instructlab/sdg/__init__.py
```diff
@@ -1,3 +1,13 @@
 # SPDX-License-Identifier: Apache-2.0
+
+# NOTE: This package imports Torch and other heavy packages.
+__all__ = (
+    "GenerateException",
+    "SDG",
+    "generate_data",
```
markmc, line 7: You can't create an SDG without a Pipeline, and you can't create a Pipeline without a Flow, so I'd add

```
pipeline.Pipeline
default_flows.Flow
default_flows.SimpleKnowledgeFlow 
default_flows.SimpleFreeformSkillFlow
default_flows.SimpleGroundedSkillFlow
default_flows.MMLUBenchFlow
default_flows.SynthKnowledgeFlow
default_flows.SynthSkillsFlow
default_flows.SynthGroundedSkillsFlow
```

We're expecting people to develop custom flows too, which will mean exposing Blocks too, but that can wait

src/instructlab/sdg/generate_data.py
```diff
@@ -30,7 +29,8 @@
     SynthSkillsFlow,
 )
 from instructlab.sdg.pipeline import Pipeline
-from instructlab.sdg.utils import models
+from instructlab.sdg.sdg import SDG
```
markmc, line 32: Why `sdg.sdg`?

src/instructlab/sdg/generate_data.py
```diff
@@ -30,7 +29,8 @@
...
+from instructlab.sdg.utils import GenerateException, models
```
markmc, line 33: Why not import `GenerateException` from `instructlab.sdg` - i.e. generate_data() uses the API as an external user would use it?

russellb (approved):

PR #74, opened 2024-07-03
Resolve some trivial TODOs in generate_data()

markmc:
Some pretty trivial changes to knock out some TODOs in `generate_data()`

commit 51df1954557af91abadd5c4b9f44c7ed1e97a165
```
    Remove unused generate_data() parameters
    
    These parameters aren't being used by the only user of this method,
    and so they can be removed. They also aren't being used within
    generate_data().
```

commit e63bfbe2a7a235302c5d2cc37e13a9ce0ebf7531
```
    Re-order generate_data() params to match CLI
    
    The CLI is the only user of this function, and we have some work
    to do to remove unused or defunct parameters. That's easier if
    they are listed in the same order as they are used!
```

commit 9d7c5211c9c020b106e79547c8abf386b49629b6
```
    Remove TODO about num_instructions_to_generate
    
    We have support for this parameter now.
```

commit 355ff30148a7f3b55073c00003f9080bad9c8ce2
```
    Remove defunct no-pipeline-defined error path
    
    We now have full and simple pipelines defined for knowledge, grounded
    skills, and freeform skills, so this error path can't be hit anymore.
```

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -283,8 +283,6 @@ def generate_data(
         else:
             sdg = sdg_freeform_skill
 
-        # TODO -- there is a parameter for how many samples to generate, but we ignore it so far
-
         logger.debug("Samples: %s" % samples)
         ds = Dataset.from_list(samples)
         logger.debug("Dataset: %s" % ds)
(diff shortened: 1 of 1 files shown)


russellb (approved):
Nice work. Thanks for doing the validation on which parameters are actually used.  e2e passed, so I'm confident the basics of the interface with the CLI are still working fine

PR #75, opened 2024-07-03
Fix mismatch in full pipeline outputs

russellb:
f37ecfc Fix mismatch in full pipeline outputs

commit f37ecfcbfcde5733b58ceea24941c184d1d5e5cf
Date:   Wed Jul 3 11:27:38 2024 -0400

    Fix mismatch in full pipeline outputs
    
    The full knowledge pipeline had `question` and `response` as output
    columns, while the skills pipelines used `question` and `answer`.
    
    `generate_data.py` currently expects `response` instead of `answer`.
    Instead of having to deal with both, just standardize on `response`,
    since that seems to be used more frequently. For example, various
    prompt filenames have "response" in their names.
    
### DIFF
--- a/src/instructlab/sdg/configs/skills/evaluate_freeform_pair.yaml
+++ b/src/instructlab/sdg/configs/skills/evaluate_freeform_pair.yaml
@@ -33,7 +33,7 @@ generation: |
   [End of Question]
 
   [Start of Answer]
-  {answer}
+  {response}
   [End of Answer]
 
   Begin your evaluation by providing a short explanation. Be as objective as possible. After providing your explanation, you must rate the answer on a scale of 1 to 3 as mentioned above. 
--- a/src/instructlab/sdg/configs/skills/evaluate_grounded_pair.yaml
+++ b/src/instructlab/sdg/configs/skills/evaluate_grounded_pair.yaml
@@ -43,7 +43,7 @@ generation: |
   [End of Question]
 
   [Start of Answer]
-  {answer}
+  {response}
   [End of Answer]
 
   * Return the evaluation between [Start of Evaluation] and [End of Evaluation] tags.
--- a/src/instructlab/sdg/default_flows.py
+++ b/src/instructlab/sdg/default_flows.py
@@ -329,7 +329,7 @@ class SynthSkillsFlow(Flow):
                     "client": self.client,
                     "model_id": self.model_id,
                     "model_prompt": _get_model_prompt(self.model_family),
-                    "output_cols": ["answer"],
+                    "output_cols": ["response"],
                     "batch_kwargs": {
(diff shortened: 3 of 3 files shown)


markmc:
> The `_get_response()` method handles the differences in the expected output from the "simple" and "full" pipelines. It incorrectly assumed that the "full" pipeline provided a "response" column, but it is actually "answer".

I'll admit to being thoroughly confused, but ...

Are you sure it's not "response" for the full knowledge pipeline?

Where exactly in the code do you even check this for sure? I'm looking at `output_cols` in the `LLMBlock` configs ...

russellb:
> > The `_get_response()` method handles the differences in the expected output from the "simple" and "full" pipelines. It incorrectly assumed that the "full" pipeline provided a "response" column, but it is actually "answer".
> 
> I'll admit to being thoroughly confused, but ...
> 
> Are you sure it's not "response" for the full knowledge pipeline?

Yeah ...

> Where exactly in the code do you even check this for sure? I'm looking at `output_cols` in the `LLMBlock` configs ...

Yeah, it's in there ... but not at the end like you might expect.

https://github.com/instructlab/sdg/blob/bcb79740ab3817899895591b236fe9246ede56fc/src/instructlab/sdg/default_flows.py#L153

It's at the beginning. The rest of the pipeline is checking it to potentially be discarded, but the columns we care about in what makes it out the other end come from there.

markmc:
> > > The `_get_response()` method handles the differences in the expected output from the "simple" and "full" pipelines. It incorrectly assumed that the "full" pipeline provided a "response" column, but it is actually "answer".
> > 
> > 
> > I'll admit to being thoroughly confused, but ...
> > Are you sure it's not "response" for the full knowledge pipeline?
> 
> Yeah ...
> 
> > Where exactly in the code do you even check this for sure? I'm looking at `output_cols` in the `LLMBlock` configs ...
> 
> Yeah, it's in there ... but not at the end like you might expect.
> 
> https://github.com/instructlab/sdg/blob/bcb79740ab3817899895591b236fe9246ede56fc/src/instructlab/sdg/default_flows.py#L153
> 
> It's at the beginning. The rest of the pipeline is checking it to potentially be discarded, but the columns we care about in what makes it out the other end come from there.

but ... that's `response` not `answer`?

markmc:
Cool, lgtm :+1:

russellb:
I changed this PR completely.

Now it addresses the real problem, which is that the knowledge vs skills pipelines had different outputs.  The original PR here fixed skills, but would have broken knowledge.

I'm still trying to get CI in place that can catch this stuff ...

oindrillac (approved):
lgtm

aakankshaduggal (approved):
Tested and this works, thanks @russellb

PR #77, opened 2024-07-03
Properly support batched/non-batched with vllm/llama.cpp 

njhill:
This is based on @npalaska's PR #58.

With these changes we will auto-detect whether the server supports batched inputs and if not will send them sequentially.


src/instructlab/sdg/llmblock.py
```diff
@@ -84,27 +90,40 @@ def _parse(self, generated_string) -> dict:
...
+            )
+            return [choice.text.strip() for choice in response.choices]
+
+        n = gen_kwargs.get("n", 1)
+        results = []
+        for prompt in prompts:
+            for _ in range(n):
+                response = self.client.completions.create(
+                    prompt=prompt, **generate_args
+                )
```
npalaska, line 115: We could rewrite this as 
```
responses = [
    self.client.completions.create(prompt=prompt, **generate_args)
    for prompt in prompts
    for _ in range(n)
]
```
wdyt?

src/instructlab/sdg/llmblock.py
```diff
@@ -167,29 +194,33 @@ def __init__(
...
+
+        return self.prompt_template.format(**sample).strip()
 
     def validate(self, prompt_template: str, input_dict: Dict[str, Any]) -> bool:
         if isinstance(prompt_template, dict):
             prompt_template = prompt_template[input_dict[self.selector_column_name]]
         return super()._validate(prompt_template, input_dict)
+
+
+def server_supports_batched(client, model_id: str) -> bool:
```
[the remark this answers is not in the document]
njhill: I think batched is better.. since it's referring to the inputs. Even without batched inputs it might do batching internally.

src/instructlab/sdg/llmblock.py
njhill: yes but then we would require an additional loop anyhow

src/instructlab/sdg/llmblock.py
```diff
@@ -113,24 +132,32 @@ def generate(self, samples, **gen_kwargs) -> Dataset:
...
+            extended_samples.extend([item] * num_parallel_samples)
+
+        print(f"num outputs is {len(outputs)}")
         new_data = []
-        for sample, output in zip(samples, outputs):
+        for sample, output in zip(extended_samples, outputs):
             parsed_outputs = self._parse(output)
-            # pylint: disable=consider-using-generator
+
             max_length = max([len(value) for value in parsed_outputs.values()])
```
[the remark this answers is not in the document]
njhill: This logic still needs to be cleaned up anyhow I think, it's not doing what it was intended to

njhill:
Thanks @npalaska I addressed most of those comments.

src/instructlab/sdg/llmblock.py
npalaska: Yeah

src/instructlab/sdg/llmblock.py
```diff
@@ -45,8 +46,13 @@ def __init__(
             "model": self.model,
             "temperature": 0,
             "max_tokens": 12000,
+            #"seed": 12345,  TBD
         }
 
+        # Whether the LLM server supports a list of input prompts
+        # and supports the n parameter to generate n outputs per input
+        self.server_supports_batched = server_supports_batched(client, model_id)
```
markmc, line 54: The `FlowParams` in #64 would give us a place to do this once rather than for every `LLMBlock`, but that can be fixed up later

src/instructlab/sdg/llmblock.py
```diff
@@ -45,8 +46,13 @@ def __init__(
             "model": self.model,
             "temperature": 0,
             "max_tokens": 12000,
+            #"seed": 12345,  TBD
```
markmc, line 49: Delete? Or add an explanation to the comment

src/instructlab/sdg/llmblock.py
```diff
@@ -84,27 +90,40 @@ def _parse(self, generated_string) -> dict:
...
-        return [choice.text.strip() for choice in response.choices]
+        generate_args = {**self.defaults, **gen_kwargs}
+
+        if self.server_supports_batched:
+            response = self.client.completions.create(
+                prompt=prompts, **generate_args
+            )
+            return [choice.text.strip() for choice in response.choices]
+
+        n = gen_kwargs.get("n", 1)
```
markmc, line 109: I would have imagined doing this in reverse - including "num_instructions_to_generate" in the block config and adding 'n' to gen_kwargs if batching was supported. No biggie though

src/instructlab/sdg/llmblock.py
```diff
@@ -113,21 +132,30 @@ def generate(self, samples, **gen_kwargs) -> Dataset:
         # validate each sample
         for sample in samples:
             if not self._validate(self.prompt_template, sample):
-                return None
+                logger.warning("Sample failed validation")  #TODO add details
+                #TODO remove sample from samples
```
markmc, line 136: Hmm. Should this be in a separate PR. If in this PR, the TODO should be resolved?

src/instructlab/sdg/llmblock.py
```diff
@@ -113,21 +132,30 @@ def generate(self, samples, **gen_kwargs) -> Dataset:
...
-        if batched:
-            outputs = self._generate(samples, **gen_kwargs)
-        else:
-            outputs = [self._generate([sample], **gen_kwargs)[0] for sample in samples]
-        logger.debug("Generated outputs: {}".format(outputs))
+
+        outputs = self._generate(samples, **gen_kwargs)
+        logger.debug("Generated outputs: %s", outputs)
+
+        num_parallel_samples = gen_kwargs.get("n", 1)
```
markmc, line 146: Hmm, and here's a reason to make num_parallel_samples part of the block config ... and add 'n' to gen_kwargs based on that

src/instructlab/sdg/llmblock.py
```diff
@@ -167,29 +195,33 @@ def __init__(
...
+        # Make a test call to the server to determine whether it supports
+        # multiple input prompts per request and also the n parameter
+        response = client.completions.create(
+            model=model_id, prompt=["test1", "test2"], max_tokens=1, n=3
+        )
+        # Number outputs should be 2 * 3 = 6
+        supported = len(response.choices) == 6
+    except openai.InternalServerError:
+        supported = False
+    client.server_supports_batched = supported
```
markmc, line 225: Hmm, I understand that you want to cache this ... but I don't like setting a new attribute on a class we don't own

I guess this could be removed with a move to `FlowParams`

markmc (changes requested):
Cool, looks like a great direction

At least resolve the `#TODO remove sample from samples` thing

markmc:
Also, please squash those fixup commits as per instructlab/dev-docs#110

src/instructlab/sdg/llmblock.py
markmc: See `PipelineContext` in #86 now

aakankshaduggal:
Closing in favor of #105

PR #78, opened 2024-07-04
Handle type conversion errors in FilterByValueBlock

russellb:
97d3fcd filterblock: Don't break when filtering on unexpected data
111711f Add unit tests for sdg.filterblock

commit 97d3fcd6513320edf4195ba62940af3eb79ffd8f
Date:   Wed Jul 3 19:53:23 2024 -0400

    filterblock: Don't break when filtering on unexpected data
    
    The previous code in this block did filtering assuming that all
    samples had a value that was correct for the type. For example, when
    filtering on an integer value, it assumed every row had a valid
    integer, where it may instead have garbage.
    
    This change introduces a new helper, _convert_dtype(), which properly
    handles this condition. When the conversion fails on a `ValueError`
    exception, it treats it as `None` instead of allowing the exception to
    be raised up to the caller.
    
    The fix was authored by Shiv in PR #72. I only pulled it out into a
    standalone commit.
    
commit 111711fa419b53e2f31b63bcb91f829eacd55548
Date:   Wed Jul 3 19:56:45 2024 -0400

    Add unit tests for sdg.filterblock
    
    This new file introduces some unit tests for the filterblock module.
    A previous commit introduced a bug fix to ensure that unexpected data
    does not cause this block to break. These tests exercise those fixes.
    The tests introduced here will run automatically in CI and help ensure
    that the problem does not reoccur.
    
### DIFF
--- a/tests/test_filterblock.py
+++ b/tests/test_filterblock.py
@@ -1,0 +1,31 @@
+# Standard
+from unittest.mock import patch
+import operator
+import unittest
+
+# Third Party
+from datasets import Dataset, Features, Value
+
+# First Party
+from instructlab.sdg.filterblock import FilterByValueBlock
+
+
+class TestFilterByValueBlock(unittest.TestCase):
+    def setUp(self):
+        self.block = FilterByValueBlock(
+            filter_column="age",
+            filter_value=30,
+            operation=operator.eq,
+            convert_dtype=int,
+        )
+        self.dataset = Dataset.from_dict(
+            {"age": ["25", "30", "35", "forty", "45"]},
+            features=Features({"age": Value("string")}),
+        )
+
+    @patch("instructlab.sdg.filterblock.logger")
+    def test_generate_mixed_types(self, mock_logger):
+        filtered_dataset = self.block.generate(self.dataset)
+        self.assertEqual(len(filtered_dataset), 1)
+        self.assertEqual(filtered_dataset["age"], [30])
+        mock_logger.error.assert_called()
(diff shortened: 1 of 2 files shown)


russellb:
I posted this as part of a discussion thread here: https://github.com/instructlab/sdg/pull/72#discussion_r1664920262

tests/test_filterblock.py
```diff
@@ -0,0 +1,41 @@
...
+            convert_dtype=int,
+        )
+        self.dataset = Dataset.from_dict(
+            {"age": ["25", "30", "35", "forty", "45"]},
+            features=Features({"age": Value("string")}),
+        )
+
+    def test_convert_dtype_success(self):
+        sample = {"age": "25"}
+        converted_sample = self.block._convert_dtype(sample)
```
markmc, line 28: FWIW I think I would have been inclined to not test the internal `_convert_dtype()` but instead exercise it just through Block.generate()

markmc:
Great, very clear. Well done on adding tests. lgtm

tests/test_filterblock.py
russellb: Makes sense. The last case at the end does that. I could just remove these tests of the internal thing. It’d be less code but we still have the same test coverage.

russellb: I removed the tests of the internal method as suggested, since they don't actually add additional test coverage

russellb:
CI is passing, it was approved previously as part of #72, and @markmc reviewed this version that split it out and added tests, so I'm going to merge

PR #80, opened 2024-07-04
Strip html tags from knowledge document

vincentywdeng:


russellb (commented):
Isn't the only format we support right now Markdown?

and a markdown file could definitely have some HTML in code blocks and I don't think we'd want to strip that.

Overall though, it's not super clear what problem was observed that this is fixing?

vincentywdeng:
Hi, 
The intention for this PR is try to mitigate this issue https://github.com/instructlab/instructlab/issues/1319. The root cause is if the knowledge document is large, we split it into multiple chucks and randomly choose one to generate the traning data which may not be correspond to the seed sample question. 

We definitely need a way to choose corresponding chuck to match sample question, e.g. based on similarity between seed question and chunk. Before that, I believe it is still a good idea to trim out html tags that doesn't contribute the data generation, which can reduce the number of chunk and make each chunk embedd more information to generate data.

russellb:
> Hi, The intention for this PR is try to mitigate this issue [instructlab/instructlab#1319](https://github.com/instructlab/instructlab/issues/1319). The root cause is if the knowledge document is large, we split it into multiple chucks and randomly choose one to generate the traning data which may not be correspond to the seed sample question.

That's not quite current with the latest code. It's effectively been completely rewritten. We will run the pipeline over ALL chunks of the document.

> We definitely need a way to choose corresponding chuck to match sample question, e.g. based on similarity between seed question and chunk. Before that, I believe it is still a good idea to trim out html tags that doesn't contribute the data generation, which can reduce the number of chunk and make each chunk embedd more information to generate data.

I still don't think we can blindly strip HTML. I think that's making some assumptions about what's inside the doc.

vincentywdeng:
Cool! Let me check it out

russellb:
For now I'm going to close this out, because I don't think the change is appropriate as-is. Feel free to re-open, or use an issue to continue the discussion if needed.

PR #81, opened 2024-07-04
Allow FilterByValueBlock to handle one or many values

russellb:
97d3fcd filterblock: Don't break when filtering on unexpected data

commit 6c1f3eee12d47bf4be7fa23282732d3c8a143028
Date:   Wed Jul 3 21:28:56 2024 -0400

    Allow FilterByValueBlock to handle one or many values
    
    This block previously only accepted a single value to filter on. This
    update makes it handle a list, as well. In that case, it will ensure
    that the filter matches one of the values in the list.
    
### DIFF
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -13,8 +13,21 @@ class FilterByValueBlock(Block):
     def __init__(
         self, filter_column, filter_value, operation, convert_dtype=None, **batch_kwargs
     ) -> None:
+        """
+        Initializes a new instance of the FilterByValueBlock class.
+
+        Parameters:
+        - filter_column (str): The name of the column in the dataset to apply the filter on.
+        - filter_value (any or list of any): The value(s) to filter by.
+        - operation (callable): A function that takes two arguments (column value and filter value) and returns a boolean indicating whether the row should be included in the filtered dataset.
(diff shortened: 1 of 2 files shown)


src/instructlab/sdg/filterblock.py
```diff
@@ -37,6 +37,13 @@ def generate(self, samples) -> Dataset:
                 num_proc=self.num_procs,
             )
 
+        if isinstance(self.value, list):
```
markmc, line 40: Since we all found this stuff some confusing, simplifying like this might help:

```
class FilterByValueBlock(Block):
    def __init__(...):
        self.value = filter_value
        if not isinstance(self.value, list):
            self.value = [self.value]
```

and then just have one filter code path

markmc:
Great. Nice and simple, no confusion. lgtm

src/instructlab/sdg/filterblock.py
russellb: Good suggestion. That sounds nicer / smaller / clearer

russellb: applied this suggestion, thanks!

russellb:
I rebased this and marked it ready for review now that the pre-req PR merged.

markmc:
lgtm :+1:

russellb:
> lgtm 👍

thanks for the review!

PR #84, opened 2024-07-04
Fix reuse of pipelines including IterBlock

russellb:
When running `sdg.generate()` more than once, a pipeline including an
`IterBlock` would fail because the `block_kwargs` were removed the
first time the block was used. This change removes the special
handling for `IterBlock` because it does not appear to be necessary.

The new code ends up calling the block's constructor with all of the
same arguments, but now `block_kwargs` doesn't disappear.

### DIFF
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -3,7 +3,6 @@
 from datasets import Dataset
 
 # Local
-from .iterblock import IterBlock
 from .logger_config import setup_logger
 
 logger = setup_logger(__name__)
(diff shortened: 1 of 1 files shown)


markmc:
You've also fixed the fact that we were instantiating the blocks twice:

```
     block = block_type(**block_config)
     if block_type == IterBlock:
         ...
     else:
         block = block_type(**block_config)
```

I did wonder about this special handling of `IterBlock`, but it's all confusing enough I couldn't quickly convince myself that it wasn't needed. I sort of looks like a leftover from an earlier implementation where the IterBlock's config just got passed down to the LLMBlock with block_kwargs added? Or something

Anyhow, lgtm :+1:

markmc:
Some special handling of IterBlock turns out to be needed in #86 ... so #86 should replace this PR, at least how it's currently implemented

russellb:
> Some special handling of IterBlock turns out to be needed in #86 ... so #86 should replace this PR, at least how it's currently implemented

marked this as a draft while we work through #86. This will help remind me to come back around and check if it's still relevant later.

PR #85, opened 2024-07-04
generate_data: fix support for multiple leaf nodes

russellb:
When running generate_data() with a taxonomy with more than one leaf
node, the code previously appended the resulting datasets as if they
behaved like a Python list. They are a `datasets.Dataset` and do not
support the `+` operator.

Instead, keep a list of these datasets. Also update the code that
writes these results to a file to handle the extra level of iteration
now required.

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -78,20 +78,21 @@ def _get_response(logger, synth_example):
     return parts[1].strip() if len(parts) == 2 else parts[0].strip()
 
 
-def _gen_train_data(logger, machine_instruction_data, output_file_train):
+def _gen_train_data(logger, output_datasets, output_file_train):
     train_data = []
-    for synth_example in machine_instruction_data:
-        logger.debug(synth_example)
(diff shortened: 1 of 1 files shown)


markmc:
Makes sense, :+1: 

Do you understand why e2e was passing with this bug?

russellb:
> Makes sense, 👍
> 
> Do you understand why e2e was passing with this bug?

Yes. It tested all taxonomy types, but one at a time. I ran it with two by accident locally and hit this. Test coverage improvement is needed. I will file an issue to fix this.

russellb:
Test coverage issue to fix e2e. Easy fix, should be able to do it tomorrow https://github.com/instructlab/instructlab/issues/1599

aakankshaduggal (commented):
Thanks for the PR @russellb 
Changes look good but looks like merging #89 has caused merge conflicts, can you PTAL?

russellb:
> Thanks for the PR @russellb Changes look good but looks like merging #89 has caused merge conflicts, can you PTAL?

rebased on `main`

PR #86, opened 2024-07-05
Add a YAML based file format for pipelines

markmc:
See instructlab/dev-docs#109

In order to support custom pipelines, add a YAML based file format.
    
However, to make the default pipelines easier to reason about and develop,
also convert them to the YAML file format.
   
This changes the top-level API from:
    
```
mmlu_block_configs = MMLUBenchFlow().get_flow()
knowledge_block_configs = SynthKnowledgeFlow().get_flow()
knowledge_pipe = Pipeline(ctx, mmlu_flow + knowledge_flow)
```
    
to:
    
```
knowledge_pipe = Pipeline.from_flows(
    ctx, [pipeline.MMLU_BENCH_FLOW, pipeline.SYNTH_KNOWLEDGE_FLOW]
)
```


russellb:
I looked at the rebase. It's mostly because of #82, but a little bit from #78.

I pushed a rebase here - https://github.com/russellb/sdg/tree/pr-86-rebase - not pushing to the PR branch since I haven't reviewed the code yet to know that I got it right

russellb (commented):
This is looking really nice!

I have one question about the PipelineContext. This moves the teacher model_id to the pipeline context, though one feature we want to support is having a different model id on a per LLMBlock basis. This indeed is not needed by any of our current flows in the tree, but we have a need for it in a downstream pipeline. We want to support having vllm running a model, as well as a model + an adapter. Those two cases will be served up under different model ids.

The simplest option seems to be to push that field back to be per-LLMBlock. Let me know if I've missed another way to accomplish this.

The other big thing is that I'm eager to run this in a CI job that can test the full pipeline. I've been working on that a lot in the last week. I think I'm close.

markmc:
> I looked at the rebase. It's mostly because of #82, but a little bit from #78.
> 
> I pushed a rebase here - https://github.com/russellb/sdg/tree/pr-86-rebase - not pushing to the PR branch since I haven't reviewed the code yet to know that I got it right

Thanks, I took that and rebased again.

commit 5c46881 is significant - relates to #82

oindrillac (commented):
I like the general direction but some of the designs being changed, such as the ones here (yaml vs py) were recommended design proposals we previously agreed upon or adapted our designs to from a previous format here [#98](https://github.com/instructlab/dev-docs/pull/98).

At this point, I would prioritize core functionality changes over major changes to the library design, especially since this blocks and delays the required features. I suggest going through a dev-docs proposal for such changes.

markmc:
> I like the general direction but some of the designs being changed, such as the ones here (yaml vs py) were recommended design proposals we previously agreed upon or adapted our designs to from a previous format here [#98](https://github.com/instructlab/dev-docs/pull/98).
> 
> At this point, I would prioritize core functionality changes over major changes to the library design, especially since this blocks and delays the required features. I suggest going through a dev-docs proposal for such changes.

Yes, see instructlabs/dev-docs#109 for that design discussion

My understanding is that `custom flows` is a "required feature" - see the "Alternative Approaches" section in the dev-doc for some of the thinking on that

See also instructlab/instructlab#1546

markmc:
> I have one question about the PipelineContext. This moves the teacher model_id to the pipeline context, though one feature we want to support is having a different model id on a per LLMBlock basis. This indeed is not needed by any of our current flows in the tree, but we have a need for it in a downstream pipeline. We want to support having vllm running a model, as well as a model + an adapter. Those two cases will be served up under different model ids.

It's important to consider we have 2 personas - the author of a pipeline config, and the user executing the pipeline (e.g. with `ilab data generate`

Using the model name/id in the pipeline config is a problem because user can choose a different model and so we can't require the pipeline author to have knowledge of it

I'm trying to avoid going do a path where the pipeline author has to do some templating to take into account user-defined runtime parameters

I think the use case here is likely to be:

1. Most `LLMBlock` definitions will use the default teacher model - and we can make the semantic that if the pipeline author doesn't specify a model, the default is used. And we get the default model name from the `PipelineContext`
2. In cases where a model+adapter is served, it hopefully will be the same pipeline author defining both its serving configuration and its use in an `LLMBlock`. So the pipeline author can name it in the serving config and reference it by that name in the pipeline config

We can add support for (2) when we add model serving config support

Does that make sense?

> The simplest option seems to be to push that field back to be per-LLMBlock. Let me know if I've missed another way to accomplish this.

Thank you - I've added [a comment in the dev-doc](https://github.com/instructlab/dev-docs/pull/109/files#r1668891096) as a todo to capture the above detail

russellb:
@markmc thanks, that mostly makes sense, but one question

> We can add support for (2) when we add model serving config support

but how about someone working on a downstream pipeline already making use of different model IDs per block? I don't want to break them. Do you mean we'd implement this before merging?

markmc:
> @markmc thanks, that mostly makes sense, but one question

I may be totally missing something!

> > We can add support for (2) when we add model serving config support
> 
> but how about someone working on a downstream pipeline already making use of different model IDs per block? I don't want to break them. Do you mean we'd implement this before merging?

I guess I'd like to know how those non-default model IDs are currently defined ... oh is this assuming someone running vLLM manually?

In that case, I think the pipeline author needs to be able to define required models and the user needs some way to specify a mapping of model IDs to those required models

Something like:

```
version: "1.0"
models:
  - name: myfunkyadaptor
    description: a funky adaptor for generating questions
block_configs:
  - block_type: LLMBlock
    block_config:
      block_name: gen_questions
      config_path: configs/skills/freeform_questions.yaml
      add_num_samples: True
      model: myfunkyadatpor
      output_cols:
        - question
    drop_duplicates:
      - question
```

and then something like:

```
$ ilab data generate --model=myfunkyadaptor:model1
```

Or just require the user to ensure the model is called `myfunkyadatpor`

The crucial point IMO is ... the pipeline author has to define their expectations ... right?

markmc:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9847931828), please investigate.

```
TypeError: FilterByValueBlock.__init__() got an unexpected keyword argument 'block_name'
```

Bah .. simple fix, but getting a unit test to catch it took a bit of futzing. I added a test which parses all the pipeline yamls and at least instantiates all the blocks.

markmc:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9860564906), please investigate.

```
  File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/pipeline.py", line 78, in generate
    dataset = block.generate(dataset, **gen_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/filterblock.py", line 90, in generate
    return samples.filter(
           ^^^^^^^^^^^^^^^
...
    rv = reduce(self.proto)
         ^^^^^^^^^^^^^^^^^^
TypeError: cannot pickle 'SSLContext' object
```

I think I've reproduced locally and its specific to the full freeform skills pipeline. Investigating.

markmc:
> > e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9860564906), please investigate.
> 
> ```
>   File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/pipeline.py", line 78, in generate
>     dataset = block.generate(dataset, **gen_kwargs)
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/filterblock.py", line 90, in generate
>     return samples.filter(
>            ^^^^^^^^^^^^^^^
> ...
>     rv = reduce(self.proto)
>          ^^^^^^^^^^^^^^^^^^
> TypeError: cannot pickle 'SSLContext' object
> ```
> 
> I think I've reproduced locally and its specific to the full freeform skills pipeline. Investigating.

It only happens with `num_proc` set in the call to `samples.filter()`. Not sure what's happening, since we use `num_procs=8` in `main` here. Will come back to it tomorrow.

tiran:
> > e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9860564906), please investigate.
> 
> ```
>   File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/pipeline.py", line 78, in generate
>     dataset = block.generate(dataset, **gen_kwargs)
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/filterblock.py", line 90, in generate
>     return samples.filter(
>            ^^^^^^^^^^^^^^^
> ...
>     rv = reduce(self.proto)
>          ^^^^^^^^^^^^^^^^^^
> TypeError: cannot pickle 'SSLContext' object
> ```
> 
> I think I've reproduced locally and its specific to the full freeform skills pipeline. Investigating.

A SSLContext cannot be pickled. I'm not familiar with the code base and the code does not have enough type annotations to understand it.

My instincts tell me that you are probably handing `PipelineContext` around. The context has an OpenAI client attribute, which itself holds a `httpx` client, which itself holds on to a SSLContext object. You probably have to make the OpenAI client attribute lazy, so it is initialized in the external process.

markmc:
> My instincts tell me that you are probably handing `PipelineContext` around. The context has an OpenAI client attribute, which itself holds a `httpx` client, which itself holds on to a SSLContext object. 

Thank you, you were absolutely spot on. See [c09d354](https://github.com/instructlab/sdg/pull/86/commits/c09d3540bc62c6e9168fbb44df89e6d2b1f1bb14)

markmc:
> In that case, I think the pipeline author needs to be able to define required models and the user needs some way to specify a mapping of model IDs to those required models

The models section really isn't necessary in retrospect. This will work:

> ```
> version: "1.0"
> models:
>   - name: myfunkyadaptor
>     description: a funky adaptor for generating questions
> block_configs:
>   - block_type: LLMBlock
>     block_config:
>       block_name: gen_questions
>       config_path: configs/skills/freeform_questions.yaml
>       add_num_samples: True
>       gen_kwargs:
>         model_id: myfunkyadatpor
>       output_cols:
>         - question
>     drop_duplicates:
>       - question
> ```

And then the serving config just needs to have `--lora-modules=myfunkyadapter=path/to/my_custom_adapter`

markmc:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9880592592), please investigate.

```
Failed to determine backend: Failed to determine whether the model is a GGUF format: [Errno 2] No such file or directory: 'models/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf'
```

russellb:
> > e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9880592592), please investigate.
> 
> ```
> Failed to determine backend: Failed to determine whether the model is a GGUF format: [Errno 2] No such file or directory: 'models/mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf'
> ```

https://github.com/instructlab/instructlab/pull/1657

markmc:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9898088541), please investigate.

```
File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/pipeline.py", line 74, in generate
    dataset = block.generate(dataset, **gen_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/instructlab/sdg/utilblocks.py", line 64, in generate
    samples = samples.map(self._generate, num_proc=self.ctx.num_procs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
  File "/actions-runner/_work/sdg/sdg/venv/lib64/python3.11/site-packages/dill/_dill.py", line 414, in save
    StockPickler.save(self, obj, save_persistent_id)
  File "/usr/lib64/python3.11/pickle.py", line 578, in save
    rv = reduce(self.proto)
         ^^^^^^^^^^^^^^^^^^
TypeError: cannot pickle 'SSLContext' object
```

Same issue as what commit 5cc6cd1 addresses, but this time it's CombineColumnsBlock - should be straightforward to fix

markmc:
FYI - the design doc (instructlab/dev-docs#109) has merged

I have also submitted a series of PRs with the goal of reducing the size of this PR. See #116 #117 #118 #119 #120 #121 #122 

I'll follow-up with an update to this PR rebased onto those PRs and reflecting the latest design in the dev-doc

markmc:
Ok, I've rebased onto #122. I haven't yet incorporated changes agreed in the design doc update.

markmc:
eh? e2e failed with:

```
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/instructlab/model/train.py", line 466, in train
    training_results_dir = linux_train(
                           ^^^^^^^^^^^^
...
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/transformers/generation/utils.py", line 2693, in _sample
    next_tokens = torch.multinomial(probs, num_samples=1).squeeze(1)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: probability tensor contains either `inf`, `nan` or element < 0
```

markmc:
Pushed some more changes, closer to the dev doc now. Still to do:

1. Allow generate_data() to take a directory path to a set of custom pipeline configs
2. Add ImportBlock to allow custom pipeline configs to inherit from the default ones
3. Implement the router concept - we'd use this to add a direct router that would allow generate_data() take a path to single pipeline config file

russellb:
> Pushed some more changes, closer to the dev doc now. Still to do:
> 
> 1. Allow generate_data() to take a directory path to a set of custom pipeline configs

draft PR for this piece: https://github.com/instructlab/sdg/pull/129

markmc:
> 2. Add ImportBlock to allow custom pipeline configs to inherit from the default ones

Done

> 3. Implement the router concept - we'd use this to add a direct router that would allow generate_data() take a path to single pipeline config file

I think this can safely wait until after this PR lands

markmc:
> > Pushed some more changes, closer to the dev doc now. Still to do:
> > 
> > 1. Allow generate_data() to take a directory path to a set of custom pipeline configs
> 
> draft PR for this piece: #129

Pulled that in, thanks

src/instructlab/sdg/flows/simple_freeform_skills.yaml
```diff
@@ -0,0 +1,13 @@
+version: "1.0"
+block_configs:
+  - block_type: LLMBlock
+    block_config:
+      block_name: gen_skill_freeform
+      config_path: configs/skills/simple_generate_qa_freeform.yaml
+      output_cols:
+        - output
+    gen_kwargs:
```
russellb: When converting the 3 `simple` flows to yaml, there's an `n` value that got lost from `gen_kwargs` here

```
                "gen_kwargs": {
                    "max_tokens": 2048,
                    "temperature": 0.7,
                    "n": self.ctx.num_instructions_to_generate,
                },
```

you mentioned that the fix is setting it as a default. in llmblock

```diff
--- a/src/instructlab/sdg/llmblock.py
+++ b/src/instructlab/sdg/llmblock.py
@@ -60,6 +60,7 @@ class LLMBlock(Block):
         self.parser_cleanup_tags = parser_kwargs.get("parser_cleanup_tags", None)
         self.defaults = {
             "model": self.ctx.model_id,
+            "n": self.ctx.num_instructions_to_generate,
             "temperature": 0,
             "max_tokens": 12000,
         }
```

src/instructlab/sdg/flows/synth_grounded_skills.yaml
```diff
@@ -0,0 +1,65 @@
...
+block_configs:
+  - block_type: LLMBlock
+    block_config:
+      block_name: gen_contexts
+      config_path: configs/skills/contexts.yaml
+      output_cols:
+        - context
+    gen_kwargs:
+      temperature: 0.7
+      max_tokens: 2048
```
russellb:
```
                "gen_kwargs": {
                    "temperature": 0.7,
                    "max_tokens": 2048,
                    "n": self.ctx.num_instructions_to_generate,
                },
```

`n` lost here, but that's covered in another comment and I think it'll be the same fix

src/instructlab/sdg/flows/synth_grounded_skills.yaml
```diff
@@ -0,0 +1,65 @@
+version: "1.0"
+block_configs:
+  - block_type: LLMBlock
```
russellb: This block needs:

```
    drop_duplicates:
      - context
```

src/instructlab/sdg/pipeline.py
```diff
@@ -69,3 +81,60 @@ def generate(self, dataset) -> Dataset:
...
+        )
+    if major <= _FLOW_PARSER_MAJOR and minor > _FLOW_PARSER_MINOR:
+        logger.warning(
+            "The custom flow file may have new features that will be ignored."
+        )
+
+    if not "block_configs" in content:
+        raise FlowParserError(
+            "The custom flow file contains no 'block_configs' section"
+        )
```
russellb: This is not a blocking thing, but a nice minor improvement to these Execption messages would be to include the specified version from the file we loaded as well as the latest version the code understands

tests/test_importblock.py
```diff
@@ -0,0 +1,103 @@
...
+
+# Third Party
+from datasets import Dataset, Features, Value
+
+# First Party
+from instructlab.sdg.importblock import ImportBlock
+from instructlab.sdg.pipeline import Pipeline
+
+
+class TestImportBlockWithMockPipeline(unittest.TestCase):
```
russellb, line 15: Thank you for adding tests!!

russellb (changes requested):
I caught a couple things lost in the yaml conversion. Other comments are non-blocking. Awesome work

src/instructlab/sdg/flows/synth_grounded_skills.yaml
russellb: resolved in this commit - https://github.com/instructlab/sdg/pull/86/commits/5a0b7a6f0e3d78dfec063b430b977601b8c54741

src/instructlab/sdg/filterblock.py
```diff
@@ -69,8 +100,10 @@ def __init__(
         super().__init__(ctx, block_name)
         self.value = filter_value if isinstance(filter_value, list) else [filter_value]
         self.column_name = filter_column
-        self.operation = operation
-        self.convert_dtype = convert_dtype
+        self.operation = _get_operator_func(operation)
+        self.convert_dtype = _get_convert_dtype(convert_dtype)
+        if self.convert_dtype:
+            self.value = [self.convert_dtype(value) for value in self.value]
```
russellb, line 106: This changed the type of the parameters passed to the constructor. I will push a commit to update the docstring that is just above this

russellb: https://github.com/instructlab/sdg/pull/86/commits/0c4cdc20e0140e87e9c1e4e892b18f2d1c157d7e

src/instructlab/sdg/llmblock.py
```diff
@@ -78,7 +78,6 @@ def __init__(
             "model": self.ctx.model_id,
             "temperature": 0,
             "max_tokens": 12000,
-            "n": self.ctx.num_instructions_to_generate,
```
russellb: quoting the commit message i made this change in ...

```
I made some past changes to how we set `n` that were not correct. The
fixes here include:

 - Re-add the one place where `n` was hard coded to 10. This was
   intentional and should be kept as-is.

 - Fix the `n` logic to be:

   - use what's specified for `n` in config if present
   - otherwise set it to 1

We never want to specify n>1 when also using a prompt that makes use
of `num_samples`, as we effectively end up with `n` * `num_samples`
results.

This restores intended behavior of the `full` pipeline, but it also
breaks applying `--num-instructions` from the CLI to be the `n` value
used with the simple pipeline. That needs to be fixed in a follow-up
commit.
```

so now I need to figure out something for the simple pipelines so that `n` is set based on runtime configuration (num_instructions_to_generate), but only for specific blocks where we want that to occur (the simple pipelines).

russellb: follow up issue filed here so it doesn't have to block merge of this if everything else is working -- https://github.com/instructlab/sdg/issues/130

russellb:
As of this commit from @xukai92 -- https://github.com/instructlab/sdg/pull/86/commits/2c527702bb8ddebfe786a6dd4eb58b815cfb44fd

The testing that compared the results with this fork (https://github.com/aakankshaduggal/sdg) was successful. @xukai92 confirmed in that that "with this we have parity" and  "https://github.com/instructlab/sdg/pull/86 should be good to merge"

I'm waiting for CI to finish one more time (both e2e jobs). If they're successful, then we can merge.

russellb (approved):
ready to merge!

Issue #60, opened 2024-07-01
Convert generated dataset to messages format 


shivchander:
Currently the generated synthetic data has the q/a/context in its own columns, the new training api assumes the datasets are formatted in [messages format](https://huggingface.co/docs/transformers/en/chat_templating)

Will need a simple util function to run post generation to handle the conversion 

Columns to have: 

```

messages: list[dict] - with roles for system, user, assistant
metadata: json str - json string version of a dictionary which acts like a catch-all for all other columns in the dataset

```

nathan-weinberg:
Eval is expecting this as well for MMLU! Tracker issue on our side: https://github.com/instructlab/eval/issues/35

russellb:
the current output is consistent with what the CLI has always done.

we can change it, but we have to coordinate it across all components, including code still in the CLI, that depends on it

shivchander:
yes I see that it converts to messages - but we still need to handle all other columns so that it doesn't break when we concatenate with other datasets during training. 

Added the required some more detail about the schema we need in my original comment

RobotSail:
The training library currently expects messages, but I believe legacy Linux training and MacOS training doesn't.

oindrillac:
Where do we see this conversion happening? Would this be in [_gen_train_data](https://github.com/instructlab/sdg/blob/main/src/instructlab/sdg/generate_data.py#L80) in generate_data.py? Will we be maintaining 2 formats, the existing one that is for qlora and the additional new format as proposed above for the full train?

aakankshaduggal:
Yes you are spot on @oindrillac 
https://github.com/instructlab/sdg/blob/45ecc73ada3d8a06b246f21ebe87b5a07b206654/src/instructlab/sdg/generate_data.py#L80-L98

But we would want to have two separate outputs, one in the format the CLI expects for legacy training, and one for the new version that would expect the messages format.

oindrillac:
cool, and based on whether `pipeline == "full"` or `pipeline == "simple"` we can enforce the final output to be in a certain required format?

aakankshaduggal:
Yes! 
@RobotSail mentioned that we just need to pass the data_file_path so for full pipeline mode we can mention the messages file instead.

russellb:
> cool, and based on whether `pipeline == "full"` or `pipeline == "simple"` we can enforce the final output to be in a certain required format?

or we could just always produce both?

nathan-weinberg:
[...]

PR #92, opened 2024-07-08
converts dataset format messages required for training

oindrillac:
This PR adds a function that converts dataset format to messages and saves an additional jsonl along with each generate and does that for both train and test dataset.

This resolves https://github.com/instructlab/sdg/issues/60

PR #94, opened 2024-07-08
converts dataset format messages required for training

oindrillac:
This PR adds a function that converts dataset format to messages and saves an additional jsonl along with each generate and does that for both train and test dataset.

This resolves https://github.com/instructlab/sdg/issues/60

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -110,24 +110,26 @@ def _gen_train_data(
 ):
     train_data = []
     messages_data = []
-    for synth_example in machine_instruction_data:
-        logger.debug(synth_example)
-        user = _get_question(logger, synth_example)
-        if len(synth_example.get("context", "")) > 0:
-            user += "\n" + synth_example["context"]
-        assistant = _unescape(_get_response(logger, synth_example))
-        train_entry = {
-            "system": _SYS_PROMPT,
-            "user": _unescape(user),
-            "assistant": assistant,
-        }
-        train_data.append(train_entry)
-        sample = {
-            "inputs": _unescape(user),
-            "targets": assistant,
-            "system": _SYS_PROMPT,
-        }
-        messages_data.append(_convert_to_messages(sample))
+
+    for output_dataset in machine_instruction_data:
+        for synth_example in output_dataset:
+            logger.debug(synth_example)
+            user = _get_question(logger, synth_example)
+            if len(synth_example.get("context", "")) > 0:
+                user += "\n" + synth_example["context"]
+            assistant = _unescape(_get_response(logger, synth_example))
+            train_entry = {
+                "system": _SYS_PROMPT,
+                "user": _unescape(user),
+                "assistant": assistant,
+            }
+            train_data.append(train_entry)
+            sample = {
+                "inputs": _unescape(user),
+                "targets": assistant,
+                "system": _SYS_PROMPT,
+            }
+            messages_data.append(_convert_to_messages(sample))
 
     with open(output_file_train, "w", encoding="utf-8") as outfile:
         for entry in train_data:
@@ -259,15 +261,17 @@ def generate_data(
 
     name = Path(model_name).stem  # Just in case it is a file path
     date_suffix = datetime.now().replace(microsecond=0).isoformat().replace(":", "_")
-    output_file_generated_train = f"generated_{name}_{date_suffix}.json"
-    output_file_messages_train = f"train_messages_{name}_{date_suffix}.jsonl"
+    output_file_messages = f"messages_{name}_{date_suffix}.json"
+    output_file_test = f"test_{name}_{date_suffix}.jsonl"
+    # train data in messages format that will be mixed and split up into train test eventually
+    output_file_train = f"train_{name}_{date_suffix}.jsonl"
 
     _gen_test_data(
         leaf_nodes,
         os.path.join(output_dir, output_file_test),
     )
 
-    logger.debug(f"Generating to: {os.path.join(output_dir, output_file_generated)}")
+    logger.debug(f"Generating to: {os.path.join(output_dir, output_file_test)}")
 
     orig_cert = (tls_client_cert, tls_client_key, tls_client_passwd)
     cert = tuple(item for item in orig_cert if item)
@@ -334,8 +338,8 @@ def generate_data(
     _gen_train_data(
         logger,
         generated_data,
-        os.path.join(output_dir, output_file_generated_train),
-        os.path.join(output_dir, output_file_messages_train),
+        os.path.join(output_dir, output_file_train),
+        os.path.join(output_dir, output_file_messages),
     )
 
     # TODO
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -105,24 +105,37 @@ def _convert_to_messages(sample):
     return sample
 
 
-def _gen_train_data(logger, output_datasets, output_file_train):
+def _gen_train_data(
+    logger, machine_instruction_data, output_file_train, output_file_messages
+):
     train_data = []
-    for output_dataset in output_datasets:
-        for synth_example in output_dataset:
-            logger.debug(synth_example)
-            user = _get_question(logger, synth_example)
-            if len(synth_example.get("context", "")) > 0:
-                user += "\n" + synth_example["context"]
-            train_data.append(
-                {
-                    "system": _SYS_PROMPT,
-                    "user": _unescape(user),
-                    "assistant": _unescape(_get_response(logger, synth_example)),
-                }
-            )
+    messages_data = []
+    for synth_example in machine_instruction_data:
+        logger.debug(synth_example)
+        user = _get_question(logger, synth_example)
+        if len(synth_example.get("context", "")) > 0:
+            user += "\n" + synth_example["context"]
+        assistant = _unescape(_get_response(logger, synth_example))
+        train_entry = {
+            "system": _SYS_PROMPT,
+            "user": _unescape(user),
+            "assistant": assistant,
+        }
+        train_data.append(train_entry)
+        sample = {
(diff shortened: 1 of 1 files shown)


src/instructlab/sdg/generate_data.py
```diff
@@ -105,27 +105,40 @@ def _convert_to_messages(sample):
...
-                user += "\n" + synth_example["context"]
-            train_data.append(
-                {
-                    "system": _SYS_PROMPT,
-                    "user": _unescape(user),
-                    "assistant": _unescape(_get_response(logger, synth_example)),
-                }
-            )
+    messages_data = []
+    for synth_example in machine_instruction_data:
```
russellb, line 113: It looks like this dropped the extra `for` loop that was added around this

src/instructlab/sdg/generate_data.py
```diff
@@ -319,16 +331,18 @@ def generate_data(
...
+    )
 
     # TODO
     # This is for backwards compatibility. The file existing previously, so we'll keep it for now.
     # I believe the github bot assumes it is present for presenting generated data to a taxonomy
     # reviewer or contributor. Otherwise, I don't see a consumer of it in this repo or the
     # `ilab` CLI.
-    _gen_train_data(
-        logger, generated_data, os.path.join(output_dir, output_file_generated)
-    )
```
russellb, line 304: As discussed in #92 , this should remain

russellb (changes requested):

src/instructlab/sdg/generate_data.py
oindrillac: changed the params so it is still saving the original file, it is called output_file_generated_train instead of output_file_messages_train.
@aakankshaduggal ran an e2e to confirm

src/instructlab/sdg/generate_data.py
oindrillac: fixed this and re introduced the for loop

aakankshaduggal (approved):
Thanks @oindrillac, tested and it gives out desired output files!

src/instructlab/sdg/generate_data.py
oindrillac: these 4 files are being generated, does that make sense @russellb 

Screenshot 2024-07-08 at 3 04 07 PM

src/instructlab/sdg/generate_data.py
```diff
@@ -78,33 +78,77 @@ def _get_response(logger, synth_example):
...
+                "system": _SYS_PROMPT,
+            }
+            messages_data.append(_convert_to_messages(sample))
 
     with open(output_file_train, "w", encoding="utf-8") as outfile:
         for entry in train_data:
             json.dump(entry, outfile, ensure_ascii=False)
             outfile.write("\n")
 
+    with open(output_file_messages, "w", encoding="utf-8") as outfile:
```
shivchander, line 139: There is a simpler way to do this, the dataset is an HF Dataset - you do something like: 

```
train_data.to_json(outfile, orient='records', lines=True)
```

shivchander (approved):

oindrillac:
The e2e is failing, is it because I changed the name of the output file? Will re-instating the name back fix it?

src/instructlab/sdg/generate_data.py
oindrillac: Right now its a list of dictionaries, but we can use HF datasets too

shivchander: Yes please, we are generating HF datasets from the sdg pipeline - doesn't make sense to convert them back to list of dicts. 

Also the `_get_train_data` and `_get_test_data` is unnecessary - there are cleaner ways to do it using `.train_test_split(test_size=<split_size>)`

russellb:
> The e2e is failing, is it because I changed the name of the output file? Will re-instating the name back fix it?

yes, changing the filenames would break it

russellb (dismissed):
The e2e failure should be resolved before merging, thanks!

russellb:
> e2e workflow succeeded on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9847939171), congrats!

just FYI -- this job runs the full pipeline, but does not run training, as we do not have the new training method (training library) working in CI yet.

oindrillac:
Fixed the file names for E2E to pass. Seems like the E2E is failing on a training step.

oindrillac:
> > e2e workflow succeeded on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9847939171), congrats!
> 
> just FYI -- this job runs the full pipeline, but does not run training, as we do not have the new training method (training library) working in CI yet.

so is the e2e expected to fail right now?

src/instructlab/sdg/generate_data.py
oindrillac: here train and test data split is not happening from the same dataset, the test data is the leaf node data that is being used for inference purposes

russellb:
> > > e2e workflow succeeded on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9847939171), congrats!
> > 
> > 
> > just FYI -- this job runs the full pipeline, but does not run training, as we do not have the new training method (training library) working in CI yet.
> 
> so is the e2e expected to fail right now?

No, it should be passing.

You can see the history of runs here: https://github.com/instructlab/sdg/actions/workflows/e2e.yml

My comment above was about the separate job that is executed on demand and runs the full pipeline. The one that runs automatically (and is failing) runs the simple pipeline.

src/instructlab/sdg/generate_data.py
oindrillac: We dont need test dataset in messages format as that split will happen after data mixing, removing the test dataset

src/instructlab/sdg/generate_data.py
oindrillac: FYI - this has changed the 3 files that will be saved are as follows:
`generated_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.json`
`test_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.jsonl`
`train_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.jsonl`

src/instructlab/sdg/generate_data.py
oindrillac: We can also handle this in a follow-up in https://github.com/instructlab/sdg/issues/100

oindrillac:
We are generating  these 3 files
```
1. generated_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.json
2. test_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.jsonl
3. train_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.jsonl
```
The e2e fails since it expects these keys in the train data for the full train 3. as well as the qlora train 1.

```
 "system": _SYS_PROMPT,
  "user": _unescape(user),
  "assistant": assistant
```
The e2e needs to be updated to validate different formats for both 1 and 3

The SDG generates
* Training dataset in a format needed for qlora train `generated_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.json`
```
 "system": _SYS_PROMPT,
  "user": _unescape(user),
  "assistant": assistant
```
* Generated dataset in the messages format needed for full train `train_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.jsonl`
```
 "inputs": _unescape(user),
"targets": assistant,
"system": _SYS_PROMPT,
```
For test
* Testing dataset used for testing old/new model behavior with the original questions `test_Mixtral-8x7B-Instruct-v0_2024-07-08T18_23_10.jsonl`
```
"system": _SYS_PROMPT,
"user": _unescape(user),
"assistant": _unescape(seed_example["output"]),
```

oindrillac:
Spoke to @RobotSail and we are not sure what the generated_*.jsonl form is being used for right now. Added a change to drop that file and keep train_ and test_ files intact to be compatible with the CLI.

oindrillac:
ah seems like that seems to fix it, the CI passes

oindrillac:
@russellb can you please take a quick look?

russellb:
> @russellb can you please take a quick look?

don't worry about that last failure. The CI job broke because of https://github.com/instructlab/instructlab/pull/1471

russellb:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9873106667), please investigate.

Ignore again. Still broken related to https://github.com/instructlab/instructlab/pull/1471

russellb:
> e2e workflow failed on this PR: [View run](https://github.com/instructlab/sdg/actions/runs/9875537673), please investigate.

sigh. continue to ignore these.

oindrillac:
I think this is ready for merge. @russellb is this good to go from your end

aakankshaduggal (approved):

russellb:
> I think this is ready for merge. @russellb is this good to go from your end

yeah, you don't need to block on me. I'll run the full sdg CI workflow again on this repo when it's fixed. Thanks for checking!

russellb:
... though I just looked and it's still removing one of the files that used to be created?  (the one with a generated_ prefix)

but maybe nothing cares

oindrillac:
yeah, it is removing that one since it was a duplicate, does not seem to be needed on the CLI from when @RobotSail and I checked, hopefully its just a duplicate

russellb:
> yeah, it is removing that one since it was a duplicate, does not seem to be needed on the CLI from when @RobotSail and I checked, hopefully its just a duplicate

I see it used in the CLI repo in `functional_tests.sh`

oindrillac:
> functional_tests.sh
Thanks for finding it. Maybe that should also use the train_.jsonl, but that should be tracked as a cli issue, I can keep it here then

russellb:
both the simple and full SDG pipelines are now passing in CI based on this PR, so I'm good with it.  thanks everyone!

russellb (approved):

RobotSail (commented):
LGTM

Issue #97, opened 2024-07-08
grounded flow not capturing context


aakankshaduggal:
e2e for grounded skills is not able to identify if it has context or not because of the wrong key name, it's looking for `context` instead of `seed_context`

russellb:
Fixed by #98

PR #98, opened 2024-07-08
Update generate_data.py to capture context key

aakankshaduggal:
The key for context is `seed_context` but we were looking for `context`, that was just resulting in all grounded skills to take the freeform workflow instead. With this change, it is getting the seed_context and using the grounded workflow when generating data.

Addresses #97

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -272,7 +272,7 @@ def generate_data(
         sdg = None
         if samples[0].get("document"):
             sdg = sdg_knowledge
-        elif samples[0].get("context"):
+        elif samples[0].get("seed_context"):
             sdg = sdg_grounded_skill
         else:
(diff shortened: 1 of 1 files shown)


russellb:
I think we should make pipelines have to explicitly define their input expectations.

1. It would be easier to integrate with. This has happened a few times.
2. We can provide much better error handling to make this sort of thing super clear when it happens.

russellb:
> I think we should make pipelines have to explicitly define their input expectations.
> 
> 1. It would be easier to integrate with. This has happened a few times.
> 2. We can provide much better error handling to make this sort of thing super clear when it happens.

to be clear, I wasn't suggesting that has to be done for this PR. It was more comments for the future (I actually thought I was looking at an issue instead of a PR at first!)

russellb (approved):
thanks!  good catch

PR #104, opened 2024-07-08
Remove two files that are now unused

russellb:
After removing the old original sdg implementation, these files are no
longer used.


danmcp (approved):

russellb:
Thanks, @danmcp !

PR #113, opened 2024-07-11
Change generate_data() to accept OpenAI client

tiran:
The `generate_data()` function now accepts a single `openai_client: openai.OpenAI` argument instead of half a dozen
parameters for endpoint and httpx. This simplifies the API and gives the caller better control over the configuration of the client, e.g. timeout, organization, and project arguments.

`generate_data` now uses a module-level logger like the rest of the code.

See instructlab/instructlab#1647 and instructlab/instructlab#1658

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Optional
 import json
+import logging
 import os
 import time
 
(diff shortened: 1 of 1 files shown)


markmc:
Unfortunately, I think we need to do this with a (short, I hope) deprecation period for the existing API - I don't think we can merge an instructlab/instructlab and instructlab/sdg change in lockstep - we have to assume folks might be using newer SDG with older instructlab, even for a short period

Can you make the OpenAI client optional and retain support for the existing args? Thanks

FWIW, I don't think `generate_data()` is the right level of abstraction for the library API - see #1519

markmc:
Note - overlaps with #114 and see also [this comment from @tiran](https://github.com/instructlab/instructlab/pull/1649#issuecomment-2221266846)

derekhiggins:
the openai client can now be passed in, https://github.com/instructlab/sdg/pull/114 
and a bug has been created to use it in the cli once available https://github.com/instructlab/instructlab/issues/1846

tiran:
Fixed by https://github.com/instructlab/sdg/pull/114

PR #115, opened 2024-07-11
Update messages datafile extension to be .jsonl

Maxusmusti:
File was being saved as `.json`, but contents are actually a list of jsons


aakankshaduggal (approved):
Thanks @Maxusmusti

russellb (approved):

PR #116, opened 2024-07-11
Replace LLMBlock model_prompt param with model_family

markmc:
In preparation for custom pipeline configuration files, do not require `model_prompt` as an `LLMBlock` param - it can have built-in knowledge of the correct prompt to use per `model_family`.

### DIFF
--- a/src/instructlab/sdg/default_flows.py
+++ b/src/instructlab/sdg/default_flows.py
@@ -10,23 +10,6 @@ from .filterblock import FilterByValueBlock
 from .llmblock import LLMBlock
 from .utilblocks import CombineColumnsBlock
 
-MODEL_FAMILY_MIXTRAL = "mixtral"
-MODEL_FAMILY_MERLINITE = "merlinite"
-
-_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST]"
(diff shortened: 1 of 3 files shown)


russellb (approved):
Thanks - I read this part before in #86  and it makes sense to me. AFAICT there's no functional impact since we have no need to pass in custom model prompts.

I understand you're trying to pull out the dynamic parts of current flow configuration.

markmc:
Pushed a very small fix to `ConditionalLLMBlock`, sorry - this type isn't actually used in the default pipeline configs

markmc:
Pushed this:

```
@@ -36,7 +36,7 @@ class _SimpleFlow(Flow):
                     "config_path": "",  # must be set by subclass
                     "client": self.client,
                     "model_id": self.model_id,
-                    "model_prompt": self.model_family,
+                    "model_family": self.model_family,
                     "output_cols": ["output"],
                 },
                 "gen_kwargs": {
@@ -330,7 +330,7 @@ class SynthGroundedSkillsFlow(Flow):
                     ),
                     "client": self.client,
                     "model_id": self.model_id,
-                    "model_prompt": self.model_family,
+                    "model_family": self.model_family,
                     "output_cols": ["context"],
                 },
                 "gen_kwargs": {
```

/me shakes fist at pesky e2e test finding hard to spot subtle bugs in a trival PR!

markmc:

PR #117, opened 2024-07-11
Add a PipelineContext class

markmc:
(This is based on #116 which needs to merge first)

First things first, the multiprocessing workarounds are horrible. Ideas welcome for how to avoid the OpenAI client getting sucked into the data serialized to map/filter workers!

49c87d5 Add a PipelineContext class
7cfbaa9 Fix multiprocessing issues in FilterByValueBlock
9d92548 Fix multiprocessing issues in utilblocks

```
commit 49c87d5

    Add a PipelineContext class
    
    In order to prepare for pipeline definitions in YAML, remove
    runtime parameters like the OpenAI client, model ID, and model
    family from the pipeline definition into a PipelineContext
    object that all blocks have access to.
```

```
commit 7cfbaa9

    Fix multiprocessing issues in FilterByValueBlock
    
    This addresses issues with using num_proc>1 with Dataset.map()
    and Dataset.filter().
    
    The first issue is:
    
      File "/usr/lib64/python3.11/pickle.py", line 578, in save
        rv = reduce(self.proto)
             ^^^^^^^^^^^^^^^^^^
    TypeError: cannot pickle 'SSLContext' object
   
    What was happening here is that the entire FilterByValueBlock
    object was being serialized to send to the multiprocessing
    worker. And now that this includes PipelineContext, which
    includes the OpenAI client object, which includes SSLContext,
    we hit a known issue: uqfoundation/dill#308
    
    The second issue is specific to map():
    
    ValueError: The features can't be aligned because the key score of features {'task_description': Value(dtype='string', id=None), 'seed_question': Value(dtype='string', id=None), 'seed_response': Value(dtype='string', id=None), 'num_samples': Value(dtype='int64', id=None), 'question': Value(dtype='string', id=None), '__index_level_0__': Value(dtype='int64', id=None), 'evaluation': Value(dtype='string', id=None), 'score': Value(dtype='string', id=None)} has unexpected type - Value(dtype='string', id=None) (expected either Value(dtype='float64', id=None) or Value("null").
    
    It appears the the datasets, only in the case of num_proc>1,
    when we hit the "error converting dtype" case and set the column
    to None, it ends up being still considered a string column rather
    than the new dtype.
    
    This second issue deserves further investigation and may require
    a fix to the datasets library.
```

```
commit 9d92548 (HEAD -> pipeline-context, markmc/pipeline-context)

    Fix multiprocessing issues in utilblocks
    
    Address the following issue with using num_proc>1 with Dataset.map():
    
    File "/usr/lib64/python3.11/pickle.py", line 578, in save
        rv = reduce(self.proto)
             ^^^^^^^^^^^^^^^^^^
    TypeError: cannot pickle 'SSLContext' object
    
    The entire block object is being serialized to sent to the
    multiprocessing worker. And now that this includes PipelineContext,
    which includes the OpenAI client object, which includes SSLContext,
    we hit a known issue: uqfoundation/dill#308
```

### DIFF
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -11,12 +11,19 @@ logger = setup_logger(__name__)
 
 class FilterByValueBlock(Block):
     def __init__(
-        self, filter_column, filter_value, operation, convert_dtype=None, **batch_kwargs
+        self,
+        ctx,
+        filter_column,
+        filter_value,
+        operation,
+        convert_dtype=None,
+        **batch_kwargs,
     ) -> None:
         """
         Initializes a new instance of the FilterByValueBlock class.
 
         Parameters:
+        - ctx (PipelineContext): A PipelineContext object containing runtime parameters.
         - filter_column (str): The name of the column in the dataset to apply the filter on.
         - filter_value (any or list of any): The value(s) to filter by.
         - operation (callable): A function that takes two arguments (column value and filter value) and returns a boolean indicating whether the row should be included in the filtered dataset.
@@ -26,7 +33,7 @@ class FilterByValueBlock(Block):
         Returns:
         None
         """
-        super().__init__(block_name=self.__class__.__name__)
+        super().__init__(ctx, block_name=self.__class__.__name__)
         self.value = filter_value if isinstance(filter_value, list) else [filter_value]
         self.column_name = filter_column
         self.operation = operation
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -9,6 +9,39 @@ from .logger_config import setup_logger
 logger = setup_logger(__name__)
 
 
+# Note - this is not a method on the class below in order to avoid
+# serializing the object itself when multi-processing is used.
+# In particular, SSLContext - embedded in the OpenAI client object -
+# cannot be pickled.
+def _filter_by_values(samples, column, op, values, num_proc=1):
+    return samples.filter(
+        lambda x: any(op(x[column], value) for value in values),
+        num_proc=num_proc,
+    )
+
+
+def _map_dtype(samples, column, dtype, num_proc=1):
+    def convert_column(sample):
+        try:
+            sample[column] = dtype(sample[column])
+        except ValueError as e:
+            logger.error(
+                "Error converting dtype: %s, filling with None to be filtered later", e
+            )
+            sample[column] = None
+        return sample
+
+    # FIXME: it appears multiprocessing map has issues with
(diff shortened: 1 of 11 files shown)


src/instructlab/sdg/filterblock.py
```diff
@@ -9,6 +9,39 @@
...
+            sample[column] = None
+        return sample
+
+    # FIXME: it appears multiprocessing map has issues with
+    # None columns. If we pass num_proc>1 here and the error
+    # case is triggered above, we get:
+    #   ValueError: The features can't be aligned ...
+    # because the column is still considered a string not
+    # the new dtype.
+    num_proc = 1
```
russellb, line 40: Thanks for the details, both in the commit message and the comment block here.  I filed an issue here to track following up on this: https://github.com/instructlab/sdg/issues/123

I wonder if setting the column to `None` just isn't a valid way to handle the case when we can't convert to the target datatype. Maybe we need a default value we set that's native to whatever we're trying to convert to. Otherwise right now we have a mix of things converted to a numeric type (in the case of the commit message) and None isn't valid for those.

Knowing what value we can use in this error case doesn't seem simple. It seems like it needs to be provided as configuration to the FilterBlock.

anyway ... issue filed to look closer

russellb (approved):
Thanks.

The biggest change seemed like a lot, but it was a lot of moving the same few pieces of data so I was able to read through it pretty quickly the way you had it in its own commit.

These multiprocessing issues seem really tricky. Thank you for the detailed explanation of what's going on, and the various comments in the code.

You noted the one bug we need to fix, but it looks like you have a workaround in place, so I don't think we need to block on it. I filed an issue here for that: https://github.com/instructlab/sdg/issues/123

src/instructlab/sdg/filterblock.py
markmc: Thanks for filing that issue - it's good practice to file an issue for any `FIXME` like that, but somehow I'm reluctant to file an issue for something that hasn't merged yet! But you're right ... I probably would never have filed the issue

On "is None valid" - see the comment I left in the issue. I think it is, but None handling was known to be buggy at one point. It feels like we've just hit a bug.

And yeah, I did consider using a different default value than None - any value of that type that is not in filter_value would work, so we could randomly choose one? Ugh.

I also considered adding an "invalid_value" field, but adding something like that to the format to work around a bug? Ugh.

```
block_type: FilterByValueBlock
   block_config:
     block_name: filter_questions
     filter_column: score
     filter_value: 1.0
     invalid_value: 0
     operation: eq
     convert_dtype: float
```

markmc: Best to continue this discussion #123, I think :+1:

markmc:

PR #118, opened 2024-07-11
Allow block_config.config_path to be relative

markmc:
(Depends on #117)

In order to remove another runtime parameter from pipeline definitions to allow us to move to using YAML files.

### DIFF
--- a/src/instructlab/sdg/block.py
+++ b/src/instructlab/sdg/block.py
@@ -14,7 +14,8 @@ logger = setup_logger(__name__)
 
 
 class Block(ABC):
-    def __init__(self, block_name: str) -> None:
+    def __init__(self, ctx, block_name: str) -> None:
+        self.ctx = ctx
         self.block_name = block_name
 
     @staticmethod
--- a/src/instructlab/sdg/block.py
+++ b/src/instructlab/sdg/block.py
@@ -3,6 +3,7 @@
 from abc import ABC
 from collections import ChainMap
 from typing import Any, Dict, Union
+import os.path
 
 # Third Party
 import yaml
@@ -42,8 +43,13 @@ class Block(ABC):
         """
         Load the configuration file for this block.
 
+        If the supplied configuration file is a relative path, it is assumed
+        to be part of this Python package.
+
         :param config_path: The path to the configuration file.
         :return: The loaded configuration.
         """
+        if not os.path.isabs(config_path):
+            config_path = os.path.join(self.ctx.sdg_base, config_path)
         with open(config_path, "r", encoding="utf-8") as config_file:
             return yaml.safe_load(config_file)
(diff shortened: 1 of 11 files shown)


src/instructlab/sdg/block.py
```diff
@@ -42,8 +43,13 @@ def _load_config(self, config_path: str) -> Union[Dict[str, Any], None]:
...
         Load the configuration file for this block.
 
+        If the supplied configuration file is a relative path, it is assumed
+        to be part of this Python package.
+
         :param config_path: The path to the configuration file.
         :return: The loaded configuration.
         """
+        if not os.path.isabs(config_path):
+            config_path = os.path.join(self.ctx.sdg_base, config_path)
```
russellb, line 53: a non-blocking improvement here would be to check if `config_path` exists prior to `open()` on the next line. If it doesn't exist, we could log a friendly message about it.

or maybe the pythonic way is to catch the exception from `open()` and log from there (and then re-raise it)

russellb (approved):
just one minor non-blocking thought. I don't know what the error would look like if the path doesn't exist. It'd be good to check and ensure it's clear enough from the CLI what was wrong. Maybe it needs a friendly log message like I suggested inline, or maybe something else ... 

this isn't super relevant until later when custom configs are supported, though. we can do some testing and see what error handling looks like once that's in place

src/instructlab/sdg/block.py
markmc: Yep, similar to Dan's error handling work in instructlab/eval#61

markmc:
> just one minor non-blocking thought. I don't know what the error would look like if the path doesn't exist. It'd be good to check and ensure it's clear enough from the CLI what was wrong. Maybe it needs a friendly log message like I suggested inline, or maybe something else ...
> 
> this isn't super relevant until later when custom configs are supported, though. we can do some testing and see what error handling looks like once that's in place

In general, I think when there is an error with a block, we don't make it easy to know which block in which pipeline caused it

This case looks like

```
  File "/home/markmc/sdg/src/instructlab/sdg/sdg.py", line 19, in generate
    dataset = pipeline.generate(dataset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/pipeline.py", line 57, in generate
    block = block_type(self.ctx, **block_config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/llmblock.py", line 66, in __init__
    self.block_config = self._load_config(config_path)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/block.py", line 54, in _load_config
    with open(config_path, "r", encoding="utf-8") as config_file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/markmc/sdg/src/instructlab/sdg/configs/skills/freeform_questions.yaml.foobar'
```

Which isn't so bad, because you just have to find the block with that path

Filed #128

markmc:

PR #119, opened 2024-07-11
Fix block_name handling

markmc:
(Depends on #118)

All Block subclasses but LLMBlock are failing to pass the block_name from block_config down to the base class, instead they are incorrectly passing the block type as its name.


russellb (approved):
thanks!

markmc:

PR #120, opened 2024-07-11
Move FilterByValue multiprocessing config to PipelineContext

markmc:
(Depends on #119)

In every use of FilterByValue in the default flows, we use batch_kwargs to set num_proc=8.
    
This doesn't appear to be a pipeline author concern, but rather a runtime parameter which should in future be based on the number of available CPUs and (perhaps) user configuration.
    
For now, just move it from batch_kwargs to PipelineContext.

### DIFF
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -1,4 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
+# Standard
+from importlib import resources
+
 # Third Party
 from datasets import Dataset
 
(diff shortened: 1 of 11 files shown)


src/instructlab/sdg/pipeline.py
```diff
@@ -20,6 +20,8 @@ def __init__(
         self.model_id = model_id
         self.num_instructions_to_generate = num_instructions_to_generate
         self.sdg_base = resources.files(__package__)
+        # FIXME: base this on the available number of CPUs
+        self.num_procs = 8
```
russellb, line 24: It would be great to make this a parameter.  `ilab data generate` already exposes this as a tunable thing.

```
$ ilab data generate --help
...
  --num-cpus INTEGER          Number of processes to use.  [default: 10]
...
```

It defaults to 10, though it should be fixed to default to the number of CPUs available. Either way, at least for our `ilab` use case, we have this passed in to `generate_data()` but we're not using it right now.

russellb (approved):
Moving this to be runtime config in the pipeline context makes sense to me.

I left a comment inline that we actually already have a user tunable value to use here. I filed an issue for it so we don't have to block on it. but also feel free to include it here. I'm OK either way.

https://github.com/instructlab/sdg/issues/124

markmc:
> Moving this to be runtime config in the pipeline context makes sense to me.
> 
> I left a comment inline that we actually already have a user tunable value to use here. I filed an issue for it so we don't have to block on it. but also feel free to include it here. I'm OK either way.
> 
> #124

Good call. Easy enough to do, but I'm trying to keep the changes to a minimum

markmc:

PR #121, opened 2024-07-11
Add `add_num_samples` to LLMBlock config

markmc:
(Depends on #120)

Two pipelines include an LLMBlock which use `{num_samples}` in their instructions to the teacher model. There needs to be some way to configure the LLMBlock so that `num_samples` will be included, but as per #82 (commit a01b04e) the value of `num_samples` should be based on the `num_instructions_to_generate` parameter.

### DIFF
--- a/scripts/test_freeform_skills.py
+++ b/scripts/test_freeform_skills.py
@@ -5,7 +5,7 @@ from openai import OpenAI
 # First Party
 from src.instructlab.sdg import SDG
 from src.instructlab.sdg.default_flows import SynthSkillsFlow
-from src.instructlab.sdg.pipeline import Pipeline
+from src.instructlab.sdg.pipeline import Pipeline, PipelineContext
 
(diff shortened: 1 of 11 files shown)


russellb (approved):
I'm OK with this as a step toward the need to pull runtime configuration out of the flow configuration.

I do wonder if it could be even simpler and `num_samples` is just always set instead of having it configured like this.  Doing it this way seems to be the safer change for the moment, since it maintains existing behavior.

markmc:
> I'm OK with this as a step toward the need to pull runtime configuration out of the flow configuration.
> 
> I do wonder if it could be even simpler and `num_samples` is just always set instead of having it configured like this. Doing it this way seems to be the safer change for the moment, since it maintains existing behavior.

Yep, this is firmly in the realm of "I don't feel safe touching this right now" for me

markmc:

PR #122, opened 2024-07-12
Remove batch_kwargs

markmc:
This appears to be unused now - now pipeline definitions include it, and it's not used in LLMBlock anywhere.

This change might throw up cases where we're passing unknown configuration to LLMBlock. Here's why ...

It's hard to spot, but this:

```    
        def __init__(self, ..., **batch_kwargs):
            ...
            self.batch_params = batch_kwargs.get("batch_kwargs", {})
```
    
is equivalent to this:

```    
        def __init__(self, ..., **kwargs):
            ...
            self.batch_params = kwargs.get("batch_kwargs", {})
```
    
which is equivalent to this:

```    
        def __init__(self, ..., batch_kwargs={}, **kwargs):
            ...
            self.batch_params = batch_kwargs
```
    
except that trailing `**kwargs` meant we were silently accepting unknown block_config parameters.

### DIFF
--- a/scripts/test_freeform_skills.py
+++ b/scripts/test_freeform_skills.py
@@ -5,7 +5,7 @@ from openai import OpenAI
 # First Party
 from src.instructlab.sdg import SDG
 from src.instructlab.sdg.default_flows import SynthSkillsFlow
-from src.instructlab.sdg.pipeline import Pipeline
+from src.instructlab.sdg.pipeline import Pipeline, PipelineContext
 
 # for vLLM endpoints, the api_key remains "EMPTY"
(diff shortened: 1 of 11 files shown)


russellb (approved):
The change looks fine, though I'd request putting the extra explanation from the PR description in the commit message, as well

markmc:
> The change looks fine, though I'd request putting the extra explanation from the PR description in the commit message, as well

Yeah, there's two commits in this PR, and the first one has that explanation. Easy to miss with the stack of PRs - see commit 18f1513897f6d31f0dd059e398305fa8792843bf

PR #129, opened 2024-07-12
generate_data: Allow pipeline arg to be a path to a directory

russellb:
On top of #86 and still needs to be tested

- **generate_data: Allow pipeline arg to be a path to a directory**

commit 1ad79e20b670fbed2cd21cf571581e55ba8c3650
Date:   Fri Jul 12 11:54:05 2024 -0400

    generate_data: Allow pipeline arg to be a path to a directory
    
    In addition to `simple`, and `full`, allow a path to a directory that
    contains the same 3 files we include in the `sdg` library for the
    built-in pipelines. This will allow use of custom pipelines instead of
    our built-in ones if desired.
    

markmc:
Add to #86

PR #132, opened 2024-07-14
Automate validation of pipeline configs

russellb:
Closes #131 

1cfd4bb Add jsonschema for validating pipeline configuration
c3bd917 Add a script that will validate pipeline configurations
18b4741 Add tox env for validating pipeline configuration
3090e65 Run pipeline config validation in CI

commit 1cfd4bbd494ca936ce7e072f498dc208b8e0dfec
Author: Russell Bryant <rbryant@redhat.com>
Date:   Sun Jul 14 12:50:45 2024 -0400

    Add jsonschema for validating pipeline configuration
    
    This commit introduces a `v1.json` jsonschema for use with validating
    pipeline configuration. This follows the same format used in the
    `instructlab/schema` repository which we use for validating taxonomy
    `qna.yaml` files.
    
    It's very likely that the schema is missing some details that are
    allowed. It does pass against all configurations currently in the
    source tree, though.
    
    Signed-off-by: Russell Bryant <rbryant@redhat.com>

commit c3bd917e55dd66f241da152d782e783385742294
Author: Russell Bryant <rbryant@redhat.com>
Date:   Sun Jul 14 12:53:43 2024 -0400

    Add a script that will validate pipeline configurations
    
    This script loads the pipeline validation schema and validates all
    configurations under `src/instructlab/sdg/pipelines/`.
    
    Signed-off-by: Russell Bryant <rbryant@redhat.com>

commit 18b47418651b8d1f3f5c000a5477894398662501
Author: Russell Bryant <rbryant@redhat.com>
Date:   Sun Jul 14 12:52:42 2024 -0400

    Add tox env for validating pipeline configuration
    
    To run the script that validates pipeline configurations, you may now
    run:
    
        tox -e validate-pipelines
    
    or
    
        make validate-pipelines
    
    Signed-off-by: Russell Bryant <rbryant@redhat.com>

commit 3090e657515a0782260644cd63ba9513fa32b920
Author: Russell Bryant <rbryant@redhat.com>
Date:   Sun Jul 14 13:00:03 2024 -0400

    Run pipeline config validation in CI
    
    Updating the existing linting job to include pipeline config
    validation, as well.
    
    Signed-off-by: Russell Bryant <rbryant@redhat.com>

PR #137, opened 2024-07-15
Update grounded_skills.yaml to add seed value

### LINKED ISSUE #108: Add seed parameter to gen context block
Suggested fix - 

add seed parameter to gen context block in the default flows.

--

**Seed**: This parameter is used to set a seed value for the random number generator used by the model. This allows you to generate reproducible results by providing the same seed value each time.

 If seed is not provided, or the model changes, or the input changes (by as much as a single token) then the result will be a completely new result. It's only there for reproducibility.

aakankshaduggal:
Resolves #108 

This commit adds seed parameter to gen_context block in the grounded skills flow.

What is a seed and why we need it?
Seed: This parameter is used to set a seed value for the random number generator used by the model. This allows you to generate reproducible results by providing the same seed value each time.

If seed is not provided, or the model changes, or the input changes (by as much as a single token) then the result will be a completely new result. It's only there for reproducibility.

### DIFF
--- a/src/instructlab/sdg/pipelines/full/grounded_skills.yaml
+++ b/src/instructlab/sdg/pipelines/full/grounded_skills.yaml
@@ -10,6 +10,7 @@ blocks:
       temperature: 0.7
       max_tokens: 2048
       n: 10
+      seed: 42
     drop_duplicates:
       - context
   - name: gen_grounded_questions
--- a/src/instructlab/sdg/pipelines/schema/v1.json
+++ b/src/instructlab/sdg/pipelines/schema/v1.json
@@ -34,6 +34,7 @@
           },
           "gen_kwargs": {
             "type": "object",
+            "additionalProperties": false,
             "properties": {
               "model": {
                 "type": "string"
@@ -45,6 +46,9 @@
                 "type": "number"
               },
               "n": {
+                "type": "number"
+              },
+              "seed": {
                 "type": "number"
               },
               "extra_body": {


src/instructlab/sdg/pipelines/full/grounded_skills.yaml
```diff
@@ -10,6 +10,7 @@ blocks:
       temperature: 0.7
       max_tokens: 2048
       n: 10
+      seed: 42
```
russellb, line 13: I think you'll have to update the schema we just merged in #132 to allow this new parameter.  Try running `make validate-pipelines`

russellb (changes requested):
One tweak needed -- try rebasing this on `main` and running `make validate-pipelines`. We need to add `seed` to the schema to reflect this as an allowed parameter.

russellb:
@mergifyio rebase

russellb:
noticed the CI job that should fail because of the schema validation didn't run here -- fix for that is here -- https://github.com/instructlab/sdg/pull/140

src/instructlab/sdg/pipelines/full/grounded_skills.yaml
aakankshaduggal: I did that and looks like it passed. Is there anything else that I need to do here?

russellb: Sorry, the schema wasn't set properly to catch this.  See this commit - https://github.com/instructlab/sdg/pull/137/commits/94b3c0a3145d6e830a6577a9b0848b6eec5e040e

If `additionalParameters: false` had been set, this change would have failed validation. I fixed that now, though!

russellb (approved):
this lgtm now - i added a commit to update the pipeline config schema

src/instructlab/sdg/pipelines/full/grounded_skills.yaml
russellb: @aakankshaduggal since I added a commit to the PR, will you review and approve as well that it looks OK to you?

aakankshaduggal (commented):
lgtm

PR #142, opened 2024-07-15
Drop `__index_level_0__` columns 

### LINKED ISSUE #109: Drop duplicate columns like `__index_level_0__`
We are generating a column `__index_level_0__` in some gen blocks, we need to ensure that we drop them after they are used to avoid redundancy.

aakankshaduggal:
resolves #109 

Dropping duplicate columns like` __index_level_0__`, this column gets added at different steps and cause redundancy and confusion due to the same key name. 

Dropping this key after `gen_questions` resolves the issue.

### DIFF
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -57,8 +57,9 @@ class Pipeline:
         Drop duplicates from the dataset based on the columns provided.
         """
         df = dataset.to_pandas()
-        df.drop_duplicates(subset=cols, inplace=True)
-        return Dataset.from_pandas(df)
+        df = df.drop_duplicates(subset=cols).reset_index(drop=True)
+        ds = Dataset.from_pandas(df)
+        return ds
 
     def generate(self, dataset) -> Dataset:
         """


markmc:
I feel like we're working around something here that can be fixed at the source of the issue - where is this column coming form and why?

Is it coming from `drop_duplicates()` ?

```
    def _drop_duplicates(self, dataset, cols):
        """                                                                                                                                                                                                     
        Drop duplicates from the dataset based on the columns provided.                                                                                                                                         
        """
        df = dataset.to_pandas()
        df.drop_duplicates(subset=cols, inplace=True)
        return Dataset.from_pandas(df)
```

I've tested quickly here, and it seems that add `preserve_index=False` to `Dataset.from_pandas()` fixes the issue

See https://huggingface.co/docs/datasets/v2.20.0/en/package_reference/main_classes#datasets.Dataset.from_pandas

Could you try this out and update the PR to use this approach if it works?

markmc (changes requested):
Let's try using `from_pandas(preserve_index=False)`

aakankshaduggal:
Thanks @markmc for the suggestion, but my suspect is that it wouldn't work as above because we are referring to drop_columns and not drop_duplicates.

https://github.com/instructlab/sdg/blob/fa3603e7fd64168d43c8441b77d271d5657773ad/src/instructlab/sdg/pipeline.py#L86

markmc:
I'm saying these lines:

```
    drop_duplicates:
      - question
```

are what is introducing this index column. And we can fix this code so the column is never created. Please try my suggestion

shivchander:
> Let's try using `from_pandas(preserve_index=False)`

Yep - this seems to show up when we convert datasets to dataframes and vice versa. We could do one of the two: 

### 1. Reset Index Before Conversion


```python
# add this before converting to ds
df = df.reset_index(drop=True)
ds = Dataset.from_pandas(df)
```

### 2. Avoiding Index in Conversion
```python
ds = Dataset.from_pandas(df, preserve_index=False)
```

aakankshaduggal:
Thanks @markmc and @shivchander -- I will test this and update the PR

markmc (approved):

shivchander (approved):

Issue #123, opened 2024-07-12
Investigate Dataset.map() multiprocessing failure


russellb:
This issue is called out in one of the commits in PR #117 

> The second issue is specific to map():
> 
> ```
> ValueError: The features can't be aligned because the key score of features {'task_description': Value(dtype='string', id=None), 'seed_question': Value(dtype='string', id=None), 'seed_response': Value(dtype='string', id=None), 'num_samples': Value(dtype='int64', id=None), 'question': Value(dtype='string', id=None), '__index_level_0__': Value(dtype='int64', id=None), 'evaluation': Value(dtype='string', id=None), 'score': Value(dtype='string', id=None)} has unexpected type - Value(dtype='string', id=None) (expected either Value(dtype='float64', id=None) or Value("null").
> ```
> 
> It appears the the datasets, only in the case of num_proc>1,
> when we hit the "error converting dtype" case and set the column
> to None, it ends up being still considered a string column rather
> than the new type.
> 
> This second issue deserves further investigation and may require
> a fix to the datasets library.

The related code in filterblock.py as of that PR is:

```python
def _map_dtype(samples, column, dtype, num_proc=1):
    def convert_column(sample):
        try:
            sample[column] = dtype(sample[column])
        except ValueError as e:
            logger.error(
                "Error converting dtype: %s, filling with None to be filtered later", e
            )
            sample[column] = None
        return sample

    # FIXME: it appears multiprocessing map has issues with
    # None columns. If we pass num_proc>1 here and the error
    # case is triggered above, we get:
    #   ValueError: The features can't be aligned ...
    # because the column is still considered a string not
    # the new dtype.
    num_proc = 1

    return samples.map(convert_column, num_proc=num_proc)
```

We need to investigate this error more deeply to figure out the best fix

markmc:
For reference, the type conversion failure handling case was introduced in #78

markmc:
A tiny bit of searching led me to stuff like huggingface/datasets#3195 and huggingface/datasets#3676 - giving the impression that None is supposed to be handled, but it wouldn't surprise anyone if there were still latent bugs in None handling

markmc:
From @russellb 

> Maybe we need a default value we set that's native to whatever we're trying to convert to. Otherwise right now we have a mix of things converted to a numeric type (in the case of the commit message) and None isn't valid for those.
>
> Knowing what value we can use in this error case doesn't seem simple. It seems like it needs to be provided as configuration to the FilterBlock.

[...]

PR #143, opened 2024-07-15
filterblock: add default_value for use with convert_dtype

### LINKED ISSUE #133: Improve FilterByValueBlock invalid value error handling
In #110 we see:

```
    self.operation(x[self.column_name], value) for value in self.value
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>=' not supported between instances of 'NoneType' and 'float'
```

The issue here is that `convert_dtype` couldn't handle the value in the column, so we set it to None. And then `operator.ge` can't handle None

If conversion fails, it's not straightforward to pick a default value - it needs to be the same type as `convert_dtype` and it should also (presumably) fail to pass the filter operator

Proposal:

1. Add a `default_value` field - this is the value that will be used if `convert_dtype()` fails
2. We will define a default `default_value` for each supported `convert_dtype` - `float` and `int` would be zero, and `bool` would be `False

e.g. 
```
block_type: FilterByValueBlock
   block_config:
     block_name: filter_questions
     filter_column: score
     filter_value: 1.0
     operation: eq
     convert_dtype: float
     default_value: 0
```

We would also be able to resolve #123 with this approach - i.e. re-enable multi-processing for the type conversion

markmc:
Fixes #133

We have been using None as a default value when dtype conversion fails. This works fine with the eq operator, but not others e.g. ge

In many cases a default like zero or False works fine, but there are cases where the pipeline author will need to specify a different default, for example when using something like `le(1)` you would want a default greater than 1.

With this change, we're acknowledging that this is default value handling is very normal, and we can drop from error logging to debug logging.

### DIFF
--- a/src/instructlab/sdg/filterblock.py
+++ b/src/instructlab/sdg/filterblock.py
@@ -18,7 +18,7 @@ class FilterByValueBlockError(Exception):
 
 def _get_operator_func(op):
     if not op in dir(operator):
-        raise FilterByValueBlockError("Unknown FilterByValueBlock operation '{op}'")
+        raise FilterByValueBlockError(f"Unknown FilterByValueBlock operation '{op}'")
     return getattr(operator, op)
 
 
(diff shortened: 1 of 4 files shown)


markmc:
Also fixes #123

PR #144, opened 2024-07-15
Set gen_kwargs['n'] dynamically in the simple pipelines

### LINKED ISSUE #130: Fix applying num_instructions_to_generate to simple pipelines
In #86, I knowingly broke the simple pipelines in favor of restoring desired `full` pipeline behavior.

https://github.com/instructlab/sdg/pull/86/commits/8179df58921de285cb4d427967c31d216edc7733

That commit message explains what is needed to fix the simple ones to use the `n` parameter as desired.  We need a way to specify in config that we want `n` set to this runtime parameter, since we do not want that for all cases `n` is not explicitly specified in config.

russellb:
We need a way to allow `--num-instructions` to influence how many
instructions we generate using the simple pipelines. The way to do
this seems to be to set `n` to this value. Since this is a runtime
parameter, and we only want to set it for `n` in certain cases, add a
new value for gen_kwargs['n'] called `dynamic` which is a hint to use
the runtime parameter here.

Closes #130

### DIFF
--- a/src/instructlab/sdg/llmblock.py
+++ b/src/instructlab/sdg/llmblock.py
@@ -133,7 +133,14 @@ class LLMBlock(Block):
             gen_kwargs["max_tokens"] = int(gen_kwargs["max_tokens"])
         if "temperature" in gen_kwargs:
             gen_kwargs["temperature"] = float(gen_kwargs["temperature"])
+        gen_kwargs["n"] = self._get_n(gen_kwargs)
         return gen_kwargs
+
+    def _get_n(self, gen_kwargs):
+        n = gen_kwargs.get("n", 1)
+        if isinstance(n, str) and n == "scaled":
+            n = self.ctx.num_instructions_to_generate
+        return n
 
     def _generate(self, samples, **gen_kwargs) -> list:
         prompts = [
@@ -148,10 +155,9 @@ class LLMBlock(Block):
             )
             return [choice.text.strip() for choice in response.choices]
 
-        n = gen_kwargs.get("n", 1)
         results = []
         for prompt in prompts:
-            for _ in range(n):
+            for _ in range(generate_args["n"]):
                 response = self.ctx.client.completions.create(
                     prompt=prompt, **generate_args
                 )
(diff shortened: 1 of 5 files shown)


russellb:
in draft form so I can let CI run.

I'm also not in love with this implementation, but I don't have any better ideas ...

markmc:
I think we're saying "the pipeline author needs to tell us they want --sdg-scale-factor" to be applied to this block (as per instructlab/instructlab#1570)

Would `n: scaled` be more intuitive?

russellb:
> I think we're saying "the pipeline author needs to tell us they want --sdg-scale-factor to be applied to this block" (as per [instructlab/instructlab#1570](https://github.com/instructlab/instructlab/pull/1570))
> 
> Would `n: scaled` be more intuitive?

Thanks, I like that name better.

russellb:
> > I think we're saying "the pipeline author needs to tell us they want --sdg-scale-factor to be applied to this block" (as per [instructlab/instructlab#1570](https://github.com/instructlab/instructlab/pull/1570))
> > Would `n: scaled` be more intuitive?
> 
> Thanks, I like that name better.

I pushed an update to rename the value to `scaled` as suggested

src/instructlab/sdg/llmblock.py
```diff
@@ -148,10 +155,9 @@ def _generate(self, samples, **gen_kwargs) -> list:
             )
             return [choice.text.strip() for choice in response.choices]
 
-        n = gen_kwargs.get("n", 1)
         results = []
         for prompt in prompts:
-            for _ in range(n):
+            for _ in range(generate_args["n"]):
```
markmc, line 160: Why is this not `get_n()` ?

russellb: That would work and make the code more clear, but in this particular code path, the value was already converted up several lines when it did

```python
    generate_args = self._gen_kwargs(**gen_kwargs)
```

russellb: pro: always use `self._get_n()` so nobody is trying to think of when to use it or not

con: technically not required here 

shrug

markmc (approved):

russellb:
I checked the e2e CI output and it looks correct

PR #150, opened 2024-07-16
Indent simple pipeline "principle" content

derekhiggins:
The first line was being parsed as a new entry in the yaml

Fix #149

PR #151, opened 2024-07-16
ci: run e2e on pipeline config related changes

russellb:
I noticed that the e2e workflow did not run on #150. That PR only
changed some yaml files. When yaml files under `src/` change, we want
to run e2e, just like we do with python source changes.


derekhiggins:
lgtm

markmc (approved):

Issue #128, opened 2024-07-12
Error handling - identify which block in which pipeline triggers an exception


markmc:
See #118 for the background

In general, I think when there is an error with a block, we don't make it easy to know which block in which pipeline caused it

A `config_path` not found case looks like

```
  File "/home/markmc/sdg/src/instructlab/sdg/sdg.py", line 19, in generate
    dataset = pipeline.generate(dataset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/pipeline.py", line 57, in generate
    block = block_type(self.ctx, **block_config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/llmblock.py", line 66, in __init__
    self.block_config = self._load_config(config_path)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/block.py", line 54, in _load_config
    with open(config_path, "r", encoding="utf-8") as config_file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/markmc/sdg/src/instructlab/sdg/configs/skills/freeform_questions.yaml.foobar'
```

Which isn't so bad, because you just have to find the block with that path

But in other cases it might not be so easy to identify which block is causing an error - if we catch exceptions and add the pipeline+block name, that would be a big improvement

gabe-l-hart:
Per a pointer from @russellb, I'm going to take a look at this one today and will report my progress here. I don't have write access to assign to myself though.

markmc:
In #155, @russellb asked for before/after examples for the fix. What's currently proposed is to catch exceptions from `block.generate()`, so here's an example of the current behavior:

```
INFO 2024-07-17 07:08:14,120 pipeline.py:77: generate Running block: gen_questions
INFO 2024-07-17 07:08:14,120 pipeline.py:78: generate Dataset({
    features: ['task_description', 'seed_question', 'seed_response'],
    num_rows: 5
})
...
  File "/home/markmc/sdg/src/instructlab/sdg/sdg.py", line 20, in generate
    dataset = pipeline.generate(dataset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #155, opened 2024-07-16
Block Name In Errors

gabe-l-hart:
## Description

This PR adds `try/except` handling around the core loop in `Pipeline`. In the handler clause, it creates a wrapper exception, attempting to match the type of the internal error, with the `block_name` and `block_type` included.

Closes: https://github.com/instructlab/sdg/issues/128

### DIFF
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -66,45 +66,47 @@ class Pipeline:
         Generate the dataset by running the pipeline steps.
         dataset: the input dataset
         """
-        try:
-            for block_prop in self.chained_blocks:
-                block_name = block_prop["name"]
-                block_type = _lookup_block_type(block_prop["type"])
-                block_config = block_prop["config"]
-                drop_columns = block_prop.get("drop_columns", [])
-                drop_duplicates_cols = block_prop.get("drop_duplicates", False)
-                block = block_type(self.ctx, self, block_name, **block_config)
+        for block_prop in self.chained_blocks:
+            # Parse and instantiate the block
+            block_name = block_prop["name"]
+            block_type = _lookup_block_type(block_prop["type"])
+            block_config = block_prop["config"]
+            drop_columns = block_prop.get("drop_columns", [])
+            drop_duplicates_cols = block_prop.get("drop_duplicates", False)
+            block = block_type(self.ctx, self, block_name, **block_config)
+            logger.info("Running block: %s", block_name)
+            logger.info(dataset)
 
-                logger.info("Running block: %s", block_name)
-                logger.info(dataset)
-
+            # Execute the block and wrap errors with the block name/type
+            try:
                 dataset = block.generate(dataset)
 
-                # If at any point we end up with an empty data set, the pipeline has failed
-                if len(dataset) == 0:
-                    raise EmptyDatasetError(
-                        f"Pipeline stopped: Empty dataset after running block: {block_name}"
-                    )
+            except Exception as err:
+                block_exc_err = (
+                    f"BLOCK ERROR [{block_type.__name__}/{block_name}]: {err}"
+                )
 
-                drop_columns_in_ds = [
-                    e for e in drop_columns if e in dataset.column_names
-                ]
-                if drop_columns:
-                    dataset = dataset.remove_columns(drop_columns_in_ds)
+                # Try to raise the same exception type. This can fail if the
+                # exception is a non-standard type that has a different init
+                # signature, so fall back to raising a RuntimeError in that case.
+                try:
+                    wrapper_err = type(err)(block_exc_err)
+                except TypeError:
+                    wrapper_err = RuntimeError(block_exc_err)
+                raise wrapper_err from err
 
-                if drop_duplicates_cols:
-                    dataset = self._drop_duplicates(dataset, cols=drop_duplicates_cols)
-        except Exception as err:
-            block_exc_err = f"BLOCK ERROR [{block_type.__name__}/{block_name}]: {err}"
+            # If at any point we end up with an empty data set, the pipeline has failed
+            if len(dataset) == 0:
+                raise EmptyDatasetError(
+                    f"Pipeline stopped: Empty dataset after running block: {block_name}"
+                )
 
-            # Try to raise the same exception type. This can fail if the
-            # exception is a non-standard type that has a different init
-            # signature, so fall back to raising a RuntimeError in that case.
-            try:
-                wrapper_err = type(err)(block_exc_err)
-            except TypeError:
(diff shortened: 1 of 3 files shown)


src/instructlab/sdg/pipeline.py
```diff
@@ -65,31 +65,45 @@ def generate(self, dataset) -> Dataset:
...
+                block_type = _lookup_block_type(block_prop["type"])
+                block_config = block_prop["config"]
+                drop_columns = block_prop.get("drop_columns", [])
+                drop_duplicates_cols = block_prop.get("drop_duplicates", False)
+                block = block_type(self.ctx, self, block_name, **block_config)
+
+                logger.info("Running block: %s", block_name)
+                logger.info(dataset)
+
+                dataset = block.generate(dataset)
```
russellb, line 80: It looks like these lines capture the block-specific code, so you could localize your addition to this part, I think.

russellb (commented):
one suggestion to localize the try/except handling to only the parts calling block-specific code.

Also, do you have an example of what an error looked like before and after the change?

src/instructlab/sdg/pipeline.py
gabe-l-hart: Good point! Do you think it's worth containing the config parsing as well? It seems like the use of direct indexing (vs `.get`) for `"name"`, `"type"`, and `"config"` implies that these are all required fields, but I don't see anywhere that they're validated (though that might happen elsewhere that I haven't found yet?). If it's not done elsewhere, it may be worth doing that validation in the `__init__` to avoid getting part-way through the expensive pipeline operation before discovering configuration errors.

russellb: we now have some config validation, at least for the configs we have in the tree.  See `make validate-pipelines`. It uses a jsonschema definition

gabe-l-hart: Ok, cool, makes sense!

gabe-l-hart:
Thanks for the suggestion! I've localized the error handling to just the `generate` call.

>Also, do you have an example of what an error looked like before and after the change?

I don't have a concrete example of before/after, but this uses [raise ... from ...](https://docs.python.org/3/library/exceptions.html#BaseException.__cause__) syntax. The parent exception will attempt to match the exception type of the one raised in the block with a message `"BLOCK ERROR [{block_type class name}/{block_name}]: {child exc message}"`. The original exception can be accessed via the `__cause__` attribute of the wrapper exception.

markmc:
> Also, do you have an example of what an error looked like before and after the change?

Here's what an exception from `LLMBlock.generate()` looks like:

```
INFO 2024-07-17 07:08:14,120 pipeline.py:77: generate Running block: gen_questions
INFO 2024-07-17 07:08:14,120 pipeline.py:78: generate Dataset({
    features: ['task_description', 'seed_question', 'seed_response'],
    num_rows: 5
})
Traceback (most recent call last):
...
  File "/home/markmc/sdg/src/instructlab/sdg/sdg.py", line 20, in generate
    dataset = pipeline.generate(dataset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/pipeline.py", line 80, in generate
    dataset = block.generate(dataset)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markmc/sdg/src/instructlab/sdg/llmblock.py", line 181, in generate
    raise Exception("TEST")
Exception: TEST
```

This isn't actually so problematic since we log which block is being executed, so the user has some context to use in order to debug the error

The example  I gave in #128 is failing to load a config file when instantiating a block - an error with the pipeline definition, but also one that can't be detected by the schema

src/instructlab/sdg/pipeline.py
```diff
@@ -67,17 +67,33 @@ def generate(self, dataset) -> Dataset:
...
             logger.info("Running block: %s", block_name)
             logger.info(dataset)
 
-            dataset = block.generate(dataset)
+            # Execute the block and wrap errors with the block name/type
+            try:
+                dataset = block.generate(dataset)
+
+            except Exception as err:
+                block_exc_err = (
```
markmc, line 85: What would be so wrong with simply doing

```
raise BlockGenerateError(f"Error occurred in block.generate() '{block.name}': {str(err)}") from err
```

gabe-l-hart: Good question. I was trying to be minimally invasive in case there was type-specific error handling somewhere above here in the CLI, but I think that's probably just being overly cautious. I'll switch to a well-defined error like this.

src/instructlab/sdg/pipeline.py
```diff
@@ -32,6 +33,32 @@ def __init__(
...
 
 
+# This is part of the public API.
+class BlockGenerationError(Exception):
+    """A BlockGenerationError occurs when a block generates an exception during
+    generation. It contains information about which block failed and why.
+    """
+
+    def __init__(self, block: Block, exception: Exception):
+        self.block = block
```
markmc, line 43: nit: underscore prefix this if you want just `.block_name` and `.block_type` to be used?

src/instructlab/sdg/pipeline.py
```diff
@@ -67,17 +94,21 @@ def generate(self, dataset) -> Dataset:
         dataset: the input dataset
         """
         for block_prop in self.chained_blocks:
+            # Parse and instantiate the block
             block_name = block_prop["name"]
             block_type = _lookup_block_type(block_prop["type"])
             block_config = block_prop["config"]
             drop_columns = block_prop.get("drop_columns", [])
             drop_duplicates_cols = block_prop.get("drop_duplicates", False)
             block = block_type(self.ctx, self, block_name, **block_config)
```
markmc, line 103: I know this contradicts what @russellb said, but I think it would be good to wrap the instantiation too - e.g. the config_path not found error I showed in the issue

(I agree with not wrapping the whole contents of this for loop though)

src/instructlab/sdg/pipeline.py
```diff
@@ -32,6 +33,32 @@ def __init__(
         self.num_procs = 8
 
 
+# This is part of the public API.
+class BlockGenerationError(Exception):
```
markmc, line 37: I suggested this, so sorry I'm changing - `PipelineBlockError` if it wraps the block instantiation too?

Also, add it to `__all__` in `__init__.py`

src/instructlab/sdg/pipeline.py
gabe-l-hart: I left that one "public" thinking that some might like to be able to actually muck with the block itself when handling the exception.

markmc (approved):
Thanks, @gabe-l-hart !

hickeyma (approved):

Issue #162, opened 2024-07-18
[Epic] Support for mixing generated datasets before training


bbrowning:
## Overview

The research team that developed InstructLab's processes has determined that we need a way to mix generated datasets before training. This is necessary to get the best results we can when adding knowledge to a model.

This issue tracks the work across the SDG and other repos required to implement this change.

## **instructlab/sdg repository**
`0.2.0` milestone: https://github.com/instructlab/sdg/milestone/4

In-progress PR at #163

* [x]  cherry-pick only the data-mixing commits from https://github.com/aakankshaduggal/sdg/pull/4/ on top of instructlab/sdg main (some other changes related to knowledge schema and other bits snuck into there)
* [x]  determine if [batching and parallel generation](https://github.com/aakankshaduggal/sdg/pull/6) is a hard prerequisite for data mixing or can be done separately - if separate, create an issue to track merging batching and parallel generation
  * This can be done separately, and #135 is already tracking that work.
* [x]  determine if [caching](https://github.com/aakankshaduggal/sdg/pull/9) is a hard prerequisite for data mixing or can be done separately - if separate, create an issue to track merging caching
  * This can be done separately, and #168 created to track merging that in.
* [x] fix unresolved placeholder value in `src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml` causing `<path_to_dataset>` to actually get used as a path when attempting skill data generation
  * This depends on a precomputed dataset being released somewhere (such as HuggingFace), and code modified to download and reference that dataset from its coordinates listed in default_recipe.yaml. For now I just removed the placeholder (so skills are running with no precomputed data added in), and opened #171 to track figuring this out.
* [x] Update data mixing code to work with "simple" pipeline or, if not possible, discuss fallout if it only works in the "full" pipeline
  * I was able to get this working with the simple pipeline by utilizing the _get_question and _get_response helpers.
* [x]  Fix `build_raft_dataset` in `parse_and_convert.py` to not infinitely loop when working with a small dataset, such as 1-2 pieces of generated data that we'll encounter in the "simple" CI pipeline or with users testing locally with very small numbers of instructions.
* [x] Fix bug in `generate_data.py` where if a knowledge taxonomy leaf gets generated first, it treats all subsequent taxonomy leaves as knowledge even though they may be skills, which blows up
* [x] Ensure legacy train continues to work by continuing to produce `train_*.jsonl` and `test_*.jsonl` files.
[...]

PR #163, opened 2024-07-18
Introduce a way to mix generated datasets before sending to training

bbrowning:
See #162 

Rebased from aakankshaduggal#4,  aakankshaduggal#13, and aakankshaduggal#14 reducing scope and splitting what can out into subsequent follow-up PRs for the overall data mixing functionality instead of landing everything as one big PR.

Refactored and changed by me to be as least disruptive as possible in merging in, which means leaving the legacy train and messages outputs in place in generated_data.py.

This introduces a new Recipe class that references multiple generated datasets and has logic to combine those into a single mixed dataset based on a configurable number / ratio of samples to take from each dataset.

This adds new output artifacts for each data generation run:

* a `node_datasets_*` subfolder that contains the raw generated samples   from each taxonomy leaf node as a set of jsonl files - within here,  each skill taxonomy node gets a single jsonl and knowledge nodes get   two jsonl files used in separate downstream steps of training
* a `skills_train_msgs_*.jsonl` and a `knowledge_train_msgs_*.jsonl`  file that contains mixtures of the above raw generated samples based  on proportions specified during the mixing process

```
Co-authored-by: shivchander <shivchander.s30@gmail.com>
Co-authored-by: Khaled Sulayman <khaled@thesulaymans.com>
Co-authored-by: abhi1092 <abhi1092@gmail.com>
Co-authored-by: Aakanksha Duggal <aduggal@redhat.com>
```

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -17,9 +17,9 @@ import openai
 import platformdirs
 
 # First Party
-from instructlab.sdg.eval_data import generate_eval_task_data, mmlubench_pipe_init
-
 # pylint: disable=ungrouped-imports
+from instructlab.sdg.datamixing import DataMixer, _get_question_hack, _get_response_hack
+from instructlab.sdg.eval_data import generate_eval_task_data, mmlubench_pipe_init
 from instructlab.sdg.llmblock import MODEL_FAMILY_MERLINITE, MODEL_FAMILY_MIXTRAL
 from instructlab.sdg.pipeline import (
     FULL_PIPELINES_PACKAGE,
@@ -34,52 +34,20 @@ from instructlab.sdg.utils.taxonomy import (
     read_taxonomy_leaf_nodes,
 )
 
-_SYS_PROMPT = "You are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior."
+_SYS_PROMPT = "I am, Red Hat® Instruct Model based on Granite 7B, an AI language model developed by Red Hat and IBM Research, based on the Granite-7b-base language model. My primary function is to be a chat assistant."
 
 
 def _unescape(s):
-    return bytes(s, "utf-8").decode("utf-8")
-
-
-# This is a hack because the simple workflow returns a q/a pair as a single output.
-# We could possibly try to ask for them separately, but it would cost twice the inference
-# API calls. All of this is because the smallest models we use on small environments
-# for testing and demos weren't good enough to follow the strict formatting instructions used
-# in the full pipeline.
-def _get_question(logger, synth_example):
-    if "question" in synth_example:
-        return synth_example["question"]
-
-    if not synth_example.get("output"):
-        raise GenerateException(
-            f"Error: output not found in synth_example: {synth_example}"
-        )
-
-    parts = synth_example["output"].split("?", 1)
-    if len(parts) != 2:
-        logger.warning(f"Failed to split generated q&a: {synth_example['output']}")
-    return parts[0].strip() + "?" if len(parts) == 2 else ""
-
-
-# This is also a hack. See the comment above _get_question.
-def _get_response(logger, synth_example):
-    if "response" in synth_example:
-        return synth_example["response"]
-
-    if "output" not in synth_example:
-        raise GenerateException(
-            f"Error: output not found in synth_example: {synth_example}"
-        )
-
-    parts = synth_example["output"].split("?", 1)
-    if len(parts) != 2:
-        logger.warning(f"Failed to split generated q&a: {synth_example['output']}")
-    return parts[1].strip() if len(parts) == 2 else parts[0].strip()
+    return bytes(s, "utf-8").decode("utf-8").strip()
 
 
 def _convert_to_messages(sample):
     """
     Convert a sample dictionary to contain 'messages' and 'metadata' columns required for training.
+
+    Note that this is for the legacy messages format, used before data
+    mixing was introduced. Once we can drop the older `messages_*.jsonl`
+    output files, this can go away.
     """
     # Create user query message
     user_query = sample["inputs"]
@@ -107,16 +75,24 @@ def _convert_to_messages(sample):
 def _gen_train_data(
     logger, machine_instruction_data, output_file_train, output_file_messages
 ):
+    """
+    Generate training data in the legacy system/user/assistant format
+    used in train_*.jsonl as well as the legacy messages format used
+    in messages_*.jsonl files.
+
+    This can be dropped once we no longer need those formats and are fully
+    using the new data mixing messages format.
+    """
     train_data = []
     messages_data = []
 
     for output_dataset in machine_instruction_data:
         for synth_example in output_dataset:
             logger.debug(synth_example)
-            user = _get_question(logger, synth_example)
+            user = _get_question_hack(synth_example)
             if len(synth_example.get("context", "")) > 0:
                 user += "\n" + synth_example["context"]
-            assistant = _unescape(_get_response(logger, synth_example))
+            assistant = _unescape(_get_response_hack(synth_example))
             train_entry = {
                 "system": _SYS_PROMPT,
                 "user": _unescape(user),
@@ -159,6 +135,10 @@ def _gen_test_data(
     leaf_nodes,
     output_file_test,
 ):
+    """
+    Generate test data in the format needed by the legacy Linux training
+    in instructlab/instructlab.
+    """
     test_data = []
     for _, leaf_node in leaf_nodes.items():
         for seed_example in leaf_node:
@@ -324,7 +304,6 @@ def generate_data(
     date_suffix = datetime.now().replace(microsecond=0).isoformat().replace(":", "_")
     output_file_messages = f"messages_{name}_{date_suffix}.jsonl"
     output_file_test = f"test_{name}_{date_suffix}.jsonl"
-    # train data in messages format that will be mixed and split up into train test eventually
     output_file_train = f"train_{name}_{date_suffix}.jsonl"
 
     _gen_test_data(
@@ -362,6 +341,8 @@ def generate_data(
 
     mmlu_bench_pipe = mmlubench_pipe_init(ctx)
 
+    mixer = DataMixer(output_dir, date_suffix, _SYS_PROMPT, ctx.dataset_num_procs)
+
     if console_output:
         logger.info(
             "Synthesizing new instructions. If you aren't satisfied with the generated instructions, interrupt training (Ctrl-C) and try adjusting your YAML files. Adding more examples may help."
@@ -369,6 +350,8 @@ def generate_data(
 
     generated_data = None
     for leaf_node in leaf_nodes.values():
+        is_knowledge = False
+        leaf_node_path = leaf_node[0]["taxonomy_path"].replace("->", "_")
         samples = leaf_node_to_samples(leaf_node, server_ctx_size, chunk_word_count)
 
         if not samples:
@@ -376,8 +359,11 @@ def generate_data(
 
         if samples[0].get("document"):
             sdg = sdg_knowledge
+            is_knowledge = True
+
         elif samples[0].get("seed_context"):
             sdg = sdg_grounded_skill
+
         else:
             sdg = sdg_freeform_skill
 
@@ -393,9 +379,8 @@ def generate_data(
         logger.info("Generated %d samples" % len(generated_data))
         logger.debug("Generated data: %s" % generated_data)
 
-        if samples[0].get("document"):
+        if is_knowledge:
             # generate mmlubench data for the current leaf node
-            leaf_node_path = leaf_node[0]["taxonomy_path"].replace("->", "_")
             generate_eval_task_data(
                 mmlu_bench_pipe,
                 leaf_node_path,
@@ -403,6 +388,8 @@ def generate_data(
                 output_dir,
                 date_suffix,
             )
+
+        mixer.collect(leaf_node_path, new_generated_data, is_knowledge)
 
     if generated_data is None:
         generated_data = []
@@ -414,11 +401,7 @@ def generate_data(
         os.path.join(output_dir, output_file_messages),
     )
 
-    # TODO
-    # This is for backwards compatibility. The file existing previously, so we'll keep it for now.
-    # I believe the github bot assumes it is present for presenting generated data to a taxonomy
-    # reviewer or contributor. Otherwise, I don't see a consumer of it in this repo or the
-    # `ilab` CLI.
+    mixer.generate()
 
     generate_duration = time.time() - generate_start
     logger.info(f"Generation took {generate_duration:.2f}s")
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -1,0 +1,471 @@
+# Standard
+from typing import Optional
+import json
+import os.path
+import random
+import uuid
+
+# Third Party
+from datasets import Dataset, concatenate_datasets, load_dataset
+
+# First Party
+from instructlab.sdg.logger_config import setup_logger
+from instructlab.sdg.utils import GenerateException
+
+ALLOWED_COLS = ["id", "messages", "metadata"]
+logger = setup_logger(__name__)
+
+
+def _adjust_train_sample_size(ds: Dataset, num_samples: int):
+    """
+    Return a dataset with num_samples random samples selected from the
+    original dataset.
+    """
+    logger.info(f"Rebalancing dataset to have {num_samples} samples ...")
+    df = ds.to_pandas()
+    df = df.sample(n=num_samples, random_state=42, replace=True).reset_index(drop=True)
+    return Dataset.from_pandas(df)
+
+
+def _load_ds(path, sampling_size, num_proc):
+    """
+    Load a dataset from the given file path and select sampling_size
+    number/ratio of samples from it, ensuring the loaded dataset has only
+    ALLOWED_COLS columns in it with any additional columns moved to the
+    metadata section.
+    """
+    logger.info(f"Loading dataset from {path} ...")
+    dataset = load_dataset("json", data_files=path, split="train")
+    logger.info(f"Dataset columns: {dataset.column_names}")
+    logger.info(f"Dataset loaded with {len(dataset)} samples")
+
+    if sampling_size != 1.0:
+        if isinstance(sampling_size, int):
+            num_samples = sampling_size
+        else:
+            num_samples = int(len(dataset) * sampling_size)
+        dataset = _adjust_train_sample_size(dataset, num_samples)
+
+    # move any column that is not in ALLOWED_COLS to metadata
+    def _move_unallowed_cols_to_metadata(example):
+        metadata = example.get("metadata", {})
+        if isinstance(metadata, str):
+            metadata = json.loads(metadata)
+        for col in dataset.column_names:
+            if col not in ALLOWED_COLS:
+                metadata[col] = example[col]
+                example.pop(col)
+        example["metadata"] = json.dumps(metadata)
+        return example
+
+    dataset = dataset.map(_move_unallowed_cols_to_metadata, num_proc=num_proc)
+
+    # check if metadata column is string if not convert it using json.dumps
+    if not isinstance(dataset["metadata"][0], str):
+        dataset = dataset.map(
+            lambda x: {"metadata": json.dumps(x["metadata"])}, num_proc=num_proc
+        )
+
+    return dataset
+
+
+def _add_system_message(sample: dict, sys_prompt: str) -> dict:
+    """
+    Ensure every sample has a system message with the correct system prompt
+    """
+    # check if the messages have role system
+    has_system = False
+    for msg in sample["messages"]:
+        if msg["role"] == "system":
+            has_system = True
+            msg["content"] = sys_prompt
+
+    if not has_system:
+        sample["messages"].insert(0, {"role": "system", "content": sys_prompt})
+
+    return sample
+
+
+class Recipe:
+    """
+    A Recipe describes how datasets were mixed, including the path and
+    sampling size used for each included dataset as well as the system
+    prompt used when generating the data in those datasets.
+    """
+
+    def __init__(
+        self, initial_datasets: Optional[list] = None, sys_prompt: Optional[str] = ""
+    ):
+        self.recipe = {
+            "datasets": initial_datasets or [],
+            "sys_prompt": sys_prompt,
+        }
+        self.sys_prompt = self.recipe.get("sys_prompt", "")
+        self.dataset_added = False
+
+    def _create_mixed_dataset(self, num_proc):
+        """
+        Create the mixed dataset from its list of included datasets, taking
+        into account the desired sampling size from each individual dataset
+        to control the overall mix of samples in the final dataset.
+        """
+        if not self.dataset_added:
+            logger.error("No dataset added to the recipe")
+
+        mixed_ds = [
+            _load_ds(dataset["path"], dataset["sampling_size"], num_proc)
+            for dataset in self.recipe["datasets"]
+        ]
+
+        mixed_ds = concatenate_datasets(mixed_ds)
+        mixed_ds = mixed_ds.map(
+            _add_system_message,
+            fn_kwargs={"sys_prompt": self.sys_prompt},
+            num_proc=num_proc,
+        )
+
+        # assert that the dataset only has the allowed columns
+        assert set(mixed_ds.column_names) == set(
+            ALLOWED_COLS
+        ), "Dataset has invalid columns"
+        return mixed_ds
+
+    def add_dataset(self, path, sampling_size):
+        """
+        Add a dataset to this recipe.
+
+        Args:
+            path: The file path to the dataset's samples, as jsonl
+            sampling_size: An int or float that specifices the number of
+                           samples (if int) or the ratio of samples (if
+                           float) to include in the mixed dataset. A value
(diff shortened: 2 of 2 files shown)


bbrowning:
The e2e test finishes synthetic data generation now, although fails during training because of `/tmp/tmp.JJXU361JVM/.local/share/instructlab/datasets does not contain training or test files, did you run `ilab data generate`?` This is because the mixed files do not match the `train_*` or `test_*` prefixes that it expects those files to have, and instead have names like `skills_train_msgs_2024-07-18T16_10_49.jsonl` or `knowledge_train_msgs_2024-07-18T16_10_49.jsonl`.

src/instructlab/sdg/generate_data.py
```diff
@@ -318,47 +265,78 @@ def generate_data(
...
+                output_dir, f"node_datasets_{date_suffix}/node_{i}_p07.jsonl"
+            )
+            skills_fpath = os.path.join(
+                output_dir, f"node_datasets_{date_suffix}/node_{i}_p10.jsonl"
+            )
+            knowledge_phase_data.to_json(knowledge_fpath, orient="records", lines=True)
+            skills_phase_data.to_json(skills_fpath, orient="records", lines=True)
+
+            knowledge_recipe.add_dataset(knowledge_fpath)
+            skills_recipe.add_dataset(skills_fpath)
```
markmc, line 313: Is the roundtrip to disk necessary? i.e. we're writing to the dataset to a file just for it to be loaded again shortly after?

The benefit is that we have a record of this intermediate stage? Or reduced memory usage?

src/instructlab/sdg/utils/datamixing.py
```diff
@@ -0,0 +1,122 @@
...
+        if isinstance(metadata, str):
+            metadata = json.loads(metadata)
+        for col in dataset.column_names:
+            if col not in ALLOWED_COLS:
+                metadata[col] = example[col]
+                example.pop(col)
+        example["metadata"] = json.dumps(metadata)
+        return example
+
+    dataset = dataset.map(move_unallowed_cols_to_metadata, num_proc=8)
```
markmc, line 48: Another place where we should be using `PipelineContext.num_proc`

src/instructlab/sdg/generate_data.py
```diff
@@ -264,29 +192,48 @@ def generate_data(
     """
     generate_start = time.time()
 
+    knowledge_recipe = Recipe(
+        "src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml"
+    )
+    skills_recipe = Recipe(
+        "src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml"
```
markmc, line 199: I feel like the main purpose of this file is to specify some initial data sets that should be mixed with the generated datasets?

Couldn't that be specified in the pipeline config:

```
version: "1.0"
initial_datasets:
  - path: ../../datasets/mydataset.jsonl
    sampling_size: 30
blocks:
  - name: gen_questions
    type: LLMBlock
```

?

It also seems to serve a secondary purpose of writing a record of the datasets that were mixed to produce the final output?

src/instructlab/sdg/utils/parse_and_convert.py
```diff
@@ -0,0 +1,313 @@
...
+        {"content": response, "role": "assistant"},
+    ]
+
+    return sample
+
+
+def create_auxiliary_dataset(generated_dataset: Dataset):
+    if "dataset_type" not in generated_dataset.column_names:
+        return None
+    auxilary_path = "src/instructlab/sdg/config/kowledge/auxilary_instructions.yaml"
```
markmc, line 129: "kowlege" typo ... and the file doesn't exist

src/instructlab/sdg/utils/parse_and_convert.py
```diff
@@ -0,0 +1,313 @@
...
+            ]
+        ]
+    )
+    print(unique_document_auxiliary)
+    unique_document_auxiliary = unique_document_auxiliary.rename_columns(
+        {"raw_document": "context", "document": "response"}
+    )
+
+    def __create_auxiliary_ds(rec):
+        instruction = random.choice(auxiliary_inst[rec["dataset_type"]])
```
markmc, line 163: Unclear to me where `dataset_type` is coming from and what it's valid values are?

markmc (commented):
A bunch of comments all over the place, not terribly useful I'm afraid. Will try to follow up with some higher-level comments

src/instructlab/sdg/generate_data.py
```diff
@@ -241,31 +151,40 @@ def generate_data(
...
     """
+    logger = setup_logger(__name__)
     generate_start = time.time()
 
+    knowledge_recipe = Recipe("src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml")
+    skills_recipe = Recipe("src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml")
+
+    sys_prompt = knowledge_recipe.sys_prompt    
+    logger.info(f"System prompt: {sys_prompt}")
+    assert sys_prompt == skills_recipe.sys_prompt, "System prompts must be the same for both knowledge and skills"
```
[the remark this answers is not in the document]
bbrowning: @aakankshaduggal Can you answer Mark's question about default recipes? They only specify a system prompt, and the skill and knowledge recipes (2 different files) are required to have the identical system prompt. What's the overall purpose these default recipes are serving?

src/instructlab/sdg/utils/parse_and_convert.py
```diff
@@ -0,0 +1,313 @@
...
+        {
+            "role": "pretraining",
+            "content": f"<|user|>\n{rec['messages'][0]['content']}\n<|assistant|>\n{rec['messages'][1]['content']}",
+        }
+    ]
+    return rec
+
+
+def create_phase10_ds(generated_dataset: Dataset):
+    # Phase 1.0
```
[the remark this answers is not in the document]
bbrowning: @aakankshaduggal @shivchander Do either of you have a document or description of what exactly Phase 0.7 and Phase 1.0 mean? This sounds both like something we should write into a dev-doc (see Mark's request for that at https://github.com/instructlab/sdg/pull/163#issuecomment-2239071197) as well as something that either needs to make its way into code comments or I can swap those terms out for something more human-friendly in the method names and references here, if there's some common name given to these training phases outside of phase 0.7 and phase 1.0.

src/instructlab/sdg/generate_data.py
```diff
@@ -27,120 +27,25 @@
...
-        for entry in train_data:
-            json.dump(entry, outfile, ensure_ascii=False)
-            outfile.write("\n")
-
-    with open(output_file_messages, "w", encoding="utf-8") as outfile:
-        for entry in messages_data:
-            json.dump(entry, outfile, ensure_ascii=False)
-            outfile.write("\n")
+# Constants
+NUM_SYNTH_SKILLS = 30
```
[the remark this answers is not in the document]
bbrowning: @aakankshaduggal Is there some background on why 30 was chosen as the sampling size for skills recipes? Why not 5, 100, 500? Is this actually a constant that no one should ever modify, or is this something that would need to be tweaked depending on the dataset in use?

src/instructlab/sdg/utils/parse_and_convert.py
bbrowning: This feels like dead/outdated code to me, and I'm tempted to rip out all the auxiliary dataset bits in this PR to let someone else add them later with a follow-up and explanation of what it is and how it's used. I can't see anything that would cause this to actually run today, and it's code not exercised by any of the manual or automated testing I've performed. Looping in @aakankshaduggal and @shivchander in case they feel otherwise and can explain what this does.

src/instructlab/sdg/generate_data.py
bbrowning: I split the concerns of figuring out how to inject initial data sets into #171, which means for the scope of this PR the default recipes are really just bringing the system prompt. These files don't seem like a great way to manage that, since as you pointed out they have to be equal in both knowledge and skills recipes. I'm giving a chance for some of the others I tagged to speak up here, but if they don't then will proceed with removing the default recipe files entirely from this PR for now.

src/instructlab/sdg/generate_data.py
```diff
@@ -333,23 +497,34 @@ def generate_data(
         batch_size=batch_size,
         batch_num_workers=num_cpus,
     )
+    sdg_knowledge, sdg_freeform_skill, sdg_grounded_skill = sdgs
+    # Get a PipelineContext to use in dataset operations below
+    ctx_list = [pipe.ctx for sdg in sdgs for pipe in sdg.pipelines]
```
derekhiggins, line 502: Decoding this is difficult, maybe something like
```
if not (len(sdgs) and len(sdgs[0].pipelines)):
    assert (.......
```

src/instructlab/sdg/generate_data.py
```diff
@@ -365,6 +540,38 @@ def generate_data(
         logger.info("Generated %d samples" % len(generated_data))
         logger.debug("Generated data: %s" % generated_data)
 
+        if is_knowledge:
+            knowledge_phase_data = _create_phase07_ds(logger, new_generated_data)
+            output_file_leaf_knowledge = (
+                f"node_datasets_{date_suffix}/node_{i}_p07.jsonl"
```
derekhiggins, line 546: I'm missing context here but I'd love to know the signifigance of p07 and p10 below

derekhiggins (commented):
I think I need to see the e2e runthough to  understand whats going on here

src/instructlab/sdg/generate_data.py
```diff
@@ -27,16 +29,20 @@
...
 from instructlab.sdg.sdg import SDG
 from instructlab.sdg.utils import GenerateException, models
+from instructlab.sdg.utils.datamixing import Recipe
 from instructlab.sdg.utils.taxonomy import (
     leaf_node_to_samples,
     read_taxonomy_leaf_nodes,
 )
 
-_SYS_PROMPT = "You are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior."
+_SYS_PROMPT = "I am, Red Hat® Instruct Model based on Granite 7B, an AI language model developed by Red Hat and IBM Research, based on the Granite-7b-base language model. My primary function is to be a chat assistant."
```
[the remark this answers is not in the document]
bbrowning: This is what the prompt was changed to in the research fork, although there they are loading it from default_recipe.yaml files that I've moved out of this PR just for simplicity sake for now. However, it was hardcoded in those files to this value.

src/instructlab/sdg/generate_data.py
bbrowning: That's not a bad idea - I wasn't completely happy with this, as it feels like an awkward way to get a context back from `_sdg_init`. A prior iteration of this just modified `_sdg_init` to also return a PipelineContext, but then that felt odd in a different way because the existing return value is a list of Pipeline objects. It may be cleaner to create a new method that generates a PipelineContext, call that from `generate_data`, and then pass that context into `_sdg_init` instead of having `_sdg_init` create the PipelineContext itself.

src/instructlab/sdg/generate_data.py
bbrowning: I don't have full context on why they're named this, but did push some additional changes that attempts to document what's in these p07 and p10 datasets at least.

src/instructlab/sdg/generate_data.py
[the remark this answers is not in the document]
bbrowning: @aakankshaduggal Do you know if this needs to be conditional and configurable? It was hardcoded into the recipe files before - does that mean it was generally just expected to be hardcoded? Or was there an expectation that an out-of-band process might be replacing those recipes files on disk before running data generation, thus overwriting this system prompt?

src/instructlab/sdg/generate_data.py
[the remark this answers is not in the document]
bbrowning: Ok, I documented this constant to the best of my current understanding.

src/instructlab/sdg/generate_data.py
```diff
@@ -264,29 +192,48 @@ def generate_data(
...
     leaf_nodes = read_taxonomy_leaf_nodes(taxonomy, taxonomy_base, yaml_rules)
+    logger.info(f"Found {len(leaf_nodes)} leaf nodes in the taxonomy.")
+    logger.info(f"Generating data for {list(leaf_nodes.keys())} leaf nodes.")
+
     if not leaf_nodes:
         raise GenerateException("Error: No new leaf nodes found in the taxonomy.")
 
     name = Path(model_name).stem  # Just in case it is a file path
     date_suffix = datetime.now().replace(microsecond=0).isoformat().replace(":", "_")
-    output_file_messages = f"messages_{name}_{date_suffix}.jsonl"
```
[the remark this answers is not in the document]
bbrowning: I've refactored this PR to be as purely additive as possible, including leaving the previous code in place that generates the `train_*.jsonl` files and `messages_*.jsonl` files. So, we're no longer removing that file from our output. This leaves the door open to a later discussion and follow-up about deprecating/removing those legacy file outputs separately from adding in the new mixed datasets.

src/instructlab/sdg/generate_data.py
markmc: See https://github.com/instructlab/sdg/pull/180

```
-     sdg_knowledge, sdg_freeform_skill, sdg_grounded_skill = _sdg_init(
+     ctx, sdg_knowledge, sdg_freeform_skill, sdg_grounded_skill = _sdg_init(
         pipeline,
         client,
         model_family,
```

markmc: but passing a PipelineContext to 'sdg_init()' would be fine too

markmc: Changed #180 to do what you suggested, thanks

https://github.com/instructlab/sdg/commit/57b2a77bdbcdf8b15a6d3453298f3b2854c659df

src/instructlab/sdg/utils/datamixing.py
```diff
@@ -0,0 +1,108 @@
+import json
```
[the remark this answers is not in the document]
bbrowning: Agreed, will bring this up one level and ensure private methods use the leading underscore naming convention to control what what we expose from here.

src/instructlab/sdg/generate_data.py
bbrowning: I split writing recipe files to disk out as #185 to reduce scope of this PR.

src/instructlab/sdg/generate_data.py
bbrowning: My understanding is that it's so we have an artifact trail of the raw generated data as well as the mixed data, and when #185 goes in the recipe files will reference this raw generated data to show what was mixed into the overall mixed dataset. The mixing process can reduce the number of samples from a given taxonomy leaf in the overall mixed dataset, so these intermediate files also allow the entire generated output for each leaf node to be retained as an artifact.

Reduce memory usage is the other side-effect of this, since we're only ever keeping the dataset for a single leaf node in memory until the mixing process, where we sample some number/percent of the overall generated samples for each leaf instead of loading the entirety of all leaf node generated samples at once.

src/instructlab/sdg/generate_data.py
```diff
@@ -318,47 +265,78 @@ def generate_data(
...
 
         if not samples:
             raise GenerateException("Error: No samples found in leaf node.")
 
         if samples[0].get("document"):
             sdg = sdg_knowledge
+            logger.info(f"Generating data for leaf node {i} with knowledge pipeline.")
+            is_knowledge = True
+            # add to 0.7 recipe
+            # add to 1.0 recipe
```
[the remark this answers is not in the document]
bbrowning: They were not TODOs, but were confusing and so are removed with separate docs added to explain phase 0.7 and 1.0 in their relevant _create_phase methods.

src/instructlab/sdg/generate_data.py
```diff
@@ -264,29 +192,48 @@ def generate_data(
...
+    )
+    skills_recipe = Recipe(
+        "src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml"
+    )
+
+    sys_prompt = knowledge_recipe.sys_prompt
+    logger.info(f"System prompt: {sys_prompt}")
+    assert (
+        sys_prompt == skills_recipe.sys_prompt
+    ), "System prompts must be the same for both knowledge and skills"
```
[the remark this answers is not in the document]
bbrowning: Agreed, and I'll let #171 handle dealing with configurable system prompts that match the precomputed datasets as support for precomputed datasets is added. For now this was changed to retain the previous behavior of a hardcoded _SYS_PROMPT until/unless we need configurable system prompt for other reasons.

src/instructlab/sdg/generate_data.py
```diff
@@ -318,47 +265,78 @@ def generate_data(
...
+        is_knowledge = False
         samples = leaf_node_to_samples(leaf_node, server_ctx_size, chunk_word_count)
+        ds = Dataset.from_list(samples)
 
         if not samples:
             raise GenerateException("Error: No samples found in leaf node.")
 
         if samples[0].get("document"):
             sdg = sdg_knowledge
+            logger.info(f"Generating data for leaf node {i} with knowledge pipeline.")
```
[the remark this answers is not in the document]
bbrowning: The logging improvements were removed and can be made in a follow-up PR to reduce scope of this one.

src/instructlab/sdg/utils/parse_and_convert.py
```diff
@@ -0,0 +1,313 @@
...
+    sample["user"] = user_query
+    sample["assistant"] = response
+
+    return sample
+
+
+def _convert_to_messages(sample: dict, sys_prompt: str):
+    """
+    Convert a sample dictionary to contain 'messages'
+    and 'metadata' columns required for training.
```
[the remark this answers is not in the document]
bbrowning: Outdated now, as the previous _convert_to_messages is now retained which does have a `metadata` column.

bbrowning:
I think I've addressed most of the initial feedback, although still waiting on a couple of answers from others to ensure I've documented things properly and didn't prune too much out in my attempts to streamline this.  I've also parked a copy of my fork's branch at https://github.com/bbrowning/instructlab-sdg/commits/data-mixing-full-history/ up to this point for reference to myself or anyone else that works on follow-up PRs to remove things I'm pruning out of scope for this one.

Now, on to the squash/rebase phase to get the list of commits down to a reasonable number and ensure authorship is correctly attributed. Once that's done, I think this will be in decent shape for additional review and merging consideration, with the understanding that it's not perfect but has to get in under specific time constraints and is a prerequisite for some smaller follow-up PRs around recipe files, system prompts, and removal of legacy train/messages formats as well as overall integration testing end-to-end.

src/instructlab/sdg/generate_data.py
```diff
@@ -241,31 +151,40 @@ def generate_data(
                   We expect three files to be present in this directory: "knowledge.yaml",
                     "freeform_skills.yaml", and "grounded_skills.yaml".
     """
+    logger = setup_logger(__name__)
     generate_start = time.time()
 
+    knowledge_recipe = Recipe("src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml")
+    skills_recipe = Recipe("src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml")
```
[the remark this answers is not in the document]
bbrowning: Resolving this as that code is no longer present in this PR.

src/instructlab/sdg/generate_data.py
bbrowning: Thanks, rebased now that 180 merged and resolving this as the context is now gotten without this type of hack.

bbrowning:
Ok, I feel like this is in a point where it's good enough to review. I'll be ready to address any concerns quickly to keep round-trip time of that whole process as short as possible. After merging we'll need some follow-up PRs on top of this work (and tracked in #162) to write out recipe yaml files, add auxiliary datasets, add precomputed datasets, and remove the legacy train/messages json output formats.

src/instructlab/sdg/generate_data.py
```diff
@@ -393,7 +637,6 @@ def generate_data(
 
         if samples[0].get("document"):
```
markmc, line 638: This can use `is_knowledge`

src/instructlab/sdg/generate_data.py
```diff
@@ -260,6 +332,169 @@ def load_pipeline(yaml_basename):
...
+
+        return {"messages": messages, "metadata": metadata, "id": msg_id}
+
+    knowledge_ds = generated_dataset.map(
+        __create_qa_row, remove_columns=generated_dataset.column_names
+    )
+    return knowledge_ds
+
+
+def _build_raft_dataset(ds: Dataset, p, num_doc_in_context=4):
```
markmc, line 390: ChatGPT offers 2 possible explanations for what "raft" could mean here:

> Random Augmented Feature Training (RAFT): In machine learning, "RAFT" might stand for Random Augmented Feature Training, a technique where random augmentation is applied to the features of a dataset to improve the model's robustness and performance. The function's description suggests that additional context from random samples is added, which could be a form of data augmentation.
>
> Reference Augmented Fine-Tuning (RAFT): In NLP, RAFT might refer to adding additional context or references to enhance the training data. By incorporating context from other samples, the function could be aiming to provide a more comprehensive background, potentially improving the quality of responses generated by a language model.

Seems like it's the latter? Expanding the acronym and linking to https://aka.ms/raft-paper would help

src/instructlab/sdg/generate_data.py
```diff
@@ -402,6 +645,40 @@ def generate_data(
                 date_suffix,
             )
 
+        if is_knowledge:
```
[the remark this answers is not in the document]
markmc: Ok, take a look at https://github.com/markmc/sdg/commit/b01f1ca7d8e7af802bf12ab409aed9c0e1267fa4 in https://github.com/markmc/sdg/tree/data-mixing-mixer

src/instructlab/sdg/generate_data.py
bbrowning: Yes, good catch! I missed that when rebasing on top of the latest mmlubench changes.

src/instructlab/sdg/generate_data.py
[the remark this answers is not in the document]
bbrowning: That keeps generate_data.py much cleaner, nice! I haven't had a chance to pull and run it locally yet, but it looks like we may be missing the final call to `mixer.generate()`? Or I'm just not finding it?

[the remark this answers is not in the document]
markmc: Fixed, thank you!

src/instructlab/sdg/generate_data.py
bbrowning: I do think this in the spirit of that RAFT paper, but I'll give this a different name that doesn't require understanding the acronym. Something like `_add_extra_contexts_to_samples`  since it's adding extra contexts to each sample in the dataset.

markmc:
Looking into:

```
 File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/instructlab/sdg/generate_data.py", line 404, in generate_data
    mixer.generate()
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/instructlab/sdg/datamixing.py", line 468, in generate
    self._gen_mixed_data(
...
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/instructlab/sdg/datamixing.py", line 38, in _load_ds
    dataset = load_dataset("json", data_files=path, split="train")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/datasets/data_files.py", line 411, in resolve_pattern
    raise FileNotFoundError(error_msg)
FileNotFoundError: Unable to find '/home/runner/work/sdg/sdg/node_datasets_2024-07-24T14_23_22/compositional_skills_extraction_inference_qualitative_e2e-siblings.jsonl'
```

markmc:
> Looking into:
> 
> ```
> FileNotFoundError: Unable to find '/home/runner/work/sdg/sdg/node_datasets_2024-07-24T14_23_22/compositional_skills_extraction_inference_qualitative_e2e-siblings.jsonl'
> ```

I think it's this:

```
-        recipe.add_dataset(output_file_leaf_node, sampling_size)
+        recipe.add_dataset(output_file, sampling_size)
```

bbrowning:
> > Looking into:
> > ```
> > FileNotFoundError: Unable to find '/home/runner/work/sdg/sdg/node_datasets_2024-07-24T14_23_22/compositional_skills_extraction_inference_qualitative_e2e-siblings.jsonl'
> > ```
> 
> I think it's this:
> 
> ```
> -        recipe.add_dataset(output_file_leaf_node, sampling_size)
> +        recipe.add_dataset(output_file, sampling_size)
> ```

I hit this error locally when testing this a minute ago, and making the change above fixes things on my local machine.

src/instructlab/sdg/generate_data.py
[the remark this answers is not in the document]
bbrowning: Leaving this hardcoded for now, as a follow-up PR around recipe yaml files will including loading system prompts from those recipe yaml files.

markmc (approved):
I'm good with merging this as-is when CI passes, I think the cleanups, documentation, refactoring, scope reduction, etc. has gotten to a place which this is much more maintainable going forward. Thanks, Ben!

bbrowning:
Ok, pushing the big green merge button on this for now as it's a prerequisite for a number of other PRs. Thanks for all the great feedback, and will keep cleaning this up with the work that builds on top of this.

Issue #154, opened 2024-07-16
Allow convenient aliases to be defined for custom pipelines


markmc:
As per the [custom pipeline format design doc](https://github.com/instructlab/dev-docs/blob/main/docs/sdg/sdg-flow-yaml.md)

> A possible improvement would be to have a well-defined place on the filesystem where custom pipeline configs can be automatically loaded and included as options to the --pipeline parameter.
>
> For example, if the config format included new parameters, name: full-extended and extends: full, and the CLI discovered and loaded it automatically, we could support --pipeline full-extended without needing the additional --pipeline-extend option.
>
> /usr/share/instructlab/sdg/ is a proposed location for this as a place for a distribution of InstructLab to include pre-defined custom pipelines, at least for Linux. See the Filesystem Hierarchy Standard for more details on why this path is appropriate for this use-case.
> 
> It would also make sense to support a configuration directory for user's own custom pipeline configurations. Assuming there is a base config directory, these could go in a sdg subdirectory. There is a separate proposal that discusses a proposed configuration location: #104. Note this is separate from the distribution-provided, read-only pipelines discussed above with a different location.
> 
> If we have a location with pipeline examples then a nice to have would be to have a ilab data generate --list-pipelines.

This could greatly improve the UX with custom pipelines

markmc:
Fixed by #166

PR #166, opened 2024-07-18
Load custom pipelines from shared data dir

derekhiggins:
Allow convenient aliases to be defined for custom pipelines, to be loaded from a well known predefined directory.

Related #154

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -14,6 +14,7 @@ import time
 from datasets import Dataset
 import httpx
 import openai
+import platformdirs
 
 # First Party
 # pylint: disable=ungrouped-imports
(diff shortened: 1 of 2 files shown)


derekhiggins:
Note: for this to work we'll also need to remove this check
https://github.com/instructlab/instructlab/blob/cf2e12799a143fbd3e6d82453720d24899982b51/src/instructlab/data/generate.py#L140

src/instructlab/sdg/pipeline.py
```diff
@@ -148,3 +148,4 @@ def _parse_pipeline_config_file(pipeline_yaml):
 # This is part of the public API.
 SIMPLE_PIPELINES_PACKAGE = "instructlab.sdg.pipelines.simple"
 FULL_PIPELINES_PACKAGE = "instructlab.sdg.pipelines.full"
+CUSTOM_PIPELINES_DATA_DIR = "/usr/share/instructlab/sdg"
```
markmc, line 151: Take a look at how instructlab has started using the XDG Base Directory spec - e.g. using the `platformdirs` library

I think we would want `iter_data_dirs()` - i.e. `$XDG_DATA_HOME:$XDG_DATA_DIRS` ? (something like that)

src/instructlab/sdg/generate_data.py
```diff
@@ -164,9 +165,20 @@ def _gen_test_data(
...
 def _sdg_init(pipeline, client, model_family, model_id, num_instructions_to_generate):
     pipeline_pkg = None
-    if pipeline == "full":
+    if os.path.exists(os.path.join(CUSTOM_PIPELINES_DATA_DIR, pipeline)):
+        pipeline = os.path.join(CUSTOM_PIPELINES_DATA_DIR, pipeline)
+        _check_pipeline_dir(pipeline)
+    elif pipeline == "full":
         pipeline_pkg = FULL_PIPELINES_PACKAGE
     elif pipeline == "simple":
         pipeline_pkg = SIMPLE_PIPELINES_PACKAGE
```
markmc, line 184: It's tempting to say that these pre-defined ones should go into `/usr/share/` too ... except more like a `DATA_DIR` specific to a the virtual environment the package is installed into? Consider it if you find something obvious, but otherwise let's leave it as it is

markmc (changes requested):
Thanks @derekhiggins !

src/instructlab/sdg/pipeline.py
derekhiggins: I've updated it to search user and site data dirs.

markmc (approved):
Thanks!

src/instructlab/sdg/generate_data.py
```diff
@@ -164,29 +165,46 @@ def _gen_test_data(
...
     pipeline_pkg = None
-    if pipeline == "full":
-        pipeline_pkg = FULL_PIPELINES_PACKAGE
-    elif pipeline == "simple":
-        pipeline_pkg = SIMPLE_PIPELINES_PACKAGE
+
+    # Search for the pipeline in User and Site data directories
+    # then for a package defined pipeline
+    # and finally pipelines referenced by absolute path
+    pd = platformdirs.PlatformDirs(appname="sdg", multipath=True)
```
markmc, line 182: Oh, can we put this under `instructlab/sdg` ?

markmc (changes requested):

src/instructlab/sdg/generate_data.py
```diff
@@ -236,7 +254,8 @@ def generate_data(
     use the SDG library constructs directly, and this function will likely be removed.
 
     Args:
-        pipeline: This argument may be either an alias defined by the sdg library ("simple", "full"),
+        pipeline: This argument may be either an alias defined in a user or site "data directory"
```
russellb, line 257: What's the easiest way for someone to know what the "user or site data directory" is? I'm thinking ahead to what we'd put in the CLI help text.

markmc: It's the usual `$XDG_DATA_HOME:$XDG_DATA_DIRS` semantics?

src/instructlab/sdg/generate_data.py
derekhiggins: done

src/instructlab/sdg/generate_data.py
```diff
@@ -164,29 +165,46 @@ def _gen_test_data(
...
     pipeline_pkg = None
-    if pipeline == "full":
-        pipeline_pkg = FULL_PIPELINES_PACKAGE
-    elif pipeline == "simple":
-        pipeline_pkg = SIMPLE_PIPELINES_PACKAGE
+
+    # Search for the pipeline in User and Site data directories
+    # then for a package defined pipeline
+    # and finally pipelines referenced by absolute path
+    pd = platformdirs.PlatformDirs(appname="instructlab/sdg", multipath=True)
```
markmc, line 182: Think you need `os.path.join("instructlab", "sdg")` here? (I honestly don't recall whether we're trying to support Windows though)

russellb (approved):

PR #182, opened 2024-07-21
Add 4 new pipeline blocks

### LINKED ISSUE #175: Add DuplicateColumns block type
See https://github.com/aakankshaduggal/sdg/pull/11/files

### LINKED ISSUE #176: Add RenameColumns block type
see https://github.com/aakankshaduggal/sdg/pull/11/files

### LINKED ISSUE #177: Add FlattenColumns block type
See https://github.com/aakankshaduggal/sdg/pull/11/files

### LINKED ISSUE #178: Add SetToMajorityValue block type
See https://github.com/aakankshaduggal/sdg/pull/11/files

derekhiggins:
"DuplicateColumns", "FlattenColumnsBlock",
"RenameColumns", "SetToMajorityValue"

Closes #175
Closes #176
Closes #177
Closes #178

### DIFF
--- a/src/instructlab/sdg/utilblocks.py
+++ b/src/instructlab/sdg/utilblocks.py
@@ -95,3 +95,82 @@ class CombineColumnsBlock(Block):
             self.separator,
             self.ctx.dataset_num_procs,
         )
+
+
+class FlattenColumnsBlock(Block):
+    """Melt/transform a data from a wide format to a long format see pandas.melt for a description
+
+    Args:
+            var_cols (list): Column(s) to unpivot. All other columns are set to use as identifier variables.
+            value_name (str): Name to use for the ‘value’ column, can’t be an existing column label.
+            var_name (str):  Name to use for the ‘variable’ column. If None it uses frame.columns.name or ‘variable’.
+    """
+
+    def __init__(
+        self, ctx, pipe, block_name: str, var_cols: list, value_name: str, var_name: str
+    ) -> None:
+        super().__init__(ctx, pipe, block_name)
+        self.var_cols = var_cols
+        self.value_name = value_name
+        self.var_name = var_name
+
+    def generate(self, samples: Dataset) -> Dataset:
+        df = samples.to_pandas()
+        id_cols = [col for col in samples.column_names if col not in self.var_cols]
+        flatten_df = df.melt(
+            id_vars=id_cols,
+            value_vars=self.var_cols,
+            value_name=self.value_name,
+            var_name=self.var_name,
+        )
+        return Dataset.from_pandas(flatten_df)
+
+
+class DuplicateColumnsBlock(Block):
+    def __init__(self, ctx, pipe, block_name: str, columns_map: dict) -> None:
+        """Create duplicate of columns specified in column map.
+
+        Args:
+            columns_map (dict): mapping of existing column to new column names
+        """
+        super().__init__(ctx, pipe, block_name)
+        self.columns_map = columns_map
+
+    def generate(self, samples: Dataset):
+        for col_to_dup in self.columns_map:
+            samples = samples.add_column(
+                self.columns_map[col_to_dup], samples[col_to_dup]
+            )
+        return samples
+
+
+class RenameColumnsBlock(Block):
+    def __init__(self, ctx, pipe, block_name: str, columns_map: dict) -> None:
+        """Rename dataset columns.
+
+        Args:
+            columns_map (dict): mapping of existing column to new column names
+        """
+        self.columns_map = columns_map
+        super().__init__(ctx, pipe, block_name)
+
+    def generate(self, samples: Dataset):
+        samples = samples.rename_columns(self.columns_map)
+        return samples
+
+
+class SetToMajorityValueBlock(Block):
+    """Set the value of the specified column to the most common value (the mode)
+
+    Args:
+        col_name (str): the column to find the "mode" of and then set universally
+    """
+
+    def __init__(self, ctx, pipe, block_name: str, col_name) -> None:
+        self.col_name = col_name
+        super().__init__(ctx, pipe, block_name)
+
+    def generate(self, samples: Dataset):
+        samples = samples.to_pandas()
+        samples[self.col_name] = samples[self.col_name].mode()[0]
+        return Dataset.from_pandas(samples)
--- a/tests/test_utilblocks.py
+++ b/tests/test_utilblocks.py
@@ -1,0 +1,101 @@
+# Standard
+from unittest.mock import MagicMock, patch
+import unittest
+
+# Third Party
+from datasets import Dataset, Features, Value
+
+# First Party
+from src.instructlab.sdg.utilblocks import (
+    DuplicateColumnsBlock,
+    FlattenColumnsBlock,
+    RenameColumnsBlock,
+    SetToMajorityValueBlock,
+)
+
+
+class TestUtilBlock(unittest.TestCase):
+    def setUp(self):
+        self.mock_ctx = MagicMock()
+        self.mock_pipe = MagicMock()
+
+        self.dataset = Dataset.from_dict(
+            {
+                "name": ["Tomato", "Rose", "Carrot", "Basil"],
+                "type": ["Fruit", "Flower", "Vegetable", "Herb"],
+                "edible": ["yes", "no", "yes", "yes"],
+            }
+        )
+
+    def test_dup(self):
+        """Test that the DuplicateColumnsBlock Block can correctly duplicate a column"""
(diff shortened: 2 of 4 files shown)


tests/test_utilblocks.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+
+    def test_dup(self):
+        dup_block = DuplicateColumns(
+            self.mock_ctx, self.mock_pipe, "blockname", columns_map={"edible": "caneat"}
+        )
+        assert "caneat" not in self.dataset.column_names
+        new_samples = dup_block.generate(self.dataset)
+        assert "edible" in new_samples.column_names
+        assert "caneat" in new_samples.column_names
```
russellb, line 37: Another test here could be to validate that the two columns have equivalent values, as well.

tests/test_utilblocks.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_renamecolumns(self):
+        rename_block = RenameColumns(
+            self.mock_ctx, self.mock_pipe, "blockname", columns_map={"edible": "caneat"}
+        )
+        new_samples = rename_block.generate(self.dataset)
+        assert "caneat" in new_samples.column_names
+        assert "edible" not in new_samples.column_names
```
russellb, line 45: same here -- would be great to make sure the values are as expected, too

tests/test_utilblocks.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_setmajority(self):
+        mv_block = SetToMajorityValue(
+            self.mock_ctx, self.mock_pipe, "blockname", col_name="edible"
+        )
+
+        new_samples = mv_block.generate(self.dataset)
+
+        for row in new_samples.select_columns(["edible"]):
+            assert row.get("edible") == "yes"
```
russellb, line 55: I don't really understand this one. Does the block need to support strings, or is it really aimed at numeric values only? How should someone expect it to behave on strings if we need to keep string support?

src/instructlab/sdg/utilblocks.py
```diff
@@ -95,3 +95,68 @@ def generate(self, samples: Dataset) -> Dataset:
             self.separator,
             self.ctx.dataset_num_procs,
         )
+
+
+class FlattenColumnsBlock(Block):
+    def __init__(
+        self, ctx, pipe, block_name: str, var_cols: list, value_name: str, var_name: str
+    ) -> None:
```
russellb, line 103: Can you add a docstring for this one?

src/instructlab/sdg/utilblocks.py
```diff
@@ -95,3 +95,68 @@ def generate(self, samples: Dataset) -> Dataset:
...
+        self.columns_map = columns_map
+        super().__init__(ctx, pipe, block_name)
+
+    def generate(self, samples: Dataset):
+        samples = samples.rename_columns(self.columns_map)
+        return samples
+
+
+class SetToMajorityValue(Block):
+    def __init__(self, ctx, pipe, block_name: str, col_name) -> None:
```
russellb, line 155: and a docstring for this one?

russellb (dismissed):
Thanks for working on this!

I'm having trouble understanding the use of two of the blocks. Docstrings would help me, I hope.

tests/test_utilblocks.py
derekhiggins: I don't think its confined to numeric values (although I haven't see it used) , it selects the most common value and sets all column values to match it

src/instructlab/sdg/utilblocks.py
russellb: you marked this resolved, but i don't see the docstring?

russellb:
@derekhiggins one last thing -- it would be good to verify in our downstream repository that these are actually used, just in case any were added but didn't end up getting used. They should be either in a file checked in or an open merge request.

src/instructlab/sdg/utilblocks.py
derekhiggins: Opps sorry I didn't look closely enough, I added 4 docstrings to the tests, I thought that was where you asked for them, will fix.

derekhiggins:
> @derekhiggins one last thing -- it would be good to verify in our downstream repository that these are actually used, just in case any were added but didn't end up getting used. They should be either in a file checked in or an open merge request.

hmm, that I can see none of them are in use (looking in the research branch, its PR's and sdg-config), is there somewhere else I should look?

russellb:
> > @derekhiggins one last thing -- it would be good to verify in our downstream repository that these are actually used, just in case any were added but didn't end up getting used. They should be either in a file checked in or an open merge request.
> 
> hmm, that I can see none of them are in use (looking in the research branch, its PR's and sdg-config), is there somewhere else I should look?

If you don’t see it anywhere, including the MRs against the downstream sdg-config repo then let’s hold this. I would ask for confirmation of where it’s used. 

If folks really want them for experimentation I guess I’m ok with that too. They’re small enough. I just want to confirm they’re still useful somewhere …

markmc:
> I've one question, all blocks currently follow a naming scheme where the end in "Block", @aakankshaduggal should/can we rename these to follow the same scheme i.e. DuplicateColumns -> DuplicateColumnsBlock RenameColumns -> RenameColumnsBlock SetToMajorityValue -> SetToMajorityValueBlock

Yes, absolutely ... we should rename them all to drop the "Block" suffix or use the suffix on any new ones. The latter is certainly less invasive.

src/instructlab/sdg/configs/skills/contexts.yaml
```diff
@@ -19,5 +19,3 @@ generation: |
 
 start_tags: ["[Start of Context]"]
 end_tags: ["[End of Context]"]
-
-sys_prompt: You are a very knowledgeable AI Assistant that will faithfully assist the user with their task.
```
markmc, line 23: I think this is unrelated? If so, I'd prefer to deal with it separately. See #184

src/instructlab/sdg/llmblock.py
```diff
@@ -189,7 +189,7 @@ def generate(self, samples: Dataset) -> Dataset:
         valid_samples = []
 
         for sample in samples:
-            if self._validate(self.prompt_template, sample):
+            if self.validate(self.prompt_template, sample):
```
markmc, line 192: This looks like a bugfix for ConditionalLLMBlock, orthogonal to these new blocks

I've filed #186 - let's remove it from this PR

markmc (changes requested):
Couple of requests to move stuff out of this PR, haven't reviewed the new blocks yet

derekhiggins:
> Couple of requests to move stuff out of this PR, haven't reviewed the new blocks yet

Both removed, they were part of the original PR

src/instructlab/sdg/utilblocks.py
```diff
@@ -95,3 +95,82 @@ def generate(self, samples: Dataset) -> Dataset:
...
+    def generate(self, samples: Dataset) -> Dataset:
+        df = samples.to_pandas()
+        id_cols = [col for col in samples.column_names if col not in self.var_cols]
+        flatten_df = df.melt(
+            id_vars=id_cols,
+            value_vars=self.var_cols,
+            value_name=self.value_name,
+            var_name=self.var_name,
+        )
+        return Dataset.from_pandas(flatten_df)
```
markmc, line 126: This likely needs a `reset_index()` call - but let's resolve that with a new utility function: https://github.com/instructlab/sdg/issues/190

src/instructlab/sdg/utilblocks.py
```diff
@@ -95,3 +95,82 @@ def generate(self, samples: Dataset) -> Dataset:
...
+    """
+
+    def __init__(self, ctx, pipe, block_name: str, col_name) -> None:
+        self.col_name = col_name
+        super().__init__(ctx, pipe, block_name)
+
+    def generate(self, samples: Dataset):
+        samples = samples.to_pandas()
+        samples[self.col_name] = samples[self.col_name].mode()[0]
+        return Dataset.from_pandas(samples)
```
markmc, line 176: https://github.com/instructlab/sdg/issues/190 again

tests/test_utilblocks.py
```diff
@@ -0,0 +1,101 @@
...
+# First Party
+from src.instructlab.sdg.utilblocks import (
+    DuplicateColumnsBlock,
+    FlattenColumnsBlock,
+    RenameColumnsBlock,
+    SetToMajorityValueBlock,
+)
+
+
+class TestUtilBlock(unittest.TestCase):
```
markmc, line 17: Thank you for adding tests!

markmc (approved):
I'm happy to follow-up on the pandas->dataset conversion issue

PR #194, opened 2024-07-24
 Remove calls to logging.basicConfig on import

### LINKED ISSUE #196: Library calls logging.basicConfig on import
`instructlab.sdg` calls `logging.basicConfig` on import. Libraries should not configure logging themselves. It's the job of the application or main script to configure logging and set up log handlers according to their needs. `logging.basicConfig` is a one-shot API. The second and any following call are a no-op, unless it is called with `force=True`.

https://github.com/instructlab/sdg/blob/d7009812e92a6026c74249ecd30300e1c4354c38/src/instructlab/sdg/logger_config.py#L9-L18

https://github.com/instructlab/sdg/blob/d7009812e92a6026c74249ecd30300e1c4354c38/src/instructlab/sdg/eval_data.py#L13-L16

I recommend to remove `setup_logging` and just use `logging.getLogger` in the code.

tiran:
Fixes #196 

commit 81a69bb633a0ef362538e06b85e7c0e6b85f7880

```
    Remove calls to logging.basicConfig on import
    
    Fixes #196
    
    instructlab.sdg calls logging.basicConfig on import.
    
    Libraries should not configure logging themselves. It's the job of the
    application or main script to configure logging and set up log
    handlers according to their needs. logging.basicConfig is a one-shot
    API. The second and any following call are a no-op, unless it is
    called with force=True.
    
    Remove setup_logging and just use logging.getLogger in the code.
    
    Co-authored-by: Christian Heimes <cheimes@redhat.com>
    Signed-off-by: Mark McLoughlin <markmc@redhat.com>
```

commit 76a8624a7d03c12bd84c68938eaf4a836719b81c

```
    Use module level logger
    
    `generate_data` now uses a module-level logger like the rest of the
    code. The `logger` parameter is now optional.
    
    Signed-off-by: Christian Heimes <cheimes@redhat.com>
```


tiran:
@derekhiggins please take a look

derekhiggins:
hmm, we also have setup_logger in https://github.com/instructlab/sdg/blob/main/src/instructlab/sdg/logger_config.py 
I wonder was the intent to converge on using that?

tiran:
Is `instructlab.sdg` also designed to be used as a standalone application? Or is it only a library?

tiran:
@derekhiggins I filed https://github.com/instructlab/sdg/issues/196 for the logging problem.

markmc (approved):

Issue #201, opened 2024-07-24
Pull taxonomy precomputed dataset from hugging face


aakankshaduggal:
In order to improve skills training, we require a precomputed dataset that is going to be mixed with the new synthetically generated dataset. This is mostly used during full training via instructlab/training, and not as important for the simpler legacy training in instructlab/instructlab.

The taxonomy precomputed dataset is hosted on hugging face -- https://huggingface.co/datasets/instructlab/InstructLabCommunity 

We need a way to incorporate this dataset from HuggingFace and mix it with the synthetic generated data during an `ilab data generate`. One proposal is at https://github.com/aakankshaduggal/sdg/pull/20, but after some discussion at https://github.com/instructlab/sdg/pull/203#issuecomment-2250444499 we want to implement this in a bit different way so that we're not hitting HuggingFace silently/automatically but instead with an explicit step to download the precomputed dataset.

Users will have to download the dataset and place it in an appropriate cache directory. Potentially, there could be an `ilab data download` command to do this with a nicer user experience that asking them to manually download it to the appropriate directory.

Allow the name and/or path to the precomputed dataset to be supplied with e.g. `ilab data generate --skills-dataset=` and ilab would construct a simple skills recipe (in memory) and pass it to the library. We could limit this just to supplying a single precomputed skills dataset, or perhaps we want to allow the user to specify a list of precomputed skills and/or knowledge datasets on the command line?

We'll want to test this precomputed dataset in our e2e CI, which means ensuring we download and cache the dataset there so it's available at data generation time.

[Edited by @bbrowning to incorporate changes from https://github.com/instructlab/sdg/pull/203#issuecomment-2250444499 as well as in-person discussions with @aakankshaduggal ].

bbrowning:
still relevant, as we aren't yet mixing in the community precomputed dataset

Issue #185, opened 2024-07-23
Write Recipe files during data mixing


bbrowning:
The writing of Recipe files was removed from #163 to reduce scope, but it's a requirement to resolve #162. We need to add back the functionality that writes the recipe yaml files to disk within the generated data output directory, as that becomes one of the necessary artifacts to document what went into the mixed dataset.

This may also be needed for #171, depending on how that gets solved.

bbrowning:
Resolved by #203

Issue #171, opened 2024-07-19
Add precomputed dataset to skills data generation


bbrowning:
As a followup to #163, we need to figure out the right way to wire a precomputed dataset into the skills data generation. One example of such a dataset is https://github.com/instructlab/training/blob/9fdeb87820d5000f7be60a199c4e24aec725772e/sample-data/train_all_pruned_SDG.jsonl , however downstream uses of InstructLab, CI, or other reasons will warrant the ability to change this out in some way.

The initial implementation dropped out of scope from #163 had a placeholder in `src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml` like below:

```yaml
datasets:
  - path: <path_to_dataset>
    sampling_size: 1.0
 ```

Discussing this with the community, @shivchander suggested we may want to pull this dataset from somewhere like HuggingFace. So, there's work to be done to figure out where the dataset should live, how the user gets it (explicitly pulls, implicitly pulls as needed, caching, etc), and how downstream uses, CI, or other scenarios will overwrite this precomputed dataset with their own.

bbrowning:
The current iteration of #163 does not read recipe files at all, so if recipe files are how the precomputed datasets will be read then we'll want to plumb some version of reading from them back in once that lands.

Also, we'll likely need the ability to specify the system prompt that was used / to use with each precomputed dataset. Previously this was read from recipe files, so that may need to be added back as well if configurable system prompts are needed to properly use precomputed datasets.

bbrowning:
still relevant

PR #203, opened 2024-07-24
Introduce data mixing recipe yaml files

bbrowning:
This introduces Recipe yaml files, which are used both as an input into the data mixing process and as an output of the process.

As an input, we have some default recipe files that specify any precomputed datasets that should be mixed with data from new skills when generating the overall mix of samples that will be sent to the training process. In our specific case, we're adding in a single precomputed dataset of `instructlab/InstructLabCommunity` that gets pulled from HuggingFace and mixed with new skill samples.

As an output of the data generation process, we write recipe yamls to document which datasets were mixed together and in what proportions along with the system prompt that was used during the generation. Here's an example of a recipe yaml put into the output directory after running data generation:

```yaml
datasets:
- path: instructlab/InstructLabCommunity
  sampling_size: 1.0
- path: node_datasets_2024-07-25T10_27_12/knowledge_tonsils_overview_e2e-tonsils_p10.jsonl
  sampling_size: 1.0
metadata:
  sys_prompt: "I am, Red Hat\xAE Instruct Model based on Granite 7B, an AI language\
    \ model developed by Red Hat and IBM Research, based on the Granite-7b-base language\
    \ model. My primary function is to be a chat assistant."
```

While this change is relatively small, it came from some much larger PRs that were co-authored by multiple people so giving credit to all of those here.

Parts of this are extracted and rebased from
https://github.com/aakankshaduggal/sdg/pull/4
https://github.com/aakankshaduggal/sdg/pull/20

Refs #162, #171, #185, #201.

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -7,6 +7,7 @@ import uuid
 
 # Third Party
 from datasets import Dataset, concatenate_datasets, load_dataset
+import yaml
 
 # First Party
 from instructlab.sdg.logger_config import setup_logger
@@ -27,18 +28,12 @@ def _adjust_train_sample_size(ds: Dataset, num_samples: int):
     return pandas.dataset_from_pandas_dataframe(df)
 
 
-def _load_ds(path, sampling_size, num_proc):
+def _sample_ds(dataset, sampling_size, num_proc):
     """
-    Load a dataset from the given file path and select sampling_size
-    number/ratio of samples from it, ensuring the loaded dataset has only
-    ALLOWED_COLS columns in it with any additional columns moved to the
-    metadata section.
+    Select sampling_size number/ratio of samples from a dataset, ensuring
+    the returned dataset has only ALLOWED_COLS columns in it with any
+    additional columns moved to the metadata section.
     """
-    logger.info(f"Loading dataset from {path} ...")
-    dataset = load_dataset("json", data_files=path, split="train")
-    logger.info(f"Dataset columns: {dataset.column_names}")
-    logger.info(f"Dataset loaded with {len(dataset)} samples")
-
     if sampling_size != 1.0:
         if isinstance(sampling_size, int):
             num_samples = sampling_size
@@ -94,29 +89,63 @@ class Recipe:
     """
 
     def __init__(
-        self, initial_datasets: Optional[list] = None, sys_prompt: Optional[str] = ""
+        self, recipe_path: Optional[str] = None, sys_prompt: Optional[str] = ""
     ):
-        self.recipe = {
-            "datasets": initial_datasets or [],
-            "sys_prompt": sys_prompt,
-        }
-        self.sys_prompt = self.recipe.get("sys_prompt", "")
+        self.recipe_path = recipe_path or ""
+        self.sys_prompt = sys_prompt
+
+        # Defaults if no recipe path given or these values don't
+        # exist in the given recipe file
+        self.datasets = []
+        if recipe_path is not None:
+            recipe = self._load_recipe()
+            if "datasets" in recipe:
+                self.datasets = recipe["datasets"]
+
         self.dataset_added = False
+
+    def _load_recipe(self):
+        with open(self.recipe_path, encoding="utf-8") as fp:
+            return yaml.safe_load(fp)
+
+    def _load_ds(self, path):
+        """
+        Load a dataset from the given location. If a jsonl file is
+        given, we load the dataset from disk. Otherwise, we load the
+        path given from HuggingFace. Relative paths are resolved
+        respective to the directory the recipe yaml itself resides in.
+        """
+        if not os.path.isabs(path):
+            path = os.path.join(os.path.dirname(self.recipe_path), path)
+        logger.info(f"Loading dataset from {path} ...")
+        dataset = load_dataset("json", data_files=path, split="train")
+        logger.info(f"Dataset columns: {dataset.column_names}")
+        logger.info(f"Dataset loaded with {len(dataset)} samples")
+        return dataset
+
+    def _load_and_sample_datasets(self, num_proc):
+        """
+        Load and sample all the datasets in this recipe, taking
+        into account the desired sampling size from each individual
+        dataset to control the overall mix of samples in the final
+        dataset.
+        """
+        return [
+            _sample_ds(
+                self._load_ds(dataset["path"]), dataset["sampling_size"], num_proc
+            )
+            for dataset in self.datasets
+        ]
 
     def _create_mixed_dataset(self, num_proc):
         """
-        Create the mixed dataset from its list of included datasets, taking
-        into account the desired sampling size from each individual dataset
-        to control the overall mix of samples in the final dataset.
+        Create the final mixed dataset by loading, sampling, and
+        concatenating all datasets in this recipe
         """
         if not self.dataset_added:
             logger.error("No dataset added to the recipe")
 
-        mixed_ds = [
-            _load_ds(dataset["path"], dataset["sampling_size"], num_proc)
-            for dataset in self.recipe["datasets"]
-        ]
-
+        mixed_ds = self._load_and_sample_datasets(num_proc)
         mixed_ds = concatenate_datasets(mixed_ds)
         mixed_ds = mixed_ds.map(
             _add_system_message,
@@ -143,7 +172,20 @@ class Recipe:
                            of the samples, and so on.
         """
         self.dataset_added = True
-        self.recipe["datasets"].append({"path": path, "sampling_size": sampling_size})
+        self.datasets.append({"path": path, "sampling_size": sampling_size})
+
+    def save_recipe(self, output_path):
+        recipe = {
+            "datasets": self.datasets,
+            "metadata": {"sys_prompt": self.sys_prompt},
+        }
+        with open(output_path, "w", encoding="utf-8") as fp:
+            yaml.dump(recipe, fp)
+        # Update this instance's recipe_path to reflect the path we
+        # just saved it to so that any subsequent loading of datasets
+        # (like via save_mixed_dataset) pulls from relative to the
+        # saved recipe_path.
+        self.recipe_path = output_path
 
     def save_mixed_dataset(self, output_path, num_proc):
         """
@@ -398,17 +440,33 @@ class DataMixer:
     # once.
     NUM_SYNTH_SKILLS = 30
 
-    def __init__(self, output_dir, date_suffix, sys_prompt, num_procs):
+    def __init__(self, data_dirs, output_dir, date_suffix, sys_prompt, num_procs):
+        self.data_dirs = data_dirs
         self.output_dir = output_dir
         self.sys_prompt = sys_prompt
         self.date_suffix = date_suffix
         self.num_procs = num_procs
 
-        self.knowledge_recipe = Recipe(sys_prompt=self.sys_prompt)
-        self.skills_recipe = Recipe(sys_prompt=self.sys_prompt)
+        self.knowledge_recipe = self._load_default_recipe("knowledge.yaml")
+        self.skills_recipe = self._load_default_recipe("skills.yaml")
 
+        self.output_file_knowledge_recipe = f"knowledge_recipe_{date_suffix}.yaml"
+        self.output_file_skills_recipe = f"skills_recipe_{date_suffix}.yaml"
         self.output_file_mixed_knowledge = f"knowledge_train_msgs_{date_suffix}.jsonl"
         self.output_file_mixed_skills = f"skills_train_msgs_{date_suffix}.jsonl"
+
+    def _load_default_recipe(self, yaml_basename):
+        """
+        Load a default system recipe from e.g. /usr/share/instructlab/sdg/default_data_recipes
+        if it exists, otherwise return an empty recipe.
+        """
+        for d in self.data_dirs:
+            default_recipe_path = os.path.join(d, "default_data_recipes", yaml_basename)
+            if os.path.exists(default_recipe_path):
+                return Recipe(
+                    recipe_path=default_recipe_path, sys_prompt=self.sys_prompt
+                )
+        return Recipe(sys_prompt=self.sys_prompt)
 
     def _gen_leaf_node_data(
         self, leaf_node_data, recipe, output_file_leaf_node, sampling_size=1.0
@@ -420,7 +478,7 @@ class DataMixer:
         """
         output_file = os.path.join(self.output_dir, output_file_leaf_node)
         leaf_node_data.to_json(output_file, orient="records", lines=True)
-        recipe.add_dataset(output_file, sampling_size)
+        recipe.add_dataset(output_file_leaf_node, sampling_size)
 
     def collect(self, leaf_node_path, new_generated_data, is_knowledge):
         if is_knowledge:
@@ -459,20 +517,27 @@ class DataMixer:
                 sampling_size=self.NUM_SYNTH_SKILLS,
             )
 
-    def _gen_mixed_data(self, recipe, output_file_mixed):
+    def _gen_mixed_data(self, recipe, output_file_recipe, output_file_data):
         """
         Mix the generated leaf node data into a single dataset and write it to
         disk. The heavy lifting is delegated to the Recipe class.
         """
         if recipe.dataset_added:
+            full_recipe_path = os.path.join(self.output_dir, output_file_recipe)
+            recipe.save_recipe(full_recipe_path)
             recipe.save_mixed_dataset(
-                os.path.join(self.output_dir, output_file_mixed),
+                os.path.join(self.output_dir, output_file_data),
                 self.num_procs,
             )
 
     def generate(self):
-        self._gen_mixed_data(self.knowledge_recipe, self.output_file_mixed_knowledge)
+        self._gen_mixed_data(
+            self.knowledge_recipe,
+            self.output_file_knowledge_recipe,
+            self.output_file_mixed_knowledge,
+        )
         self._gen_mixed_data(
             self.skills_recipe,
+            self.output_file_skills_recipe,
             self.output_file_mixed_skills,
         )
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -240,6 +240,20 @@ def _sdg_init(ctx, pipeline):
     )
 
 
+def _mixer_init(ctx, output_dir, date_suffix):
+    pd = platformdirs.PlatformDirs(
+        appname=os.path.join("instructlab", "sdg"), multipath=True
+    )
+    data_dirs = list(pd.iter_data_dirs())
+    return DataMixer(
+        data_dirs,
+        output_dir,
+        date_suffix,
+        _SYS_PROMPT,
+        ctx.dataset_num_procs,
+    )
+
+
 # This is part of the public API, and used by instructlab.
 # TODO - parameter removal needs to be done in sync with a CLI change.
 # pylint: disable=unused-argument
@@ -341,7 +355,7 @@ def generate_data(
 
     mmlu_bench_pipe = mmlubench_pipe_init(ctx)
 
-    mixer = DataMixer(output_dir, date_suffix, _SYS_PROMPT, ctx.dataset_num_procs)
+    mixer = _mixer_init(ctx, output_dir, date_suffix)
 
     if console_output:
         logger.info(
--- a/tests/test_datamixing.py
+++ b/tests/test_datamixing.py
@@ -2,11 +2,31 @@
 Unit tests for the top-level datamixing module.
 """
 
+# Standard
+from importlib import resources
+from unittest.mock import patch
+import os
+
 # Third Party
 from datasets import Dataset
 
 # First Party
-from instructlab.sdg.datamixing import _add_extra_contexts_to_samples
+from instructlab.sdg.datamixing import DataMixer, Recipe, _add_extra_contexts_to_samples
+
+# We mock out the actual things that use num_procs anyway, but just
+# for a consistent value in the tests...
+TEST_NUM_PROCS = 4
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "testdata")
+TEST_RECIPE_PATH = os.path.join(TEST_DATA_DIR, "relative_path_recipe.yaml")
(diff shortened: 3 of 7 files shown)


aakankshaduggal (approved):
Thanks! This looks great @bbrowning 🚢

src/instructlab/sdg/datamixing.py
```diff
@@ -397,9 +424,17 @@ def __init__(self, output_dir, date_suffix, sys_prompt, num_procs):
         self.date_suffix = date_suffix
         self.num_procs = num_procs
 
-        self.knowledge_recipe = Recipe(sys_prompt=self.sys_prompt)
-        self.skills_recipe = Recipe(sys_prompt=self.sys_prompt)
+        self.knowledge_recipe = Recipe(
+            recipe_path="src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml",
```
markmc, line 428: Hmm, this is relative to CWD? How do we know the CWD is `instructlab/sdg` ? What when `ilab` is consuming the library as a wheel?

For loading pipeline configs, we load it from the Python package: https://github.com/instructlab/sdg/blob/cbee53c2453876f2602c86ef3e5a803689604716/src/instructlab/sdg/generate_data.py#L229

src/instructlab/sdg/datamixing.py
```diff
@@ -93,16 +101,31 @@ class Recipe:
...
 
-    def __init__(
-        self, initial_datasets: Optional[list] = None, sys_prompt: Optional[str] = ""
-    ):
-        self.recipe = {
-            "datasets": initial_datasets or [],
-            "sys_prompt": sys_prompt,
-        }
-        self.sys_prompt = self.recipe.get("sys_prompt", "")
+    def __init__(self, recipe_path, sys_prompt: Optional[str] = ""):
```
markmc, line 104: Make recipe_path optional, default to None?

src/instructlab/sdg/datamixing.py
```diff
@@ -93,16 +101,31 @@ class Recipe:
...
+        self.recipe_path = recipe_path
+        if recipe_path != "":
```
markmc, line 106: Handle recipe_path = None

src/instructlab/sdg/datamixing.py
```diff
@@ -93,16 +101,31 @@ class Recipe:
...
+            self.recipe = self._load_recipe()
+        else:
+            self.recipe = {
+                "datasets": [],
+            }
+        if sys_prompt != "":
+            # Overwrite whatever sys_prompt was present in the recipe
+            # with the current one in use in the generation pipeline,
+            # as that's the one that is being used for generation and
+            # thus the one we'll write back out later.
```
markmc, line 116: I guess I'd imagine this being the reverse? The one in the generation pipeline would be the default if there's none specified in the file?

src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml
```diff
@@ -0,0 +1 @@
+datasets: []
```
markmc, line 1: Let's just remove this empty file and just create an empty Recipe() in the code?

markmc (changes requested):
Some minor comments, but I think the point about what `recipe_path` is relative to is probably important - e.g. has this been tested with `ilab data generate` ?

src/instructlab/sdg/datamixing.py
```diff
@@ -29,13 +30,20 @@ def _adjust_train_sample_size(ds: Dataset, num_samples: int):
...
     metadata section.
+
+    If a jsonl file is given, we load the dataset from disk. Otherwise, we
+    load the path given from HuggingFace.
     """
-    logger.info(f"Loading dataset from {path} ...")
-    dataset = load_dataset("json", data_files=path, split="train")
+    if path.endswith(".jsonl"):
+        logger.info(f"Loading dataset from {path} ...")
+        dataset = load_dataset("json", data_files=path, split="train")
```
markmc, line 43: You know, it's probably important we support relative paths here too?

e.g. if our default recipe referenced a `.jsonl` file as a dataset, we'd need it to be relative to some install path

src/instructlab/sdg/datamixing.py
bbrowning: That's a great catch on the CWD, as I know I've only tested from within an instructlab/sdg clone folder and I suspect that's how others are testing, which is why this happens to work as-is but would not reliably work. I do think we should be able to load and reference these files like the configs instead of the more naive way here.

src/instructlab/sdg/datamixing.py
bbrowning: We might be able to assume these are always absolute paths at the moment, as the recipes are written as absolute paths and any downstream usage of this I've heard assumed that these files would be in a known absolute path. But you bring up a good point that we should test that assumption, as well as an overall point about portability of the recipe file if the paths to jsonl files aren't relative to the data generation output directory, at least for the generated recipe yaml files and outputs.

src/instructlab/sdg/datamixing.py
bbrowning: Ok, after some back-and-forth discussion, it sounds like we don't really ever need to load the system prompt from these recipe files. It gets written to them to document what was used, but we don't need to load it back in. I'll make that change and unconditionally replace this with the system prompt passed in to the function call.

src/instructlab/sdg/datamixing.py
bbrowning: I had some offline discussion with the team and we agreed it likely makes more sense for these to be relative paths, relative to the generation output directory. So, my example recipe output above would change from:

```yaml
datasets:
- path: instructlab/InstructLabCommunity
  sampling_size: 1.0
- path: /home/my_user/.local/share/instructlab/datasets/node_datasets_2024-07-24T15_07_58/knowledge_tonsils_overview_e2e-tonsils_p10.jsonl
  sampling_size: 1.0
sys_prompt: "I am, Red Hat\xAE Instruct Model based on Granite 7B, an AI language\
  \ model developed by Red Hat and IBM Research, based on the Granite-7b-base language\
  \ model. My primary function is to be a chat assistant."
```
to:

```yaml
datasets:
- path: instructlab/InstructLabCommunity
  sampling_size: 1.0
- path: node_datasets_2024-07-24T15_07_58/knowledge_tonsils_overview_e2e-tonsils_p10.jsonl
  sampling_size: 1.0
sys_prompt: "I am, Red Hat\xAE Instruct Model based on Granite 7B, an AI language\
  \ model developed by Red Hat and IBM Research, based on the Granite-7b-base language\
  \ model. My primary function is to be a chat assistant."
```

For robustness sake, they should probably accept either absolute or relative with the relative option becoming relative to the output directory.

bbrowning:
Marking as draft while I work on absolute vs relative paths a bit, as well as address the other feedback here.

src/instructlab/sdg/datamixing.py
markmc: That's pretty bizarre too though ... you load a config file and some of it gets ignored?

Maybe add a 'metadata' section (or similar) to hold `sys_prompt` ... the semantics would be that anything in that section is info written out by the mixer, doesn't influence the mixing process?

src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml
```diff
@@ -0,0 +1,3 @@
+datasets:
+  - path: instructlab/InstructLabCommunity
```
markmc, line 2: This worries me a bit ... downloading stuff from Hugging Face, down in the depths of the library

UX-wise, it feels a little like 'ilab model download' ... something you would do as part of setting up your workflow

It's like, you could plausibly do `data generate` and `model train` iterations on your laptop on a plane with no Internet, except for this?

Consider also a unit test like #208 ... all of this datamixing code is self-contained, apart from this external dependency on Hugging Face?

If we had some sort if `ilab data setup` command ... and it downloaded this dataset into local cache ... what else might that command do? e.g. validate that the teacher models required are available from the configured OpenAI endpoint?

I'm not really suggesting this as a blocker for merging ... but I would like to file an issue which acknowledges that this isn't ideal and has some sort of plan for resolving it

bbrowning: I agree and share some of these concerns, as it's a download that just happens behind the scenes automatically. I know some downstream users intend to replace this HuggingFace with a file on disk as the preconfigured dataset, which is why the code supports loading from either. It's a 32MB file, which is not necessarily too large to just include as part of our python package, although that may not be ideal either.

I'd prefer to not introduce a new command for this right now, as depending on how we address this it may not require a separate command to download (like if we just ship it in some python package). Or, it may need a separate command. Either way I think a proper fix warrants a longer discussion than what we'll want for this release.

Is there a better alternative we should go with for this release? Do you think just placing the file directly in the git repository and including it as part of the SDG package is a less bad option?

markmc: A 32Mb file is not ideal, but in the context of some of these other packages ...

Paging @tiran @dhellmann @n1hility

src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml
markmc: Downstream users would need to change the contents of the package to change this recipe file anyway, so I don't think it's a problem to also change 1 line of code

If we really want to support downstream recipes, this is not how to do it - and we should file an issue for that requirement

I really don't want to leave stuff like this in place - someone upstream is quite likely to remove it, not understanding the downstream consequences. And if downstream no longer needs it, they're very unlikely to notify upstream of this.

That's how we end up with cruft that people are afraid to touch. Not a huge deal in this case, but broken windows and all that

dhellmann: Downstream we would be unlikely to put any default data file in the code package. Instead, we would install it separately as part of building the application image.

src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml
tiran: > download that just happens behind the scenes automatically

To a security engineer, this is a horror story in eight words. CC:  @sidhpurwala-huzaifa 

How is the download triggered? How is the content tracked, verified and signed? Please assume that $user wants to run InstructLab on an air-gapped system, with all data pre-loaded and without any access to the internet. Some users want to use InstructLab with PII or confidential data.

Very recently we had a similar behavior in a 3rd party dependency. It triggered a HTTPS request and a download every time we ran `ilab` command. The code ran on `import` time.

I would prefer a Python package or another mechanism to deliver and track the file.

dhellmann: Generally, libraries should not introduce side-effects like downloading files and writing to potentially protected locations on disk.

More specifically, when we get to the point of running tests for individual packages before using them downstream, those tests will have to run without accessing the internet. Everything they need will have to be installed before the tests run.

bbrowning: This entire discussion makes sense. Are there suggestions for how we do want to ship this precomputed dataset in upstream? How we want to reference it? And, how would downstream users supply their own precomputed dataset (replacing/obviating the upstream one specified here) in their own releases?

src/instructlab/sdg/datamixing.py
bbrowning: That's not a bad idea. I'm glad these discussions are happening, although it's starting to become clear this should have happened a couple of weeks ago when these features were first being developed elsewhere.

bbrowning: The `sys_prompt` only gets used when writing out the mixed dataset, where we overwrite the system prompt of every source sample to be the prompt specified here. And the reason we include it in the Recipe yaml file is so there's a record of what system prompt exists in the mixed dataset without having to go look inside the dataset file itself. So, maybe when loading a Recipe from disk or when writing it out we should always include the sys_prompt, and take the sys_prompt out of the constructor here. So the constructor would have less logic - loads a recipe from disk or doesn't, but none of this logic about which sys_prompt to use.

Then, within our own code, we'd change the sys_prompt of the recipe after loading it from disk, so that it has the system prompt we're using during generation.

Instead of:
```python
        self.knowledge_recipe = Recipe(
            recipe_path="src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml",
            sys_prompt=self.sys_prompt,
        )
```

we'd move to something like:
```python
        self.knowledge_recipe = Recipe(
            recipe_path="src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml",
        )
        self.knowledge_recipe.sys_prompt = self.sys_prompt
```

This would perhaps be cleaner and make more sense? It removes the overwriting of sys_prompt concerns out of the Recipe class and lets that just focus on loading/dumping itself. And it still lets us override the sys_prompt in our usage where we want to force the system prompt to be what we ran the generation pipeline with, regardless of what was or was not specified in some default recipe.

markmc: Isn't that the same as just not reading `sys_prompt` from the recipe file at all - e.g. by writing it out in a `metadata` section, and just not reading that section when loading?

markmc:
Ok, lots of great discussion in these threads. It's a lot to absorb, but I think it boils down to a pretty straightforward set of changes

**Default recipes - downstream**

If a downstream user/packager wants to add default recipes (and datasets), they should install them in `/usr/share/instructlab/sdg`

- `generate_data()` should supply `default_recipes_dir` path to `DataMixer` using `platformdirs.PlatformDirs`
- downstream users should install recipes into `/usr/share/instructlab/sdg/default_data_recipes/{knowledge,skills}.yaml`
- downstream users should install datasets into `/usr/share/instructlab/sdg/datasets`
- this a packaging thing - e.g. it would typically be done as part of the container image build
- the library would use any default recipes found there

**Default recipes - testing**

Since we won't have default recipes in the library, we do need some way of ensuring this code path is tested - this can be done with a unit test. But it is a requirement to merge this PR IMO.

**Community dataset recipe - upstream**

Assuming we have at least unit testing of recipes, I don't think this is a blocker to merging this PR - I'm fine with just filing an issue to capture this.

We want to enable upstream users to mix the `instructlab/InstructLabCommunity` dataset.

They would download the dataset and place it in `~/.cache` ? Potentially there would be an `ilab data download` command to do this

It can be supplied with e.g. `ilab data generate --skills-dataset=` and ilab would construct a simple skills recipe (in memory) and pass it to the library

This path could be tested in our e2e CI, which would also improve automated testing coverage of the library.

**Hugging Face downloads in library**

Given the above, there is no need for the library to have support for downloading from Hugging Face

**Relative paths**

There is no requirement to support relative paths to recipes files.

Paths to dataset should be considered relative to the recipe file.

**Empty recipes**

`Recipe()` should be sufficient to create an empty recipe.

**sys_prompt**

It sounds like we never want to use the `sys_prompt` value from a user-supplied recipe. It is included in recipe files that are written by the library for information only.

Move `sys_prompt` in the recipe format to a new `metadata` section, and ignore that whole section when reading.

bbrowning:
Thank you @markmc , that is super helpful. I've pushed a commit that should adjust to the `sys_prompt` guidance under a new `metadata` section, but it will take a bit more time/thought to implement some of the other changes. If you feel comfortable with the downstream dataset chunk of this or anything else, feel free to grab it. Otherwise I'll dig in to them in a bit.

bbrowning:
For reference, the generated recipe yaml files now look like this:
```yaml
datasets:
- path: instructlab/InstructLabCommunity
  sampling_size: 1.0
- path: node_datasets_2024-07-25T10_27_12/knowledge_tonsils_overview_e2e-tonsils_p10.jsonl
  sampling_size: 1.0
metadata:
  sys_prompt: "I am, Red Hat\xAE Instruct Model based on Granite 7B, an AI language\
    \ model developed by Red Hat and IBM Research, based on the Granite-7b-base language\
    \ model. My primary function is to be a chat assistant."
```

I've updated the original comment at the top of this PR to reflect the current state as well.

src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml
dhellmann: I'd really rather not assume we're going to manage patches downstream for packages where we have enough influence to ensure the design doesn't require it. We have enough work doing that for packages where we have no influence over the maintainers.

src/instructlab/sdg/configs/skills/data_recipe/default_recipe.yaml
dhellmann: These aren't the only inputs the system needs, right? Is there a pattern to follow in how things like models are made available for users?

markmc: See https://github.com/instructlab/sdg/pull/203#issuecomment-2250444499

src/instructlab/sdg/configs/knowledge/data_recipe/default_recipe.yaml
markmc: See https://github.com/instructlab/sdg/pull/203#issuecomment-2250444499

src/instructlab/sdg/datamixing.py
bbrowning: Resolving this, as the code has evolved significantly from here and addresses this concern.

src/instructlab/sdg/datamixing.py
bbrowning: Resolved by now writing the sys_prompt out to a metadata section.

src/instructlab/sdg/datamixing.py
bbrowning: Loading datasets from paths relative to the recipe file is now supported.

bbrowning:
Assuming CI passes, I believe all concerns are addressed now.

@markmc  added in support for loading of default recipe files from `platformdirs.PlatformDirs`, allowing downstream users to package default recipe yaml files in a place like `/usr/share/instructlab/sdg/default_data_recipes/{knowledge,skills}.yaml`. I added a unit test to ensure that if we find default recipe yaml files in those directories, they get loaded and used.

#201 was opened and expanded with more details as a result of the conversations in this PR to figure out how to properly get the precomputed dataset from HuggingFace into the upstream `ilab` experience in an easy-to-use way. This may involve a new command, such as `ilab data download`, but discussion around that design can happen in that issue.

This PR no longer includes downloading anything from HuggingFace, as several valid concerns were brought up about doing that silently and without explicit user intent.

Empty `Recipe()`s can be created, and we have unit tests ensuring that works along with a few other permutations.

We do not load sys_prompt from the recipe files, but do write it out to them under a new `metadata` section so that it's still obvious which system prompt was used during data mixing without having to look inside the data files themselves.

src/instructlab/sdg/datamixing.py
```diff
@@ -94,29 +89,67 @@ class Recipe:
...
+        respective to the directory the recipe yaml itself resides in.
+        """
+        if path.endswith(".jsonl"):
+            if not os.path.isabs(path):
+                path = os.path.join(os.path.dirname(self.recipe_path), path)
+            logger.info(f"Loading dataset from {path} ...")
+            dataset = load_dataset("json", data_files=path, split="train")
+        else:
+            logger.info(f"Loading dataset from HF {path} ...")
+            dataset = load_dataset(path, split="train")
```
markmc, line 125: I thought we were removing this?

markmc: Hugging Face download mentioned in docstring too

bbrowning: Yes, we are! However in my multi-tasking it looks like I accidentally did that in a different local branch than the one I pushed here. Need more coffee, will fix as I have a rebase conflict to fix anyway.

src/instructlab/sdg/datamixing.py
```diff
@@ -459,20 +512,31 @@ def collect(self, leaf_node_path, new_generated_data, is_knowledge):
...
         Mix the generated leaf node data into a single dataset and write it to
         disk. The heavy lifting is delegated to the Recipe class.
         """
         if recipe.dataset_added:
+            full_recipe_path = os.path.join(self.output_dir, output_file_recipe)
+            recipe.save_recipe(full_recipe_path)
+            # Mixed data is written out relative to a recipe's own
+            # path, so update the recipe's path to the newly written
+            # one before writing out its datasets
+            recipe.recipe_path = full_recipe_path
```
markmc, line 526: Maybe I'm being dumb, but I really don't understand what's going on here

We've just written a recipe file to a path, and setting `recipe.recipe_path` is going to influence where the dataset gets written? But `save_mixed_dataset()` doesn't use `recipe_path`?

Thanks!

src/instructlab/sdg/datamixing.py
bbrowning: Sorry, fixed that now.

markmc (approved):
Looks good to me, thanks @bbrowning 

Happy for you to merge this when CI passes

src/instructlab/sdg/datamixing.py
bbrowning: `save_mixed_dataset()` DOES use `recipe_path`, as it reads the datasets in from disk to mix them. I wasn't very happy with this when I wrote it, and I should have recognized the code smell myself when I wrote that comment explaining what it's doing. What happens is we loaded a default recipe from whatever path - say `/default_recipes/skills.yaml` (or instantiated an empty `Recipe()`, perhaps). Then, we save the recipe to a new path - like `/generated_data/skills_train_recipe.yaml`. At this point, the actual Recipe instance we have still thinks its `recipe_path` is `/default_recipes/skills.yaml`, even though we wrote it out to a different path. The fix staring me in the face as I type this is that when someone calls `save_recipe`, that should update the Recipe's `recipe_path` to the new path it just wrote itself out to.

The datasets are added to the recipe with a relative path, the recipe is written to a new path, and we then need to update the recipe to know its new path so that it can successfully load the datasets from the relative path, which is relative to the newly written location versus the original location.

bbrowning: I pushed one more commit that adjusts this that feels better for me, and keeps this detail encapsulated within the Recipe class instead of leaking out to consumers.

bbrowning:
Thanks! Will squash commits, unmark from Draft, and merge when CI is good.

Issue #202, opened 2024-07-24
Input dataset for `generate_eval_task_data` should be taxonomy samples, not generated data


bbrowning:
https://github.com/instructlab/sdg/blob/d5b9c31f4b19ad2029d449dc12507e3c667bf68b/src/instructlab/sdg/generate_data.py#L402

We should be passing the original samples (in `ds`) here to `generate_eval_task_data` instead of the generated samples. See https://github.com/aakankshaduggal/sdg/blob/1dfc0ce2fddfdabade6fd9525c6df33fd455375a/src/instructlab/sdg/generate_data.py#L258 - the original samples from `ds` are the ultimate input to the eval pipeline.

This causes a bug as I try to integrate auxiliary datasets that isn't triggered by the current code, but what happens is we end up getting two different places trying to add a `dataset_type` to the same samples because we're accidentally running the eval pipeline on top of our generated samples instead of on the original samples.

markmc:
See https://github.com/instructlab/sdg/pull/204#discussion_r1690607807

PR #204, opened 2024-07-25
Add support for auxiliary dataset generation

bbrowning:
This adds support for generating auxiliary datasets during knowledge data generation. An auxiliary dataset is where we ask the model to generate some additional data samples with a different prompt than the standard dataset, along with some extra instruction prompts that will get matched to the auxiliary generated samples and used during training.

Parts of this are extracted and rebased from
https://github.com/aakankshaduggal/sdg/pull/4
https://github.com/aakankshaduggal/sdg/pull/21

Refs #162.


src/instructlab/sdg/generate_data.py
```diff
@@ -381,10 +381,12 @@ def generate_data(
 
         if is_knowledge:
             # generate mmlubench data for the current leaf node
+            # note that this needs the original samples dataset as
+            # input, and NOT the newly generated samples dataset
             generate_eval_task_data(
                 mmlu_bench_pipe,
                 leaf_node_path,
-                new_generated_data,
+                ds,
```
bbrowning, line 389: The auxiliary dataset addition tripped the bug in #202 which caused the generate pipeline to crash, so this fixes that as part of this PR. However, if that issue gets fixed in isolation separately, I'll happily rebase. This gives a chance of CI working on this for now however.

bbrowning:
Checkpointing my work here to let CI chew on this and see if it works on more than just my machine. Any comments and suggestions are welcome, although keeping this marked as draft for now as I'd like to do another pass on this work myself to clean and document things a bit.

bbrowning:
Hmm, it looks like we're missing the expanded schema definitions necessary for the new pipeline blocks added in #182 . Created #205 to track this, as it's a bit orthogonal to this PR's work.

bbrowning:
Added a commit to this PR that at least partially fixes #205 for now. It feels wrong adding that to this PR, and it should likely get pulled out into its own PR perhaps with accompanying test that hands a basic yaml block definition to the schema validation for each block type just to ensure baseline expected block configuration is covered by the schemas. I know we have `tox -e validate-pipelines`, but that only validates the committed pipeline definitions we ship by default as opposed to validating the universe of possible and/or expected possible upstream or downstream pipeline definitions.

markmc:
> Added a commit to this PR that at least partially fixes #205 for now. It feels wrong adding that to this PR, and it should likely get pulled out into its own PR

done in #206 

> perhaps with accompanying test that hands a basic yaml block definition to the schema validation for each block type just to ensure baseline expected block configuration is covered by the schemas. I know we have `tox -e validate-pipelines`, but that only validates the committed pipeline definitions we ship by default as opposed to validating the universe of possible and/or expected possible upstream or downstream pipeline definitions.

Filed #207

Thanks!

src/instructlab/sdg/datamixing.py
```diff
@@ -325,6 +327,76 @@ def _conv_pretrain(rec):
     return rec
 
 
+def _load_auxiliary_instructions():
+    """
+    Load the auxiliary instructions yaml from disk, returning the loaded
+    object if found or None if the instructions yaml file does not exist.
+    """
+    auxilary_path = resources.files(__package__).joinpath(
+        "configs/knowledge/auxiliary_instructions.yaml"
```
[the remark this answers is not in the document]
bbrowning: This is mirroring where these files are already placed by some of our downstream users, so it could be moved but we'll need to coordinate with them to ensure downstream auxiliary instructions gets moved as well.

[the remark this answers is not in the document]
markmc: Yes please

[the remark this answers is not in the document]
bbrowning: Naming is hard for me, but I took a first stab at this by moving this from `configs/` to `instructions/auxiliary_knowledge.yaml`.

src/instructlab/sdg/generate_data.py
bbrowning: I removed this change from my PR since the fixed landed upstream.

markmc:
Comment from me on slack:

> ITSM there is some sort of connection between the pipeline config and the instructions ... a pipeline author would write those two things together ... some I'd be in favor of the pipeline config providing the instructions location somehow

danmcp:
@bbrowning What's left on your list before this can be moved out of draft state?

markmc:
> @bbrowning What's left on your list before this can be moved out of draft state?

One remaining issue is this:

```
    auxiliary_path = resources.files(__package__).joinpath(
        "instructions/auxiliary_knowledge.yaml"
    )
```

i.e. the only location the code can load instructions from is this path in the Python package

If downstream custom pipeline configs need different instructions, then we need to be able to load it from the same place the custom pipeline configs are installed. This is how we find pipeline configs:

```
    pd = platformdirs.PlatformDirs(
        appname=os.path.join("instructlab", "sdg"), multipath=True
    )
    for d in pd.iter_data_dirs():
        pipeline_path = os.path.join(d, "pipelines", pipeline)
        if os.path.exists(pipeline_path):
```

We also load `default_recipes` from this system director:

```
        for d in self.data_dirs:
            default_recipe_path = os.path.join(d, "default_data_recipes", yaml_basename)
            if os.path.exists(default_recipe_path):
```

so the most obvious thing would be to load from e.g. `/usr/share/instructlab/sdg/auxiliary_instructions/knowledge.yaml`

I still feel this is suboptimal, and we'll probably want to change this in future so the instructions location is a relative path specified in the pipeline config

markmc:
> so the most obvious thing would be to load from e.g. `/usr/share/instructlab/sdg/auxiliary_instructions/knowledge.yaml`

I started implementing this, but in the case where downstream custom pipelines are installed, then we need to know to use the instructions in the Python package with the `full` pipeline and then use the appropriate instructions from `/usr/share` with the downstream custom pipeline. And it's plausible we could have multiple downstream pipelines, each with different instructions.

We definitely need a way to link the pipeline config with the instructions.

markmc:
> We definitely need a way to link the pipeline config with the instructions.

https://github.com/instructlab/sdg/pull/204/commits/824e163e3b613c7e42beefd319b442ff472a76d8 adds the auxiliary instructions to the pipeline config

src/instructlab/sdg/datamixing.py
```diff
@@ -374,7 +374,70 @@ def _conv_pretrain(rec):
...
+    # If we didn't find any auxiliary instructions to load, then
+    # that's also another sign that we're not running with any
+    # auxiliary datasets enabled.
+    if auxiliary_inst is None:
+        return None
+    # This "base_document" dataset_type is set in the knowledge
+    # pipeline config, and represents samples that do not have the
+    # auxiliary generated document attached, so we filter those out.
+    auxiliary_ds = generated_dataset.filter(
+        lambda x: x["dataset_type"] != "base_document"
```
markmc, line 395: This kind of thing makes it difficult to author pipelines - you have to know to create this column, but we could do it automatically if e.g. `data.mixing.auxiliary_instructions` was set in the pipeline config

Probably not a change to make at this stage though

src/instructlab/sdg/datamixing.py
```diff
@@ -374,7 +374,70 @@ def _conv_pretrain(rec):
...
+    )
+    unique_document_auxiliary = (
+        auxiliary_ds.to_pandas()
+        .drop_duplicates(subset=["document"])
+        .reset_index(drop=True)
+    )
+    unique_document_auxiliary = Dataset.from_pandas(unique_document_auxiliary)
```
markmc, line 402: Use `instructlab.sdg.utils.pandas.dataset_from_pandas_dataframe()`

src/instructlab/sdg/pipelines/full/knowledge.yaml
```diff
@@ -1,5 +1,36 @@
...
+        - base_document
+      value_name: corrected_document
+      var_name: dataset_type
+
+  - name: rename_to_document_column
+    type: RenameColumnsBlock
+    config:
+      columns_map:
+        document: raw_document
+        corrected_document: document
```
markmc, line 32: This makes me nervous - `columns_map` is a dict ... what's guaranteeing the ordering here?

markmc: In the datasets library:

```
        def rename(columns):
            return [column_mapping[col] if col in column_mapping else col for col in columns]
```

i.e. it's applying these in the order the columns in the dataset

Ok, at least that's deterministic

markmc:
Fixed this e2e failure:

```
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/instructlab/data/generate.py", line 236, in generate
    generate_data(
  File "/home/runner/work/sdg/sdg/venv/lib/python3.11/site-packages/instructlab/sdg/generate_data.py", line 371, in generate_data
    mixer = _mixer_init(ctx, output_dir, date_suffix, sdg_knowledge.auxiliary_inst)
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SDG' object has no attribute 'auxiliary_inst'
```

src/instructlab/sdg/datamixing.py
bbrowning: Yes, updated to do that now that that helper method merged. Thanks!

src/instructlab/sdg/pipelines/full/knowledge.yaml
bbrowning: Agreed, and good catch - that seems quite fragile, and this config should probably be refactored to be an array? Or this split out into two steps, to take out any doubt of the ordering the renaming is applied.

bbrowning:
Rebased, unmarked as draft to keep this moving along.

src/instructlab/sdg/datamixing.py
bbrowning: Created as #228 to track this, as you're right that this column value has special meaning in the code and that's non-obvious to pipeline authors nor even clear they need to explicitly add it into their pipeline config versus us doing it implicitly at the right point.

src/instructlab/sdg/pipelines/full/knowledge.yaml
bbrowning: I think this could be done after merging this larger PR, since the order happens to be deterministic. Especially because it may warrant a second look at the API exposed by `RenameColumnsBlock`.

markmc (approved):

Issue #205, opened 2024-07-25
Newer pipeline blocks missing from pipelines schema


bbrowning:
While working on #204, I hit a CI failure trying to utilize the new pipeline blocks added in #182 with a schema validation error.

An example of a block that fails is:
```yaml
  - name: rename_to_document_column
    type: RenameColumnsBlock
    config:
      columns_map:
        document: raw_document
        corrected_document: document
```

The full error is visible at https://github.com/instructlab/sdg/actions/runs/10086008816/job/27887843469?pr=204 , but the important part is:
```
=======================================================
=== Validating src/instructlab/sdg/pipelines/full/knowledge.yaml
=======================================================
Validation failed for src/instructlab/sdg/pipelines/full/knowledge.yaml: {'columns_map': {'document': 'raw_document', 'corrected_document': 'document'}} is not valid under any of the given schemas

Failed validating 'anyOf' in schema['properties']['blocks']['items']['properties']['config']:
    {'anyOf': [{'type': 'object',
                'description': 'ImportBlock',
                'required': ['path'],
                'additionalProperties': False,
                'properties': {'path': {'type': 'string'}}},
               {'type': 'object',
                'description': 'FilterByValueBlock',
                'required': ['filter_column', 'filter_value', 'operation'],
                'additionalProperties': False,
                'properties': {'convert_dtype': {'type': 'string',
                                                 'enum': ['float',
                                                          'int',
                                                          'bool']},
                               'default_value': {'type': 'string'},
                               'filter_column': {'type': 'string'},
                               'filter_value': {'oneOf': [{'type': 'string'},
                                                          {'type': 'number'},
                                                          {'type': 'array',
                                                           'items': {'oneOf': [{'type': 'string'},
                                                                               {'type': 'number'}]}}]},
                               'operation': {'type': 'string',
                                             'enum': ['eq',
                                                      'ne',
                                                      'gt',
                                                      'ge',
                                                      'lt',
                                                      'le',
                                                      'contains']}}},
               {'type': 'object',
                'description': 'LLMBlock',
                'required': ['config_path', 'output_cols'],
                'additionalProperties': False,
                'properties': {'config_path': {'type': 'string'},
                               'output_cols': {'type': 'array',
                                               'items': {'type': 'string'}},
                               'model_prompt': {'type': 'string'},
                               'parser_kwargs': {'type': 'object',
                                                 'properties': {'parser_name': {'type': 'string'},
                                                                'parsing_pattern': {'type': 'string'},
[...]

PR #206, opened 2024-07-25
Update the v1 pipeline schema to include new pipeline blocks

markmc:
Update the v1 pipeline schema to include the blocks added by #182

Refs #205

Add a new schema entry for eachof these blocks.

Note that the entry for DuplicateColumnsBlock and RenameColumnsBlock is identical - so could have been combined - 
 but it will be easier to audit the correctness of the schema vs the code	if we have both entries.

### DIFF
--- a/src/instructlab/sdg/pipelines/schema/v1.json
+++ b/src/instructlab/sdg/pipelines/schema/v1.json
@@ -303,6 +303,59 @@
                     }
                   }
                 }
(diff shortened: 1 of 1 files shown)


derekhiggins (approved):
lgtm, also passes validation with a pipeline containing these Blocks and 
`python scripts/validate_pipelines.py`

PR #214, opened 2024-07-25
Move system pipelines to /usr/share/instructlab/sdg/pipelines

markmc:
Currently it is /usr/share/instructlab/sdg and we may want other SDG stuff in this directory.

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -207,9 +207,9 @@ def _sdg_init(ctx, pipeline):
         appname=os.path.join("instructlab", "sdg"), multipath=True
     )
     for d in pd.iter_data_dirs():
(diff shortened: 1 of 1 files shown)


src/instructlab/sdg/generate_data.py
```diff
@@ -206,7 +206,8 @@ def _sdg_init(ctx, pipeline):
     pd = platformdirs.PlatformDirs(
         appname=os.path.join("instructlab", "sdg"), multipath=True
     )
-    for d in pd.iter_data_dirs():
+    pipelines_dir = os.path.join(pd, "pipelines")
```
derekhiggins, line 209: Did you want to do this inside the loop ?  (with d)

markmc: Yep, thanks - dumb mistake :+1:

derekhiggins (approved):
lgtm

PR #215, opened 2024-07-25
Incorporate knowledge generation context selection improvements

bbrowning:
This incorporates fixes from
https://github.com/aakankshaduggal/sdg/pull/21 to the way we select the expanded contexts during the knowledge data generation, cleaning up the logic around how we select which other documents to include in the expanded context during edge cases where we have low numbers of unique documents


markmc (approved):
Not going to claim to have reviewed this carefully

bbrowning:
I reviewed it the overall logic here in selector distractor/golden documents (as described in the RAFT paper linked in the code comment in this change) in these edges cases with the team in-person yesterday, and believe it to be correct. And, the generated results look better now which should lead to better training results when we have lower numbers of unique documents to pick from. Thanks for the quick work!

Issue #195, opened 2024-07-24
[Epic] Add data checkpointing and recovery


markmc:
In order to ensure that progress is not completely lost if there is an error during a long-run data generation process, add a checkpointing system that saves intermediate states during data generation. This allows for more granular progress tracking and recovery.

See https://github.com/aakankshaduggal/sdg/pull/19

* [x] Add capability to library -- https://github.com/instructlab/sdg/pull/222
* [x] Enable checkpointing in the CLI -- https://github.com/instructlab/instructlab/pull/1915
* [x] Enable batching in CLI (checkpointing is coupled with batching) https://github.com/instructlab/instructlab/pull/1797
* [x] File an issue to consider making use of `save_freq` in future -- https://github.com/instructlab/sdg/issues/225

PR #222, opened 2024-07-26
Add data checkpointing capability

derekhiggins:
Fixes https://github.com/instructlab/sdg/issues/195

Needs the following CLI changes to enable it:

- `--batch-size` -- https://github.com/instructlab/instructlab/pull/1889
- `--checkpoint-dir` -- https://github.com/instructlab/instructlab/pull/1797

---

c1c70a5 Add data checkpointing capability

```    
    Introduce a comprehensive data checkpointing mechanism that saves
    intermediate states during data generation. This allows for more
    granular progress tracking and recovery.
    
    Checkpoints are saved periodically based on the save_freq setting,
    preventing data loss and enabling resumption from the last saved state
    in case of interruptions.
    
    The system can resume from a saved checkpoint by comparing the
    generated data with the seed data to identify and process missing
    data.
    
    Each checkpoint is uniquely identified using UUIDs, ensuring distinct
    and traceable save points.
    
    Co-authored-by: Derek Higgins <derekh@redhat.com>
    Co-authored-by: Mark McLoughlin <markmc@redhat.com>
```

63b3599 checkpointing: fix "missing data" check with removed columns
    
```
    Fix the logic error that means if a pipeline removes a column
    that was present in the original dataset, then checkpointing
    causes the column not be present before the pipeline starts.
    
    Add a test case to the unit test to cover this, but note I
    had to add an additional column to ensure there is at least
    one column in common between the original dataset and the
    checkpoint dataset.
```

### DIFF
--- a/src/instructlab/sdg/checkpointing.py
+++ b/src/instructlab/sdg/checkpointing.py
@@ -80,8 +80,7 @@ class Checkpointer:
         generated_df = generated_data_common.to_pandas()
 
         # Identify missing rows
-        missing_df = seed_df[
-            ~seed_df.apply(tuple, 1).isin(generated_df.apply(tuple, 1))
-        ]
+        missing_rows = ~seed_df.apply(tuple, 1).isin(generated_df.apply(tuple, 1))
 
+        missing_df = seed_data.to_pandas()[missing_rows]
         return pandas.dataset_from_pandas_dataframe(missing_df)
--- a/src/instructlab/sdg/checkpointing.py
+++ b/src/instructlab/sdg/checkpointing.py
@@ -1,0 +1,87 @@
+# Standard
+import logging
+import uuid
+
+# Third Party
+from datasets import Dataset, concatenate_datasets, load_dataset
+from datasets.data_files import EmptyDatasetError
+
+# First Party
+from instructlab.sdg.utils import pandas
+
+logger = logging.getLogger(__name__)
+
+
+class Checkpointer:
+    def __init__(self, checkpoint_dir=None, save_freq=1):
+        self._checkpoint_dir = checkpoint_dir
+
+        self._save_freq = save_freq
+        self._cache = []
+
+    def checkpoint(self, dataset):
+        self._cache.append(dataset)
+        if len(self._cache) < self._save_freq:
+            return
+        self.save()
+        self._cache.clear()
+
+    def done(self):
+        if self._cache:
+            self.save()
+            self._cache.clear()
+
+    def save(self):
+        if self._checkpoint_dir is None:
+            return
+        checkpoint_id = uuid.uuid4().hex
+        checkpoint_file = (
+            f"{self._checkpoint_dir}/data_checkpoint_{checkpoint_id}.jsonl"
+        )
+        logger.info(f"Saving checkpoint to {checkpoint_file}")
+        # Saves all the current records to new file in the checkpoint dir
+        concatenate_datasets(self._cache).to_json(
+            checkpoint_file, orient="records", lines=True
+        )
+
+    def load(self, dataset: Dataset) -> Dataset:
+        if self._checkpoint_dir is None:
+            return dataset, None
+
+        try:
+            pre_generated_data = load_dataset(
+                "json", data_dir=self._checkpoint_dir, split="train"
+            )
+        except EmptyDatasetError:
+            logger.info(
+                f"No existing checkpoints found in {self._checkpoint_dir}, generating from scratch"
+            )
+            return dataset, None
+
+        logger.info(
+            f"Loading existing checkpoints from {self._checkpoint_dir}, with {pre_generated_data.num_rows} rows"
+        )
+        seed_data = self._get_missing_data(dataset, pre_generated_data)
+        logger.info(f"Found {seed_data.num_rows} missing rows in the dataset")
+        return seed_data, pre_generated_data
+
+    def _get_missing_data(self, seed_data, generated_data):
+        # Get the common columns between the two datasets
+        common_columns = list(
+            set(seed_data.column_names) & set(generated_data.column_names)
+        )
+
+        # Extract the relevant data based on common columns
+        seed_data_common = seed_data.select_columns(common_columns)
+        generated_data_common = generated_data.select_columns(common_columns)
+
+        # Convert to Pandas DataFrames for easier comparison
+        seed_df = seed_data_common.to_pandas()
+        generated_df = generated_data_common.to_pandas()
+
+        # Identify missing rows
+        missing_df = seed_df[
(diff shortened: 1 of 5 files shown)


src/instructlab/sdg/pipeline.py
```diff
@@ -193,6 +204,7 @@ def _generate_single(self, dataset) -> Dataset:
             if drop_duplicates_cols:
                 dataset = self._drop_duplicates(dataset, cols=drop_duplicates_cols)
 
+        self.checkpointer.save(dataset)
```
derekhiggins, line 207: will be moving this out of the tread

src/instructlab/sdg/checkpointing.py
```diff
@@ -0,0 +1,93 @@
...
+        # Extract the relevant data based on common columns
+        seed_data_common = seed_data.select_columns(common_columns)
+        generated_data_common = generated_data.select_columns(common_columns)
+
+        # Convert to Pandas DataFrames for easier comparison
+        seed_df = seed_data_common.to_pandas()
+        generated_df = generated_data_common.to_pandas()
+
+        # Identify missing rows
+        missing_df = seed_df[
```
markmc, line 74: Ok, this is identifying missing rows - but it looks like it will only return the common columns for those missing rows. What if the pipeline had deleted one of the original columns?

Should we instead be selecting from `seed_data`

the whole thing is quite confusing. Here's my attempt to make it more clear

```
def to_tuples(ds, columns):
    """Convert the dataset into a dataframe of tuples - one tuple per row - but only including the specified columns"""
   df = ds.select_columns(columns).to_pandas()
   return df.apply(tuple, axis=1)

# Get the common columns between the two datasets
common_columns = list(
    set(seed_data.column_names) & set(generated_data.column_names)
)

seed_tuples = to_tuples(seed_data, common_columns)
generated_tuples = to_tuples(generated_data, common_columns)

missing_rows = ~seed_tuples.isin(generated_tuples)

missing_df = seed_data.to_pandas()[missing_rows]

return utils.pandas.dataset_from_pandas_dataframe(missing_df)
```

src/instructlab/sdg/checkpointing.py
```diff
@@ -0,0 +1,104 @@
...
+                pre_generated_data = load_dataset(
+                    "json", data_dir=self._checkpoint_dir, split="train"
+                )
+                logger.info(
+                    f"Loading existing checkpoints from {self._checkpoint_dir}, with {pre_generated_data.num_rows} rows"
+                )
+                seed_data = self._get_missing_data(dataset, pre_generated_data)
+                logger.info(f"Found {seed_data.num_rows} missing rows in the dataset")
+
+            except EmptyDatasetError:
```
markmc, line 58: What raises this? Is it only `load_dataset()`? If so, let's move everything out of this try/except block

src/instructlab/sdg/generate_data.py
```diff
@@ -183,12 +183,17 @@ def _context_init(
...
+    checkpoint_dir: Optional[str] = None,
+    save_freq: Optional[int] = 1,
 ):
     extra_kwargs = {}
     if batch_size is not None:
         extra_kwargs["batch_size"] = batch_size
         extra_kwargs["batch_num_workers"] = batch_num_workers
 
+    extra_kwargs["checkpoint_dir"] = checkpoint_dir
+    extra_kwargs["save_freq"] = save_freq
```
markmc, line 195: You shouldn't need to use extra_kwargs for this ... it's only for args that are set conditionally based on batch_size?

src/instructlab/sdg/generate_data.py
```diff
@@ -284,6 +289,7 @@ def generate_data(
     client: Optional[openai.OpenAI] = None,
     pipeline: Optional[str] = "simple",
     batch_size: Optional[int] = None,
+    checkpoint_dir: Optional[str] = None,
```
markmc, line 292: Are we expecting this to be a CLI arg?

src/instructlab/sdg/generate_data.py
```diff
@@ -400,6 +410,8 @@ def generate_data(
 
         if is_knowledge:
             # generate mmlubench data for the current leaf node
+            # Make sure checkpointing is disabled (we don't want this pipeline to load checkpoints from the main pipeline)
+            ctx.checkpoint_dir = None
```
markmc, line 414: Nasty :)

How about, when creating `mmlu_pipe`:

```
# Make sure checkpointing is disabled (we don't want this pipeline to load checkpoints from the main pipeline)
mmlu_ctx = dataclasses.replace((ctx, checkpoint_dir=None)
mmlu_bench_pipe = mmlubench_pipe_init(mmlu_ctx)
```

src/instructlab/sdg/generate_data.py
derekhiggins: I thought it would be, a user would need to be aware checkpoints are being used and control if they want it or not  (checkpoint_dir=None)
https://github.com/instructlab/instructlab/pull/1889

src/instructlab/sdg/generate_data.py
derekhiggins: will do 

> Nasty :)
Yup, I knew when I was doing it

src/instructlab/sdg/checkpointing.py
derekhiggins: I beleive so yes, will do

src/instructlab/sdg/checkpointing.py
derekhiggins: > Ok, this is identifying missing rows - but it looks like it will only return the common columns for those missing rows. What if the pipeline had deleted one of the original columns?

In that case we would have less columns in common,  which I guess would result in samples being marked as checkpointed in error.

> Should we instead be selecting from seed_data

yes, this would make sure a column that was deleted by a pipe would be in the returned dataset  but I think wouldn't help if samples were marked as checkpointed in error because of that missing column (I think that makes sense)


There is also a problem with rows that were filtered out by the pipeline, these never get checkpointed so will all be retried on subsequent runs

markmc: Take a look at https://github.com/instructlab/sdg/pull/222/commits/ca11adf965879740172a2a9142c4553627f2dc14

I think that's a genuine issue?

derekhiggins: Yup, agreed, reproduced it locally with a rename block and your patch fixed it

derekhiggins:
> instructlab.sdg.pipeline.PipelineBlockError: PipelineBlockError(<class 'instructlab.sdg.llmblock.LLMBlock'>/gen_skill_freeform): Service Unavailable

this in CI is I think because we have concurrency turned on against llama_cpp

derekhiggins (commented):
lgtm

markmc (approved):

PR #226, opened 2024-07-29
tests: Add test to validate that generate_data() is generating the files expected

### LINKED ISSUE #208: Add test to validate that generate_data() is generating the files expected
We have various users (e.g. the eval and train libraries) that expect us to spit out certain things, but nothing validating that we aren't breaking that contract

Something like:

1. patch `Pipeline.from_file()` to return a mock pipeline whose `generate()` does the bare minimum required to save the various datasets
2. Also mock out the mmlu eval benchmark `generate()`
3. Call `generate_data()` with `pipeline=testpipeline`, `output_dir=tmppath`, taxonomy with 1 each of most basic grounded+foundational skills, and knowledge
4. Verify all of the files we expect are there, with some basic validation of contents (e.g. expected columns)

This isn't about validating Pipeline, or individual blocks, it's more about the output file generation side of things

hickeyma:
Provides tests to validate that `generate_data()` API is generating the files expected. It:

1. Uses basic `Pipeline` "simple"
2. Mocks out the `LLMBlock` `_generate()` to return generated output from Model (data captured when generate process run fully)
3. Tests with taxonomies of compositional skills, and knowledge
6. Verifies all of the files expected are there, with some basic validation of contents

Closes #208


markmc:
@hickeyma I went down the pretty miserable rabbit hole of validating the contents of the output files, since my sense is that's where a lot of the brittleness with this code is going to be. What do you think?

tests/test_generate_data.py
```diff
@@ -3,12 +3,653 @@
...
+  Overview of Human tonsils, describing their types, locations, structure,
+  function, and clinical significance, with a specific focus on their role in
+  the immune system and related health issues.
+
+document:
+  repo: https://github.com/luke-inglis/il-anatomy-knowledge
+  commit: cc7c6ca
+  patterns:
+    - anatomy1.md
+"""
```
khaledsulayman, line 222: can I suggest we add these as files in a `mock_data` directory, perhaps?

hickeyma:
>  went down the pretty miserable rabbit hole of validating the contents of the output files, since my sense is that's where a lot of the brittleness with this code is going to be. What do you think?

Thanks @markmc for pushing the changes. I was between 2 minds about trying to keep the pipeline as true as possible, thereby avoiding as much mocking as possible. Faking v Making argument! Lets go with the mocking of the `_generate()`.
The validation of the contents are a rabbit hole alright. They tests are a great addition, thanks.

tests/test_generate_data.py
hickeyma: Good point, thanks @khaledsulayman

hickeyma: Updated to now load skills from test data

tests/test_generate_data.py
```diff
@@ -3,12 +3,464 @@
...
+    """Generate expected training samples in the user/assistant format
+    used by the legacy training methods such as Linux training and MacOS
+    training.
+
+    Mirroring _noop_llmblock_generate() below, we generate 10 samples
+    per input, and then follow _gen_train_data()'s output format.
+    """
+
+    def add_q(q):
+        return (q + "?") if not "?" in q else q
```
khaledsulayman, line 199: nit:
```suggestion
    def add_qestion_mark(q):
        if not q.endswith("?"):
            return q + "?"
        return q
```

tests/test_generate_data.py
```diff
@@ -3,12 +3,464 @@
...
+            else:
+                # FIXME: handle freeform skills - no context
+                train_samples.append(
+                    {
+                        "user": seed_example["context"]
+                        + f" (q{i}) "
+                        + add_q(seed_example["question"]),
+                        "assistant": f"(a{i}) " + seed_example["answer"],
+                    }
+                )
```
khaledsulayman, line 226: at this level of indentation I feel it is cleaner to export this logic to a separate function that can be just called in the loop for each seed example. Maybe something like `process_seed_example(seed_example)` or maybe something more specific if possible.

Also, is there a specific reason why we only want the first 10 Q&As? Apologies if I'm missing something.

tests/test_generate_data.py
```diff
@@ -3,12 +3,464 @@
...
+    simple pipeline is configured with 'n: scaled' and we pass
+    num_instructions_to_generate=10 to generate_data.
+    """
+
+    def strip_q(q):
+        return q.strip().rstrip("?")
+
+    output = []
+    for sample in samples:
+        for i in range(10):
```
khaledsulayman, line 254: same question about the 10. Maybe this is best stored as a global variable that each of these loops can use so as to avoid duplication.

tests/test_generate_data.py
```diff
@@ -3,12 +3,464 @@
...
+                "mmlubench_knowledge_new.jsonl",
+            ]:
+                matches = glob.glob(
+                    os.path.join(self.tmp_path, "node_datasets_*", name)
+                )
+                assert len(matches) == 1
+                if name == "knowledge_new_task.yaml":
+                    validate_lm_eval_task(matches[0])
+                elif name == "mmlubench_knowledge_new.jsonl":
+                    validate_mmlubench_dataset(matches[0])
```
khaledsulayman, line 449: I don't believe the rest of this code needs to sit under the `with` statement, just the call to `generate data()`

tests/test_generate_data.py
khaledsulayman: just want to add I assumed you were trying to specifically match strings that don't end with a question mark. If you instead mean to search for a question mark anywhere within the string, you can disregard this comment. Although I still think it's slightly more readable as a multiline `if`.

khaledsulayman:
left a few comments, mostly stylistic. Otherwise, everything looks good!

hickeyma:
Thanks @khaledsulayman  for the feedback. I'll update your comments shortly.

tests/test_generate_data.py
hickeyma: Its a valid point that you make. I think however the  use of the ternary operator in this case is simple enough to be understandable.  With that said, I have updated the inner function name from `add_q` to `add_question_mark` for better clarity.

tests/test_generate_data.py
hickeyma: > at this level of indentation I feel it is cleaner to export this logic to a separate function that can be just called in the loop for each seed example. Maybe something like process_seed_example(seed_example) or maybe something more specific if possible.

I am on the fence about this. Its on the border of being over indented and as the code block is not used anywhere else, I am going to leave it as is for now.

hickeyma: > Also, is there a specific reason why we only want the first 10 Q&As? Apologies if I'm missing something.

This is generating 10 examples for training samples, for test purposes. 10 is probably enough for the overall testing of the pipeline.

tests/test_generate_data.py
hickeyma: Good point. I'll add a constant to be used.

In this case, we generate 10 output samples per input samples, since the `LLMBlock` in the simple pipeline is configured with 'n: scaled' and we pass `num_instructions_to_generate=10` to `generate_data()`.

tests/test_generate_data.py
hickeyma: Valid point, shifted block left

khaledsulayman (approved):
Thanks for addressing those comments @hickeyma

LGTM (Let's Get This Merged 😁)

hickeyma:
@markmc This should be ready for merging. What do you think?

markmc (approved):
Looks great, thank you @hickeyma !

PR #231, opened 2024-07-29
Separate checkpoints by leaf nodes

danmcp:
The checkpoints need to be saved and loaded per leaf node.  This logic adds leaf node into the checkpoint save path.  Ex:

ls ~/.local/share/instructlab/datasets/checkpoints/knowledge_tonsils_overview_e2e-tonsils/ data_checkpoint_3c0182d541c042f6bacb4e157d1eed5c.jsonl  data_checkpoint_3dd6fc2d53544353b76070a0fe00e277.jsonl  data_checkpoint_7555281c378744c79013541e42f18e42.jsonl

### DIFF
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -131,15 +131,21 @@ class Pipeline:
             pipeline_yaml = os.path.join(resources.files(__package__), pipeline_yaml)
         return cls(ctx, pipeline_yaml, *_parse_pipeline_config_file(pipeline_yaml))
 
-    def generate(self, dataset) -> Dataset:
+    def generate(self, dataset, checkpoint_name=None) -> Dataset:
         """
         Generate the dataset by running the pipeline steps.
         dataset: the input dataset
+        checkpoint_name: unique subdir name for the checkpoint within checkpoint_dir
         """
 
         # The checkpointer allows us to resume from where we left off
         # Saving the output of pipe instances along the way
-        checkpointer = Checkpointer(self.ctx.checkpoint_dir, self.ctx.save_freq)
+        checkpoint_dir = None
+        if self.ctx.checkpoint_dir is not None and checkpoint_name is not None:
+            # Separate checkpoints with sub directories
+            checkpoint_dir = os.path.join(self.ctx.checkpoint_dir, checkpoint_name)
+
+        checkpointer = Checkpointer(checkpoint_dir, self.ctx.save_freq)
         dataset, pre_generated_data = checkpointer.load(dataset)
 
         # If not batching, simply delegate to _generate_single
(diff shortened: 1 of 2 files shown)


n1hility (approved):

src/instructlab/sdg/pipeline.py
```diff
@@ -131,15 +131,21 @@ def from_file(cls, ctx, pipeline_yaml):
             pipeline_yaml = os.path.join(resources.files(__package__), pipeline_yaml)
         return cls(ctx, pipeline_yaml, *_parse_pipeline_config_file(pipeline_yaml))
 
-    def generate(self, dataset) -> Dataset:
+    def generate(self, dataset, leaf_node_path=None) -> Dataset:
```
shivchander, line 134: I think it would make sense to pass a `checkpoint_dir` arg to a generate method. This makes it more generic and aligned to other ML libraries. 

Also in general would be good to not use `leaf_node_path` or any other var names which are tied to the way we have our generate_data setup.

shivchander (changes requested):

src/instructlab/sdg/pipeline.py
danmcp: > I think it would make sense to pass a `checkpoint_dir` arg to a generate method. This makes it more generic and aligned to other ML libraries.

checkpoint_dir is already being passed via the ctx.  So I figured it would make since to only pass the additional info.  If we are going to pass the entire checkpoint_dir, we'll probably want to remove the root from the ctx.

> Also in general would be good to not use leaf_node_path or any other var names which are tied to the way we have our generate_data setup.

Will rename

src/instructlab/sdg/pipeline.py
```diff
@@ -131,15 +131,21 @@ def from_file(cls, ctx, pipeline_yaml):
             pipeline_yaml = os.path.join(resources.files(__package__), pipeline_yaml)
         return cls(ctx, pipeline_yaml, *_parse_pipeline_config_file(pipeline_yaml))
 
-    def generate(self, dataset) -> Dataset:
+    def generate(self, dataset, checkpoint_name=None) -> Dataset:
         """
         Generate the dataset by running the pipeline steps.
         dataset: the input dataset
+        checkpoint_name: unique subdir name for the checkpoint within checkpoint_dir
```
markmc, line 138: I might have called this `run_id` or something - a unique name for this pipeline run - the only current use being for checkpoint directories

markmc:
Thanks @danmcp - merging this quickly since it seems an important fix, and my or @shivchander's comments are easily addressed with a bit of refactoring

src/instructlab/sdg/pipeline.py
shivchander: `checkpoint_dir` is not used elsewhere other than `.generate` method. So does it make sense to have it tied up with the `PipelineContext`? 

I would suggest something like this: 

```python
def generate(self, dataset, checkpoint_dir=None) -> Dataset:
```

where `checkpoint_dir` is the full path (which includes the specific leaf_node)

markmc: PipelineContext contains runtime parameters that do not differ over calls to 'generate()' - the base directory for checkpoints fits within that category. A name to identify a particular pipeline run (i.e. a call to generate()) makes sense as a parameter to 'generate()'

PR #236, opened 2024-08-06
Document dataset formats

markmc:
Document the various dataset formats used as input, output, and internally in the library. Related to #226.


markmc:
Merging. Not really expecting feedback on this, and it's easy to resolve later if I got something wrong.

PR #237, opened 2024-08-06
Document how to mix in pregenerated skills dataset

bbrowning:
This documents how a user can mix in the pregenerated InstructLab community dataset for skills training. It's a bit manual at the moment, but until we get a nicer user experience this at least shares how to do it.

### DIFF
--- a/docs/data_mixing.md
+++ b/docs/data_mixing.md
@@ -1,0 +1,41 @@
+# Data Mixing
+
+As one of the last steps in data generation, the SDG library can optionally mix multiple datasets into a single output dataset in proportions specified by a recipe yaml file. The current implementation is designed to be used with mostly static recipes, that get used by default for every `ilab data generate` run. There is not yet an easy way to specify the recipe to use with each generation run, but we do make it possible to change the default recipe used for skills and/or knowledge data generation.
+
+The primary intended use of this is to specify an optional pregenerated dataset maintained by the InstructLab community that can improve training results when attempting to teach new skills to a model. This process is a bit manual for now, and the steps to do that are documented below.
+
+## Using InstructLab Community Pregenerated Dataset
+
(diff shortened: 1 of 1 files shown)


docs/data_mixing.md
```diff
@@ -0,0 +1,41 @@
...
+
+The primary intended use of this is to specify an optional pregenerated dataset maintained by the InstructLab community that can improve training results when attempting to teach new skills to a model. This process is a bit manual for now, and the steps to do that are documented below.
+
+## Using InstructLab Community Pregenerated Dataset
+
+To use the [InstructLab Community pregenerated dataset](https://huggingface.co/datasets/instructlab/InstructLabCommunity) with all skills training, we first need to create a default recipe that specifies this dataset to include when mixing generated skills data. This recipe will get automatically picked up if placed in a `default_data_recipes/skills.yaml` subfolder and file under one of several possible locations - `'/home/<user>/.local/share/instructlab/sdg'`, `'/usr/local/share/instructlab/sdg'`, or `'/usr/share/instructlab/sdg'`. The exact list of possible locations is platform-dependent, and can be enumerated by a Python command like below:
```
markmc, line 9: It's sufficient to just say the file is at `$XDG_DATA_DIRS/default_data_recipes/skills.yaml` as per https://specifications.freedesktop.org/basedir-spec/latest/#referencing

markmc (approved):
Nice. Thanks, Ben

khaledsulayman (commented):
might be worth adding a section explaining to users how to write their own recipe entries. Could be showing an example so they see what it looks like in the yaml, or just briefly documenting `path`, `num_samples`, and `sampling_ratio`. WDYT?

aakankshaduggal (approved):
Thanks @bbrowning

markmc:
> might be worth adding a section explaining to users how to write their own recipe entries. Could be showing an example so they see what it looks like in the yaml, or just briefly documenting `path`, `num_samples`, and `sampling_ratio`. WDYT?

That's a bit beyond what Ben was trying to document here - simply how to use the pregenerated InstructLab community dataset

PR #242, opened 2024-08-15
Add taxonomy_base='null' option to support using the full taxonomy repo contents

danmcp:
Add a new `null` symbolic value `taxonomy_base`, to allow users to request using the entire taxonomy contents as input to the data generation pipeline.

The reason for this change is the to enable an end to end flow from generate -> train -> evaluate with a complete taxonomy repo.  With the rest of the pipeline as written, using the full taxonomy repo is how a user would start with a base model and be able to add a full taxonomy on top.  You can also think about this in terms of maturity along a git pipeline.  A full pipeline run you would expect to be run further down an integration chain where the taxonomy had already been committed, pushed, and reviewed.

Note: If we do later want to enable a taxonomy diff scenario for the full pipeline, we would need to handle the data mixing from previous layers of the taxonomy as part of the pipeline.

We expect users will use this option in conjunction with the `full` pipeline. For the `simple` pipeline, the use case is more about making quick changes to the taxonomy and getting a relatively quick turnaround on understanding if the change made a difference.  Which is why the diff approach makes sense there.

A full user flow might something look like:
- Make change to the taxonomy on local hardware
- ilab taxonomy diff
- ilab data generate --pipeline simple
- ilab model train
- ilab model chat # to check out the results
- commit taxonomy changes and push
- pull taxonomy onto heavier duty hardware
- ilab data generate --pipeline full --taxonomy-base null
- ilab model train
- ilab model evaluate

### DIFF
--- a/src/instructlab/sdg/utils/taxonomy.py
+++ b/src/instructlab/sdg/utils/taxonomy.py
@@ -97,6 +97,17 @@ def _get_taxonomy_diff(repo="taxonomy", base="origin/main"):
     return updated_taxonomy_files
 
 
+def _get_taxonomy(repo="taxonomy"):
+    repo = Path(repo)
+    taxonomy_file_paths = []
+    for root, _, files in os.walk(repo):
+        for file in files:
+            file_path = Path(root).joinpath(file).relative_to(repo)
+            if _istaxonomyfile(file_path):
+                taxonomy_file_paths.append(str(file_path))
+    return taxonomy_file_paths
+
+
 def _get_documents(
     source: Dict[str, Union[str, List[str]]],
     skip_checkout: bool = False,
@@ -400,15 +411,19 @@ def read_taxonomy(taxonomy, taxonomy_base, yaml_rules):
         if errors:
             raise SystemExit(yaml.YAMLError("Taxonomy file with errors! Exiting."))
     else:  # taxonomy is dir
-        # Gather the new or changed YAMLs using git diff
-        updated_taxonomy_files = _get_taxonomy_diff(taxonomy, taxonomy_base)
+        if taxonomy_base == "empty":
+            # Gather all the yamls - equivalent to a diff against "the null tree"
+            taxonomy_files = _get_taxonomy(taxonomy)
+        else:
+            # Gather the new or changed YAMLs using git diff, including untracked files
+            taxonomy_files = _get_taxonomy_diff(taxonomy, taxonomy_base)
         total_errors = 0
         total_warnings = 0
-        if updated_taxonomy_files:
-            logger.debug("Found new taxonomy files:")
-            for e in updated_taxonomy_files:
+        if taxonomy_files:
+            logger.debug("Found taxonomy files:")
+            for e in taxonomy_files:
                 logger.debug(f"* {e}")
-        for f in updated_taxonomy_files:
+        for f in taxonomy_files:
             file_path = os.path.join(taxonomy, f)
             data, warnings, errors = _read_taxonomy_file(file_path, yaml_rules)
             total_warnings += warnings
--- a/tests/test_taxonomy.py
+++ b/tests/test_taxonomy.py
@@ -33,8 +33,6 @@ task_description: 'This is a task'
 
 TEST_SEED_EXAMPLE = "Can you help me debug this failing unit test?"
 
-TEST_TAXONOMY_BASE = "main"
-
 TEST_CUSTOM_YAML_RULES = b"""extends: relaxed
 
 rules:
@@ -50,26 +48,41 @@ class TestTaxonomy:
     def _init_taxonomy(self, taxonomy_dir):
         self.taxonomy = taxonomy_dir
 
-    def test_read_taxonomy_leaf_nodes(self):
+    @pytest.mark.parametrize(
+        "taxonomy_base, create_tracked_file, create_untracked_file, check_leaf_node_keys",
+        [
+            ("main", True, True, ["compositional_skills->new"]),
+        ],
+    )
+    def test_read_taxonomy_leaf_nodes(
+        self,
+        taxonomy_base,
+        create_tracked_file,
+        create_untracked_file,
+        check_leaf_node_keys,
+    ):
         tracked_file = "compositional_skills/tracked/qna.yaml"
         untracked_file = "compositional_skills/new/qna.yaml"
-        self.taxonomy.add_tracked(tracked_file, TEST_VALID_COMPOSITIONAL_SKILL_YAML)
-        self.taxonomy.create_untracked(
-            untracked_file, TEST_VALID_COMPOSITIONAL_SKILL_YAML
+        if create_tracked_file:
+            self.taxonomy.add_tracked(tracked_file, TEST_VALID_COMPOSITIONAL_SKILL_YAML)
+        if create_untracked_file:
+            self.taxonomy.create_untracked(
+                untracked_file, TEST_VALID_COMPOSITIONAL_SKILL_YAML
+            )
+
+        leaf_nodes = taxonomy.read_taxonomy_leaf_nodes(
+            self.taxonomy.root, taxonomy_base, TEST_CUSTOM_YAML_RULES
         )
 
-        leaf_node = taxonomy.read_taxonomy_leaf_nodes(
-            self.taxonomy.root, TEST_TAXONOMY_BASE, TEST_CUSTOM_YAML_RULES
-        )
-        leaf_node_key = str(pathlib.Path(untracked_file).parent).replace(
-            os.path.sep, "->"
-        )
-        assert leaf_node_key in leaf_node
+        assert len(leaf_nodes) == len(check_leaf_node_keys)
 
-        leaf_node_entries = leaf_node.get(leaf_node_key)
-        seed_example_exists = False
-        if any(
-            entry["instruction"] == TEST_SEED_EXAMPLE for entry in leaf_node_entries
-        ):
-            seed_example_exists = True
-        assert seed_example_exists is True
+        for leaf_node_key in check_leaf_node_keys:
+            assert leaf_node_key in leaf_nodes
+
+            leaf_node_entries = leaf_nodes.get(leaf_node_key)
+            seed_example_exists = False
+            if any(
+                entry["instruction"] == TEST_SEED_EXAMPLE for entry in leaf_node_entries
+            ):
+                seed_example_exists = True
+            assert seed_example_exists is True
--- a/tests/test_taxonomy.py
+++ b/tests/test_taxonomy.py
@@ -52,6 +52,16 @@ class TestTaxonomy:
         "taxonomy_base, create_tracked_file, create_untracked_file, check_leaf_node_keys",
         [
             ("main", True, True, ["compositional_skills->new"]),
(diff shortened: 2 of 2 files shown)


bbrowning:
@danmcp If you want CI to also test the `full` pipeline, you'll need to manually trigger the job at https://github.com/instructlab/sdg/actions/workflows/e2e-nvidia-a10g-x4.yml for this PR.

With that said, I just completed two runs of the full pipeline locally with your change using a skeleton sdg taxonomy that just contains two skills and a single knowledge leaf node. It generated data for all 3 leaf nodes on the first run, as expected. On the second run, it loaded all data from checkpoints and didn't regenerate. So, at least from that basic check, it appears this is working as intended on the full pipeline.

bbrowning:
It looks like that full E2E test I referenced no longer starts successfully - not sure what happened there, but it's unrelated to this change.

bbrowning (approved):
Switching the read_taxonomy_leaf_nodes param from `simple` to `diff` made the code easier to follow, so thanks for iterating on that as you worked on this.

As I mentioned to you elsewhere, I don't love overloading the simple pipeline here to swap between diff and full behavior. Ideally that parameter would only control the pipeline itself - specifically the pipeline config or alias referenced. I understand it's a compromise for the sake of time, so not holding up approval for that but do think we should consider revisiting this in the future, perhaps once we get some more feedback from users on the typical flow they prefer when generating data for training.

A small nit is I'd love a bit more explanation in the PR description and/or commit message about the why of this change, just so our future selves remember why we did this.

We'll also have to think how to handle this in https://github.com/instructlab/instructlab?tab=readme-ov-file#-generate-a-synthetic-dataset , perhaps by letting the user know that the full pipeline will generate the entire taxonomy, so they either need to use a skeleton taxonomy repo with only their changes or pass `--taxonomy-path` to a single qna.yaml file to constrain data generation to that one file.

I'm approving with these comments in the interest of time, and trust you to merge when ready.

Thanks!

danmcp:
> I'd love a bit more explanation in the PR description and/or commit message

Done!

n1hility (approved):

markmc:
>I don't love overloading the simple pipeline here to swap between diff and full behavior. Ideally that parameter would only control the pipeline itself - specifically the pipeline config or alias referenced. I understand it's a compromise for the sake of time, so not holding up approval for that but do think we should consider revisiting this in the future

Agree, but I'm a bit more reluctant to defer resolving this to the future - I'm more worried about introducing some really weird semantics that we'll have to retain compatibility for

The API is:

```
def generate_data(
    ...,
    taxonomy: Optional[str] = None,
    taxonomy_base: Optional[str] = None,
    ...,
    pipeline: Optional[str] = "simple",
):
```

The `taxonomy` and `taxonomy_base` parameters control what taxonomy "leaf nodes" are used as input, and `pipeline` controls what data generation steps to perform on that input - we really should maintain that separation

And then on the UX side in `instructlab/instructlab`, consider side-effects like if you're running `data generate --pipeline!=simple` means that `taxonomy_base` is ignored? It's all very messy and difficult to explain to users.

On the other hand, `taxonomy_base` is the already a valid way of controlling "just my taxonomy contributions" vs "everything in the taxonomy tree"

And if e.g. users of the `full` pipeline should use the entire taxonomy tree, then I'd prefer to document how to use `taxonomy_base` for that for now, and perhaps improve that UX in future. Also, if downstream distributions want to default to `full` and "entire taxonomy tree", they can do so in their default configs with `taxonomy_base`.

The main downside is the fragility of using a commit SHA1 ID in docs or default configs - there's all sorts of scenarios where that SHA1 ID might change, e.g. if the user rebases the taxonomy tree or uses the weird `ilab config init --min-taxonomy` option

So IMO we can add a simply symbolic value for `taxonomy_base` to represent "empty taxonomy" - going to do some more digging before recommending something specific because this is not straightforward in git:

- We could have an empty commit published, e.g. `origin/empty` and use that ... I suspect we'd actually need to rebase onto it to ensure it gets cloned though, and it's probably not much less fragile
- `4b825dc642cb6eb9a060e54bf8d69288fbee4904` is "well-known" as a SHA1 ID for an empty tree, which is the result of with `git hash-object -t tree /dev/null` - so you can do a `git diff` against that. Our code currently only handles commit IDs though, would need some minor changes to handle tree IDs.
- `git diff-tree --root` is another way this is done - `GitPython` does this if you do for `other=NULL_TREE` ... but we also want the `other=None` behaviour of diffing against the working tree

The latter path is most promising - I'll continue digging into that

markmc:
Ok, bit of a rabbit hole on `git diff-tree --root` ... it does not appear to do what we want

```
$ git log --oneline
ccf7030 (HEAD -> main, origin/main, origin/HEAD) phoenix qna.yaml
7289074 Create knowledge path and attribution.txt
40fd450 complete common expressions qna.yaml added
a764e5d comp skills freeform skill path and attribution.txt added
21b279c punctuation qna.yaml
eaa09f7 comp skills grounded path and attribution.txt
4ef1cb6 Initial commit
$ git diff-tree ccf7030 -r
ccf70309902d5dddda75d8efd5cc538541e1f706
:000000 100644 0000000000000000000000000000000000000000 9c913a84d0200846139cac37c3f2744cc8ad2def A	knowledge/science/astronomy/constellations/phoenix/qna.yaml
$ git diff-tree 4ef1cb6 ccf7030 -r
:000000 100644 0000000000000000000000000000000000000000 4cdaa20145f1c8e4d261866fd411b2271a2b2daf A	compositional_skills/linguistics/freeform/complete_common_expressions/attribution.txt
:000000 100644 0000000000000000000000000000000000000000 8b2bd847f2ab6f22b4d1a9c080831d0e37b79543 A	compositional_skills/linguistics/freeform/complete_common_expressions/qna.yaml
:000000 100644 0000000000000000000000000000000000000000 25bc0adb1010bbe6392a13f3f5565516c4e4981d A	compositional_skills/writing/grounded/editing/punctuation/attribution.txt
:000000 100644 0000000000000000000000000000000000000000 6bbcd2b7cc91c906c6db352cb9957c3d5896bafb A	compositional_skills/writing/grounded/editing/punctuation/qna.yaml
:000000 100644 0000000000000000000000000000000000000000 049fd9ae9365d5b37227cd95ebad30c0f822c373 A	knowledge/science/astronomy/constellations/phoenix/attribution.txt
:000000 100644 0000000000000000000000000000000000000000 9c913a84d0200846139cac37c3f2744cc8ad2def A	knowledge/science/astronomy/constellations/phoenix/qna.yaml
$ git diff-tree ccf7030 -r --root
ccf70309902d5dddda75d8efd5cc538541e1f706
:000000 100644 0000000000000000000000000000000000000000 9c913a84d0200846139cac37c3f2744cc8ad2def A	knowledge/science/astronomy/constellations/phoenix/qna.yaml
$ git diff-tree 4ef1cb6 ccf7030 -r --root
:000000 100644 0000000000000000000000000000000000000000 4cdaa20145f1c8e4d261866fd411b2271a2b2daf A	compositional_skills/linguistics/freeform/complete_common_expressions/attribution.txt
:000000 100644 0000000000000000000000000000000000000000 8b2bd847f2ab6f22b4d1a9c080831d0e37b79543 A	compositional_skills/linguistics/freeform/complete_common_expressions/qna.yaml
:000000 100644 0000000000000000000000000000000000000000 25bc0adb1010bbe6392a13f3f5565516c4e4981d A	compositional_skills/writing/grounded/editing/punctuation/attribution.txt
:000000 100644 0000000000000000000000000000000000000000 6bbcd2b7cc91c906c6db352cb9957c3d5896bafb A	compositional_skills/writing/grounded/editing/punctuation/qna.yaml
:000000 100644 0000000000000000000000000000000000000000 049fd9ae9365d5b37227cd95ebad30c0f822c373 A	knowledge/science/astronomy/constellations/phoenix/attribution.txt
:000000 100644 0000000000000000000000000000000000000000 9c913a84d0200846139cac37c3f2744cc8ad2def A	knowledge/science/astronomy/constellations/phoenix/qna.yaml
$ git diff-tree 4ef1cb6 -r
$ git diff-tree 4ef1cb6 -r --root
4ef1cb6f4edd7ce7074bba29e3fb76987b8409f4
:000000 100644 0000000000000000000000000000000000000000 261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64 A	LICENSE
:000000 100644 0000000000000000000000000000000000000000 6f39879c48a04a3187593919a96ed290dabce3d7 A	README.md
```

i.e. I don't see a way of using `git diff-tree` to get a diff that includes the contents of the initial commit

I'm going back looking at adding support for `taxonomy_base=4b825dc642cb6eb9a060e54bf8d69288fbee4904`

bbrowning:
@markmc If we have a special symbolic value for `taxonomy_base` to signify no base (ie use the entire taxonomy), then it would be pretty straightforward to check for that value and divert to the new ` _get_taxonomy` added in this PR as opposed to doing the diff at all when that special value is used. Still going down the diff route and finding a special sha may be more confusing to users than documenting that a value of `none`, `empty`, `empty_tree`, or something along those lines triggers the full taxonomy tree to run.

markmc:
> I'm going back looking at adding support for `taxonomy_base=4b825dc642cb6eb9a060e54bf8d69288fbee4904`

See #244

markmc:
> @markmc If we have a special symbolic value for `taxonomy_base` to signify no base (ie use the entire taxonomy), then it would be pretty straightforward to check for that value and divert to the new ` _get_taxonomy` added in this PR as opposed to doing the diff at all when that special value is used. Still going down the diff route and finding a special sha may be more confusing to users than documenting that a value of `none`, `empty`, `empty_tree`, or something along those lines triggers the full taxonomy tree to run.

Yep, that's a really great point, Ben

And the code already has a case where it's getting a list of files without diff - untracked files (it took me a while to spot this, honestly!)

So, yes - I'll update Dan's patch to support taxonomy_base=`null`

I'm suggesting `null` because "the NULL tree" is at least mentioned as a concept in `git diff-tree`

markmc:
Just pushed an untested commit with that change in approach - looking at adding a unit test now, but maybe you could give it a while Ben?

bbrowning:
> Just pushed an untested commit with that change in approach - looking at adding a unit test now, but maybe you could give it a while Ben?

I tested your initial change (before the more recent commits to add a test and other cleanup) and the logic is working well for me locally - `--taxonomy-base null` used my entire taxonomy tree as expected, and other values used the git diff as expected.

markmc:
Thanks Ben. I've also updated the PR title and description to reflect this approach.

Will wait for @danmcp and @n1hility before merging

markmc:
> The only part I am a bit concerned about is when coming from instructlab config
> 
> ```
> taxonomy_base: 'null'
> ```
> 
> will work and
> 
> ```
> taxonomy_base: null
> ```
> 
> will not. null ends of getting the default of origin/main. The logic is still sound, but might be confusing.

Yeah, that's super confusing. Good catch. Will we go for `empty` instead?

bbrowning:
Ahh yes that's a really good catch that `null` has a special meaning in YAML and won't just automatically become a string with value `null` without quotes.

markmc:
Ok, updated to use `taxonomy_base=empty` :+1:

n1hility:
e2e passed before, confident latest changes are safe

PR #244, opened 2024-08-16
WIP: Support taxonomy_base=null as a way to specify that the entire contents of a taxonomy repo should be used

markmc:
This allows users to specify that they want all taxonomy files in their repo to be included, where 'null' symbolically represents the empty tree before this repo was created.

See #242 for more background.

TODO:

- [ ] Add a unit test
- [ ] Propose an `instructlab/instructlab` PR to make `ilab taxonomy diff` support this also
- [ ] Add documentation

### DIFF
--- a/src/instructlab/sdg/utils/taxonomy.py
+++ b/src/instructlab/sdg/utils/taxonomy.py
@@ -55,14 +55,14 @@ def _get_taxonomy_diff(repo="taxonomy", base="origin/main"):
 
     branches = [b.name for b in repo.branches]
 
-    head_commit = None
+    base_object = None
     if "/" in base:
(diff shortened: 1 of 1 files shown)


markmc:
From #242:

> @markmc If we have a special symbolic value for `taxonomy_base` to signify no base (ie use the entire taxonomy), then it would be pretty straightforward to check for that value and divert to the new ` _get_taxonomy` added in this PR as opposed to doing the diff at all when that special value is used. Still going down the diff route and finding a special sha may be more confusing to users than documenting that a value of `none`, `empty`, `empty_tree`, or something along those lines triggers the full taxonomy tree to run.

Great point ... closing this again, we can pursue that path in #242

PR #246, opened 2024-08-18
add foundational_skills to taxonomy folders

relyt0925:
Updated TAXONOMY_FOLDERS to include 'foundational_skills' to enable sdg parsing and generation of freeform skills

fixes: https://github.com/instructlab/sdg/issues/247


relyt0925:
I see foundational skills was ditched in latest taxonomy

Issue #251, opened 2024-08-21
Incorrect Number of Distractor Documents in RAFT-style Prompting Function


aakankshaduggal:
The `_add_extra_contexts_to_samples` function has a bug in its logic for selecting distractor documents when applying RAFT-style prompting. This bug results in inconsistencies in how context documents are selected and combined, potentially leading to unintended behavior.

**Problems Identified:**

1. **Sampling Error:** The function incorrectly samples five documents instead of the intended four, leading to an incorrect number of documents in the context.

2. **Conditional Addition of the Answer Document:** The function only adds the answer document if the sampling results in fewer than five documents. This is incorrect; the answer document should always be included, regardless of the number of documents sampled.

cc - @abhi1092 @shivchander

bbrowning:
I'm not Aakanksha, but yes #252 closes this issue.

hickeyma:
Closing as fixed by #252

PR #252, opened 2024-08-22
Fix selection logic for distractor documents 

aakankshaduggal:
Addresses #251

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -314,7 +314,7 @@ def _add_extra_contexts_to_samples(ds: Dataset, p, num_doc_in_context=4):
     all_context = list(set(ds["context"]))
 
     def __pick_documents(rec, p):
-        answer_document = [rec["context"]]
+        answer_document = rec["context"]
         selected_docs = [e for e in all_context if e != answer_document]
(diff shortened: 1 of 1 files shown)


bbrowning:
After some discussion with @abhi1092 on Slack, I'm adding some additional context on this change and why we need it.

"
So the important fix is in the document selection for context. What should ideally happen is it should select 4 documents and add the answer document to it. What is happening right are two things
We are sampling 5 documents instead of 4
We only add answer document if the sampling is less than 5
Both were wrong because we want to add answer document in all cases. We just want to sample if there are more than 4 doc or add all of them. But we always want to add answer_document
"

I've reviewed these changes and the logic looks sound to me - we'll now always add the answer document in the golden path. And, the comparison where we were filtering answer_document out of selected_documents should now work - previously we were comparing contexts to answer_document wrapped in an outer list, so the comparison failed when we thought it would succeed, which would let answer_document end up in selected_documents which was not intended.

bbrowning (approved):
We should expand upon the tests in `test_datamixing.py` to reproduce and catch this, but since this is a time-sensitive fix I can follow-up with a test afterwards that can get merged in a more leisurely fashion.

danmcp:
@mergifyio backport release-v0.2

PR #254, opened 2024-08-22
Add more tests for golden/distractor context picking

bbrowning:
When picking additional contexts to act as golden or distractor documents, we had a bug where the answer document wasn't always being added. This was fixed in PR #252, and this just adds some additional tests in that area to prevent regression.


bbrowning:
While this is the right code, it's from the wrong source branch and should be from my fork. Closing and reopening.

PR #264, opened 2024-09-03
ci: add AWS tag to show github PR number for all jobs

nathan-weinberg:
right now we can see what type of runner and which repository an AWS runner is associated with, but not which PR number triggered it

see https://github.com/instructlab/instructlab/pull/2183 for reference

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x4.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x4.yml
@@ -38,7 +38,8 @@ jobs:
           aws-resource-tags: >
             [
               {"Key": "Name", "Value": "instructlab-ci-github-large-runner"},
-              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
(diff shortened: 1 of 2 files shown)


aakankshaduggal (approved):

danmcp (changes requested):
I think we determined from here: https://github.com/instructlab/instructlab/pull/2183 this probably isn't the best approach.

nathan-weinberg:
> I think we determined from here: [instructlab/instructlab#2183](https://github.com/instructlab/instructlab/pull/2183) this probably isn't the best approach.

Should we try `github.event.pull_request.number`? cc @russellb

danmcp:
> > I think we determined from here: [instructlab/instructlab#2183](https://github.com/instructlab/instructlab/pull/2183) this probably isn't the best approach.
> 
> Should we try `github.event.pull_request.number`? cc @russellb

It looks like `github.event.number` should work

danmcp (approved):

alinaryan (approved):

khaledsulayman (approved):

PR #269, opened 2024-09-10
chore: replace platformdirs with xdg-base-dirs

jaideepr97:
we recently replaced `platformdirs` with `xdg-base-dirs` in the instructLab CLI
(https://github.com/instructlab/instructlab/pull/2202)
this is a follow up PR to make the same update in the sdg library 

This has no effect on linux, but has an impact for macOS

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,9 +6,9 @@ httpx>=0.25.0,<1.0.0
 instructlab-schema>=0.4.0
 langchain-text-splitters
 openai>=1.13.3,<2.0.0
-platformdirs>=4.2
 # Note: this dependency goes along with langchain-text-splitters and may be
 #       removed once that one is removed.
 # do not use 8.4.0 due to a bug in the library
 # https://github.com/instructlab/instructlab/issues/1389
 tenacity>=8.3.0,!=8.4.0
+xdg-base-dirs>=6.0.1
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -6,9 +6,9 @@ httpx>=0.25.0,<1.0.0
 instructlab-schema>=0.4.0
 langchain-text-splitters
 openai>=1.13.3,<2.0.0
-platformdirs>=4.2
 # Note: this dependency goes along with langchain-text-splitters and may be
 #       removed once that one is removed.
 # do not use 8.4.0 due to a bug in the library
 # https://github.com/instructlab/instructlab/issues/1389
 tenacity>=8.3.0,!=8.4.0
+xdg-base-dirs>=6.0.1
```
bjhargrave, line 14: Missing EOL.

bjhargrave:
> this is a follow up PR to make the same update in the sdg library

We will then need a followup PR in instructlab to raise the dependency version for instructlab-sdg to the version which includes this PR.

requirements.txt
jaideepr97: updated, thanks!

bjhargrave (approved):

bbrowning (approved):
The code looks functionally correct to me. I pulled the changes locally and manually verified that it attempts to load pipeline configs from `~/.local/share/instructlab/sdg/pipelines` and `/usr/local/share/instructlab/sdg/pipelines` if they exist.

I would love to see us enhance this with some unit tests that verify our assumptions about where we load the configs from. However, given there's also need to have a discussion about whether the library should directly write to the filesystem at all or leave that up to the CLI calling it, this is something we can defer until that's decided upon.

nathan-weinberg (approved):

PR #272, opened 2024-09-12
Handle empty dataset from output of sdg leaf node without raising error

### LINKED ISSUE #240: Empty dataset error kills the workflow
When the qna.yaml is not appropriate or a wrong model is used, the generation fails to happen and throws an error --
`instructlab.sdg.pipeline.EmptyDatasetError: Pipeline stopped: Empty dataset after running pipe`

Proposed solution:

- [ ] We need this to fail gracefully and continue operations if there were multiple leaf nodes.
- [ ] Provide more information on the error so that the user knows what steps need to be followed to rectify the issue.

relyt0925:
Previously, an EmptyDatasetError was raised when the dataset was empty after running the sdg pipeline of a leaf node. This change logs a warning and continues processing instead, allowing the function to handle empty datasets more gracefully and process other leaf nodes in the taxonomy. Fixes https://github.com/instructlab/sdg/issues/240

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -25,7 +25,6 @@ from instructlab.sdg.llmblock import MODEL_FAMILY_MERLINITE, MODEL_FAMILY_MIXTRA
 from instructlab.sdg.pipeline import (
     FULL_PIPELINES_PACKAGE,
     SIMPLE_PIPELINES_PACKAGE,
-    EmptyDatasetError,
     Pipeline,
     PipelineContext,
 )
@@ -359,6 +358,7 @@ def generate_data(
         )
 
     generated_data = None
+    empty_sdg_leaf_nodes = []
     for leaf_node in leaf_nodes.values():
         is_knowledge = False
         leaf_node_path = leaf_node[0]["taxonomy_path"].replace("->", "_")
@@ -382,9 +382,11 @@ def generate_data(
         logger.debug("Dataset: %s", ds)
         new_generated_data = pipe.generate(ds, leaf_node_path)
         if len(new_generated_data) == 0:
-            raise EmptyDatasetError(
-                "Pipeline stopped: Empty dataset after running pipe"
+            empty_sdg_leaf_nodes.append(new_generated_data)
+            logger.warning(
+                "Empty dataset for qna node: %s", leaf_node[0]["taxonomy_path"]
             )
+            continue
         generated_data = (
             [new_generated_data]
             if generated_data is None
@@ -418,3 +420,9 @@ def generate_data(
 
     generate_duration = time.time() - generate_start
     logger.info(f"Generation took {generate_duration:.2f}s")
+    if len(empty_sdg_leaf_nodes) > 0:
+        logger.warning(
+            "Leaf nodes with empty sdg output: {}".format(
+                " ".join(empty_sdg_leaf_nodes)
+            )
+        )
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -382,10 +382,8 @@ def generate_data(
         logger.debug("Dataset: %s", ds)
         new_generated_data = pipe.generate(ds, leaf_node_path)
         if len(new_generated_data) == 0:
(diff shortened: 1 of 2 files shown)


bbrowning:
Thanks for the pull request! Are there valid use-cases where a user expects an empty dataset from a leaf node? In other words, are we saving the user trouble by logging but ignoring any empty datasets? Or are we potentially masking a problem that they need to fix and re-run the entire generation? This might be the right way to address this, but I'm just trying to think through the user workflow here, how the user knows they have a problem, how they fix the problem, and whether a fatal error or just a warning log message makes that easier to spot and fix.

If it is appropriate to just log a message, perhaps we should also keep track of any taxonomy leaf nodes that resulted in empty datasets and log a final warning message at the very end of the generate run summarizing those? The single log message from each leaf node may be easy to miss during the generation loop if multiple leaf nodes are involved.

relyt0925:
I think logging a summary at the end is a great idea! 

I do believe it’s valid for if a user is generating sdg against a full taxonomy (using the taxonomy-base empty parameter) that one leaf node out of 140 for example (that is about the number of leaf nodes in the community taxonomy) shouldn’t cause a complete failure of the sdg run.

bbrowning:
Your example of the entire community taxonomy is a good one, and I agree that even if we can't generate data for all leaf nodes, generating what we can and logging the leaf nodes that failed is better than aborting entirely after multiple hours of generation.

relyt0925:
/hold

ready for review but want to do an e2e test before merging.

bbrowning:
I was going to ask you to please add a unit test in `test_generate_data.py` to verify that when a mocked generate block returns an empty dataset that the generation still runs to completion instead of erroring out. However, the code in generate_data.py needs a bit of refactoring to make that easy, and until that's done it requires quite a bit of mocking out things in generate_data to do that test. There are a few examples of this in test_generate_data.py if you want to take a stab at that, but if not this looks reasonable to merge as-is and we can create a separate issue to do some code cleanup and refactoring to make it easier to write new unit tests for the main generation loop.

src/instructlab/sdg/generate_data.py
```diff
@@ -418,3 +420,9 @@ def generate_data(
 
     generate_duration = time.time() - generate_start
     logger.info(f"Generation took {generate_duration:.2f}s")
+    if len(empty_sdg_leaf_nodes) > 0:
+        logger.warning(
+            "Leaf nodes with empty sdg output: {}".format(
+                " ".join(empty_sdg_leaf_nodes)
```
bbrowning, line 426: From testing this manually locally:
```
  File "/home/bbrownin/src/instructlab/sdg/venv/lib/python3.11/site-packages/instructlab/data/generate.py", line 293, in generate
    generate_data(
  File "/home/bbrownin/src/instructlab/sdg/src/instructlab/sdg/generate_data.py", line 426, in generate_data
    " ".join(empty_sdg_leaf_nodes)
TypeError: sequence item 0: expected str instance, Dataset found
```

bbrowning (changes requested):
I ran this manually locally and hit an error that my eyes-only review didn't catch. I added a comment inline to the code with the error, but it looks like we're trying to concatenate a string with a Dataset.

src/instructlab/sdg/generate_data.py
bbrowning: I pushed a new commit to address this issue, as well as add a test to confirm that we don't error and log a proper warning message at the end of the generation. So, let's reverse our roles and you review my change - feel free to amend, rebase, etc as desired. Thanks!

relyt0925:
Ben: Thank you so much for taking the time to illustrate that unit test example: I will note that general pattern for future PRs as well. These changes look great and I approve!!!!

relyt0925:
/unhold

testing complete

bbrowning (approved):

relyt0925:
@instructlab/sdg-maintainers do we think this is a potential candidate to merge? It seems like it is impacting some client workflows when they are utilizing generation against the *full* taxonomy

aakankshaduggal (approved):
LGTM, Thanks @relyt0925 for your contribution!!

PR #281, opened 2024-09-18
[Docs] Updates for SDG README

kelbrown20:
**Description:**
Adding some more content to the SDG README.md


README.md
```diff
@@ -1,8 +1,66 @@
...
+ from instructlab.sdg.generate_data import generate_data
+ from instructlab.sdg.utils import GenerateException
+```
+
+<!--Not sure what more your thinking of adding here -->
+
+## Pipelines 
+
+There are four pipelines that are used in SDG. Each pipeline requires specific hardware specifications. 
+<!--TODO: Add explanations of pipelines-->
```
kelbrown20, line 41: Not sure If I have the right context here

aakankshaduggal: Thanks @kelbrown20. This is a great start. But for the supported pipelines with SDG we only have `simple` and `full`, eval and schema are sub parts for some pipelines if and when needed. So let us limit to just simple and full. If we were to go in depth we could talk about separate pipelines for freeform, grounded and knowledge for simple and full. WDYT?

bbrowning: Today we only ship a "full" and a "simple" pipeline, although users can technically define their own pipelines as well. Exact hardware requirements for each is fluid, as with each pipeline there's a question of the models you use. Out of the box, the "full" pipeline is designed to run on a machine large enough to host the mixtral-8x7b-instruct model, which means approximately 120GB+ of GPU memory. The closest typical machine size to that is something like 8x24GB GPUs for 192GB total of GPU memory. However, you can run this pipeline with less - sometimes even much less if you instead use a quantized mixtral teacher model - so what exactly do we want to document for hardware requirements there?

The "simple" pipeline has much lower requirements - I'm not sure the exact minimum specs, but typical consumer hardware is plenty for it.

README.md
```diff
@@ -1,8 +1,81 @@
...
+## Using the library
+
+You can use the SDG library with the following items
+
+```bash
+ from instructlab.sdg.generate_data import generate_data
+ from instructlab.sdg.utils import GenerateException
+```
+
+<!--Not sure what more your thinking of adding here -->
```
khaledsulayman, line 38:
```suggestion
You can import SDG into your Python files as such:

```python
 from instructlab.sdg.generate_data import generate_data
 from instructlab.sdg.utils import GenerateException
```
```

File: README.md [hunk 3]
```diff
@@ -1,8 +1,81 @@
...
+
+## Pipelines
+
+There are four pipelines that are used in SDG. Each pipeline requires specific hardware specifications.
+<!--TODO: Add explanations of pipelines-->
+
+*Full* -
+
+Minimum hardware requirements for running the simple pipeline
```
khaledsulayman, line 47: might need someone else to keep me honest, but I'm not too sure it makes a lot of sense to specify "minimum" requirements here. The 'full' pipeline may run on smaller infra than we typically use, but the experience might be impacted (e.g. very slow, might crash, etc.).

README.md
```diff
@@ -1,8 +1,81 @@
...
+
+*Simple* -
+
+Minimum hardware requirements for running the simple pipeline
```
khaledsulayman, line 51: In its current state, simple should be able to run on most laptops

README.md
kelbrown20: I think I was just thinking of a qualitative way to determine which pipeline to use. But more general might be better. I can say something like: This pipeline is targeted for running SDG on CPUs" or something?

README.md
kelbrown20: yeah Ill try to write something more general

bbrowning:
Kelly, how would you like me to contribute to this to fill in some of the missing sections? We can merge this as-is and I can add more commits on top in another pull request. Or, I can potentially push more commits into this pull request to fill out some of the unknowns. I'm open to either way, depending on your preference and comfort with a bit more advanced git workflow, in the case of me adding commits directly to this PR.

kelbrown20:
I was also thinking about the best way to with this for collaboration. I think pushing commits to this PR would be the best option!

bbrowning:
Sorry for the noise here - I had to update the list of spellcheck words, and that caused a merge conflict so had to rebase both of our commits on top of the latest `main` to get things happy.

kelbrown20:
Thank you for adding those technical details @bbrowning!
Are there any other details we want added here or is this a good start? cc @aakankshaduggal @khaledsulayman

aakankshaduggal (approved):
Thanks @kelbrown20 and @bbrowning. This is a good start, we can continue to add more information and documentation in following PRs.

Issue #271, opened 2024-09-11
context-aware chunking 


ktam3:
**Feature Overview (aka. Goal Summary)**
_An elevator pitch (value statement) that describes the Feature clearly and concisely. Complete during New status._

Converting a document with mixed elements (e.g., a PDF with tables, images, text, image captions, multiple text columns per page like in papers, etc.) to Markdown requires identifying the objects and processing them accordingly for a proper Markdown representation (e.g. multi-page table, as an MD table, two-column pages as paragraphs in the correct order, etc.)

To achieve this, we must start working on context-aware chunking of the ingested documents and augment the elements with additional metadata required for identifying the source, location within the source, etc.

 

**Goals (aka. expected user outcomes)**
_The observable functionality that the user now has due to receiving this feature. Include the anticipated primary user type/persona and which existing features will be expanded. Complete during New status._

This feature is to adopt a tool for context-aware chunking for preprocessing the documents during the ingestion pipeline and to be used by the SDG and other steps.

- https://github.com/DS4SD/docling
- https://github.com/DS4SD/quackling 

 

**Requirements (aka. Acceptance Criteria):**
_To be considered complete, the feature must deliver a list of specific needs or objectives. Be sure to include nonfunctional requirements such as security, reliability, performance, maintainability, scalability, usability, etc. Initial completion during Refinement status._
- The change should be transparent from the user CLI perspective
- Backward compatibility of datasets generated by previous versions is not required.

 

**Done - Checklist (mandatory)**
- CI-test validating integration with docling/quackling
- Any evaluation metrics resulting from SDG process should remain not less than 5% of the existing one or, preferably, provide significant improvements

aakankshaduggal:
Yes, @bbrowning, the goal here is twofold: First, to enable users to easily convert hefty PDFs into the expected markdown format using Docling. This allows us to support a wider range of input formats, like PDFs, in addition to markdown files. Second, we aim to enhance our chunking strategy by implementing semantic chunking for the referenced documents in the qna.yaml, ensuring that chunks are more contextually meaningful rather than relying on simple markdown boundaries with overlap.

bbrowning:
Thanks @aakankshaduggal for that clarification!

noelo:
@aakankshaduggal  for what it's worth I did a quick POC using the [semantic-chunkers](https://github.com/aurelio-labs/semantic-chunkers) component. Code is [here](https://github.com/noelo/instruct-generate/blob/main/injest-splitter.ipynb) WIP

PR #284, opened 2024-09-23
Integrate Context-Aware Chunking and PDF Support

khaledsulayman:
Add context-aware document chunking for PDFs.

Checking by filetype to decide whether to use the context-aware chunker (pdf) or text-splitter (md).
Currently uses docking v1 json format.

Resolves: #271


bbrowning:
I'm keeping track of the work here, but withholding any feedback until you get to a point where you think it's ready. If you need any help tracking down CI or other test failures while working on this, let me know and I'll make time.

src/instructlab/sdg/utils/chunking.py
bbrowning, line 1: We'll need to also rename `tests/test_chunking.py` to `tests/test_docprocessor.py` and update its tests to pass in the new expected values to the `chunk_documents` method.

requirements.txt
```diff
@@ -10,5 +10,8 @@ openai>=1.13.3,<2.0.0
 #       removed once that one is removed.
 # do not use 8.4.0 due to a bug in the library
 # https://github.com/instructlab/instructlab/issues/1389
+pypdf>=5.0.0
```
bbrowning, line 13: Do we actually need pypdf here? We're using it to read the contents of the PDF files, but then we're never actually using the contents of PDF files downstream. If we really want to read the contents of the PDF, we could use something like https://github.com/DS4SD/docling-parse instead as that's already a dependency of docling anyway, so won't be bringing in anything extra?

requirements.txt
```diff
@@ -10,5 +10,8 @@ openai>=1.13.3,<2.0.0
...
+tabulate>=0.9.0
 tenacity>=8.3.0,!=8.4.0
+transformers>=4.44.2
```
bbrowning, line 16: Should we align this with instructlab/instructlab? Currently that's on 4.41.2 at https://github.com/instructlab/instructlab/blob/main/requirements.txt#L34

requirements.txt
bbrowning, line 1: We're missing an entry for `docling` here to get docling itself added as a requirement. And, since docling depends on torch, we'll want to also add a dependency on torch to pin its version to the same range as instructlab/instructlab - https://github.com/instructlab/instructlab/blob/main/requirements.txt#L31 (currently `torch>=2.3.0,<2.4.0`)

src/instructlab/sdg/utils/docprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+# Third Party
+from datasets import Dataset, concatenate_datasets
+from tabulate import tabulate
+from transformers import AutoTokenizer
+import yaml
+
+# Local
+
+logger = logging.getLogger(__name__)
+DOC_FILEPATH = Path("~/.local/share/instructlab/documents").expanduser()
```
bbrowning, line 30: This path should probably be somewhere under the `output_dir` from generate_data.py. Since we'll create new docs here for every run, perhaps under the `node_datasets_*` subdirectory for that run?

src/instructlab/sdg/utils/docprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+        return None
+
+    return concatenate_datasets(filtered_datasets)
+
+
+class DocProcessor:
+    def __init__(
+        self,
+        parsed_doc_dir: Path,
+        tokenizer_model_name: str = "TheBloke/Mixtral-8x7B-Instruct-v0.1-GGUF",
```
bbrowning, line 342: This should come from the model_name parameter passed to generate_data.py's `generate_data`. That way we load the tokenizer for the same teacher model that's being used.

src/instructlab/sdg/utils/docprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+
+    print(f"""THIS IS KHALED: CHUNKING PDF DOCS
+        {pdf_docs[0]=}
+    """)
+
+    converter = DocumentConverter(pipeline_options=PipelineOptions())
+    parsed_pdfs = [converter.convert_single(d) for d in pdf_docs]
+    parsed_dicts = [p.render_as_dict() for p in parsed_pdfs]
+
+    docling_jsons_path = DOC_FILEPATH / "docling-jsons"
```
bbrowning, line 618: Prefer `os.path.join` to combine paths to string concatenation - `os.path.join` handles various edge cases like ensuring we don't add double slashes and more complex paths.

src/instructlab/sdg/utils/taxonomy.py
```diff
@@ -108,37 +110,55 @@ def _get_documents(
...
+                        # Process PDF files
+                        with open(file_path, "rb") as file:
+                            reader = PdfReader(file)  # Use pypdf PdfReader
+                            pdf_text = ""
+                            for page in reader.pages:  # Iterating through pages
+                                pdf_text += page.extract_text()
+                            file_contents.append(pdf_text)
+
+        if file_contents:
+            filepaths = [DOC_FILEPATH / p for p in file_patterns]
```
bbrowning, line 155: Prefer `os.path.join` to combine paths to string concatenation - `os.path.join` handles various edge cases like ensuring we don't add double slashes and more complex paths.

src/instructlab/sdg/utils/docprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+            {"chunks": "document"}
+        )
+
+        # Only keep document greater than 100 tokens
+        new_ds = new_ds.filter(
+            lambda x: get_token_count(x["document"], self.tokenizer) > 100
+        )
+        return new_ds
+
+    def get_processed_dataset(self) -> Dataset:
```
bbrowning, line 468: What calls this method? I couldn't find it being used anywhere.

src/instructlab/sdg/utils/docprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+        Process all the parsed docling json files and return a dataset.
+        Returns:
+            Dataset: Dataset object.
+        """
+        datasets = []
+        for json_fp in self.docling_jsons:
+            chunk_ds = self._process_parsed_docling_json(json_fp)
+            chunk_ds_with_icls = self._add_icls(chunk_ds)
```
bbrowning, line 477: Where do we add the icl values for non-PDF files? This codepath only happens for PDFs, right?

bbrowning (changes requested):
I did an initial once-over - may not have caught everything, and some of the things I caught I think will be obvious once we can run tests (including end-to-end CI tests) on this.

We'll want to add a pdf file and data generation from it to our existing end-to-end tests to exercise that code-path, as there's a lot of new logic here that we'll want to ensure works in the CI environment.

I tried to focus mostly on the issues that will either cause packaging headaches (keeping our dependencies as slim as possible), user-impacting issues running this across multiple systems (file path handling), or breakage to things like existing markdown functionality. We can clean up other bits later once we get clean end-to-end tests working for markdown and PDFs.

requirements.txt
bbrowning: Here's a suggested change to requirements.txt that adds docling and aligns a couple of dependencies with the CLI repo:
```
diff --git a/requirements.txt b/requirements.txt
index 195dddf..84bb75c 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,17 +1,20 @@
 click>=8.1.7,<9.0.0
 datasets>=2.18.0,<3.0.0
+docling>=1.15.0,<2.0.0
 GitPython>=3.1.42,<4.0.0
 httpx>=0.25.0,<1.0.0
 instructlab-schema>=0.4.0
 langchain-text-splitters
 openai>=1.13.3,<2.0.0
+pypdf>=5.0.0
+tabulate>=0.9.0
 # Note: this dependency goes along with langchain-text-splitters and may be
 #       removed once that one is removed.
 # do not use 8.4.0 due to a bug in the library
 # https://github.com/instructlab/instructlab/issues/1389
-pypdf>=5.0.0
-tabulate>=0.9.0
 tenacity>=8.3.0,!=8.4.0
-transformers>=4.44.2
+# align torch with instructlab/instructlab
+torch>=2.3.0,<2.4.0
+transformers>=4.41.2
 xdg-base-dirs>=6.0.1
```
That would be enough to at least get some of these tests running, although until test_chunking.py is adjusted it's going to just error on every run.

src/instructlab/sdg/utils/docprocessor.py
```diff
@@ -0,0 +1,637 @@
...
+            for entry in pd:
+                json_file.write(json.dumps(entry) + "\n")
+
+    dp = DocProcessor(
+        parsed_doc_dir=str(docling_jsons_path),
+        tokenizer_model_name=model_name,
+        qna_yaml_path=Path("~/.local/share/instructlab/taxonomy").expanduser() / leaf_node_path / "qna.yaml",
+    )
+
+    chunked_pdfs = list(dp.get_processed_dataset())
```
aakankshaduggal, line 603: @bbrowning we are calling `get_processed_dataset` here.

bbrowning: Ahh, yes. That was added between my comment above and this comment, so marking the comment above as resolved :smile:

bbrowning:
Digging into things a bit more trying to run them locally, there are multiple model artifacts we'll have to deal with. Docling brings 2 of its own models, but it also uses the `easyocr` module which downloads 1+ models from GitHub as well depending on usage options. And, we also need to ensure the tokenizer lives on disk from our teacher model. So, that's at least 4 things we need to help the user explicitly download as opposed to attempting to do it silenty/transparently behind the scenes, as part of data generation - the teacher model's tokenizer, the 2 docling models, and the 1+ easyocr models. We'll also need to disable the implicit download of these probably, to be consistent with elsewhere in InstructLab where we expect models to be explicitly downloaded to disk. Or, at least surface that option to the user.

Additionally, the easyocr models prefer to run on a GPU and log a warning if only run on the CPU. We'll need to decide how we want to handle that, and whether we now expect the data generation process to also consume GPU capacity in addition to our teacher model that already runs on a GPU in a separate (sometimes external to this machine) process.

n1hility:
I agree we probably do not want implicit downloading and execution of models (at least by default, and at least without thinking about how that should apply with the other models). It would be fine if we logged an error to run ilab model download.

On the easyocr also agree. vllm is quite sensitive to competing GPU use for example, and will hard fail with activity on the same GPU during init, plus vllm can be greedy for its kv cache. How essential / impactful is GPU acceleration for this use case?

bbrowning:
I haven't done any timings yet myself to see how long a large PDF document will take to convert using only CPU vs GPU. The docling team is working to add an alternative OCR implementation (tesseract) that doesn't require GPU nor downloading any models 3rd-party sources. So, once that lands, we should be able to remove the concerns around easyocr at least.

src/instructlab/sdg/utils/docprocessor.py
bbrowning: Ok, digging into this a bit more, we shouldn't be doing anything different for icl values with PDFs, right? I'm confused why we do anything with icl_query_* here at all, as that all gets handled in taxonomy.py already. What's the intent behind the separate duplicate icl handling for pdfs?

bbrowning:
I have a fork of this work where I'm doing a basic attempt at a pdf chunking test and getting it passing end-to-end to get things wired through at https://github.com/bbrowning/instructlab-sdg/commits/docprocessor-updates/ . The changes there at least get the stubbed in test passing, but the test isn't validating everything it needs to and I think the chunks are still not quite right.

aakankshaduggal:
@bbrowning Thanks for testing the PR. Given we had a little more time for this issue, to avoid any hard coded paths or just checking for mds or pdfs, we are planning to do a better redesign that makes ingestion of other datatypes in the future a lot easier that it was to add the support for pdfs.

src/instructlab/sdg/utils/taxonomy.py
```diff
@@ -21,9 +24,13 @@
 import yaml
 
 # Local
-from . import chunking
+from .chunkers import DocumentChunker
+
+# Initialize the pdf parser
+parser = pdf_parser()
```
jaideepr97, line 30: @khaledsulayman @bbrowning 
do we know if this internally downloads the docling models automatically or would we need to externally ensure that those models are present before invoking the parsing?

jaideepr97: I would like to ensure that we error out if those models are not available on disk rather than silently download them somehow

tests/test_chunkers.py
```diff
@@ -0,0 +1,168 @@
...
+        assert (
+            "Got a larger chunk overlap (100) than chunk size (24), should be smaller"
+            in str(exc.value)
+        )
+
+
+class TestContextAwareChunker():
+    @pytest.fixture
+    def chunker(documents_dir):
+
```
bbrowning, line 129: Attempting to run the unit tests from this PR locally (via `tox -e py3-unit`), I'm getting `IndentationError: expected an indented block after function definition on line 128` from this line because there's nothing in the function body - ie needs something here for the code to actually run.

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -0,0 +1,563 @@
...
+        leaf_node_path,
+        output_dir: Path,
+        tokenizer_model_name=None,
+    ):
+        self.document_paths = document_paths
+        self.filepaths = filepaths
+        self.leaf_node_path = leaf_node_path
+        self.output_dir = self._path_validator(output_dir)
+        if tokenizer_model_name is None:
+            self.tokenizer_model_name = "mistralai/Mixtral-8x7B-Instruct-v0.1"
```
jwm4, line 194: Why not just make this the default for tokenizer_model_name instead of having it default to None and then override that here?

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -0,0 +1,563 @@
...
+            text_list (list): List of text chunks to process.
+            short_length_threshold (int): The word count threshold for determining short texts.
+        Returns:
+            list: List of fused texts.
+        """
+        fused_texts = []
+        previous_long_text = ""
+
+        for text in text_list:
+            word_count = len(text.split())
```
jwm4, line 288: Why is this using `len(text.split())` instead of just calling the tokenizer?  It seems odd to me to have a tokenizer for counting tokens during the initial chunking but then count words for the fusing.  Probably fine for now though.

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -0,0 +1,563 @@
...
+                        document_chunks.append("\n\n".join(current_buffer[:-1]))
+
+                        if (
+                            self.get_token_count(current_buffer[-1], tokenizer)
+                            >= max_token_per_chunk
+                        ):
+                            document_chunks.append(current_buffer[-1])
+                            current_buffer = []
+                        else:
+                            current_buffer = current_buffer[-1:]
```
jwm4, line 509: I can't make any sense of what this code is intended to do if it hits a book_text whose size is greater than max_token_per_chunk.  It looks to me like it just wipes out everything in the buffer _except_ the item that is too big and then continues?  I don't get it.

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -0,0 +1,563 @@
...
+    def chunk_documents(self) -> Dataset:
+        """Semantically chunk PDF documents.
+
+        Returns:
+            List: a list of chunks from the documents
+        """
+        if self.document_paths == []:
+            return []
+
+        model_artifacts_path = DocumentConverter.download_models_hf()
```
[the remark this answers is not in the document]
aakankshaduggal: Thanks @jwm4! We can address this as a part of this https://github.com/instructlab/sdg/issues/335 
For now maybe we will just add a print statement to state that "Downloading model from hugging face"

requirements.txt
```diff
@@ -1,14 +1,18 @@
 # SPDX-License-Identifier: Apache-2.0
 click>=8.1.7,<9.0.0
 datasets>=2.18.0,<3.0.0
+docling>=1.15.0,<2.0.0
```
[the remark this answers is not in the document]
aakankshaduggal: Yes, we can address this as a part of https://github.com/instructlab/sdg/issues/333

src/instructlab/sdg/utils/chunkers.py
aakankshaduggal: Thanks @jwm4, addressed!

src/instructlab/sdg/utils/chunkers.py
aakankshaduggal: Thanks @jwm4, addressed!

jwm4: @aakankshaduggal is the issue that @khaledsulayman raised going to be a problem with the new implementation?  It does look like if something passes an explicit None that this will fail when it tries to initialize the autotokenizer.

aakankshaduggal: yeah, makes sense. I will do a none check to ensure it isn't empty because initializing then

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -0,0 +1,609 @@
...
+                    current_buffer = []
+                else:
+                    if (
+                        self.get_token_count("\n\n".join(current_buffer), tokenizer)
+                        >= max_token_per_chunk
+                        and len(current_buffer) > 1
+                    ):
+                        chunk_text = "\n\n".join(current_buffer[:-1])
+                        print(
+                            f"Current chunk size {self.get_token_count(chunk_text, tokenizer)} and max is {max_token_per_chunk}"
```
jwm4, line 507: Is this print statement really helpful?  I would expect something like this in a debug level logging event or something like that.

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -0,0 +1,609 @@
...
+                        )
+
+                        document_chunks.append("\n\n".join(current_buffer[:-1]))
+
+                        if (
+                            self.get_token_count(current_buffer[-1], tokenizer)
+                            >= max_token_per_chunk
+                        ):
+                            print(
+                                f"This is too big a document to be left in the current buffer {self.get_token_count(current_buffer[-1], tokenizer)}"
```
jwm4, line 517: Maybe this would be useful as a warning level logging event?  Even for that purpose, it is not clear to me that this is well written for consumption by an end user.  For example, it is saying "too big a document" but if I understand it correctly, the issue is not the size of the _document_ but rather the size of `book_element["text"]`, which isn't going to mean much to an end user.  So if we do have a warning level log it might wind up being something more like "The following text was dropped from {document_name} because it was too long to fit into a single context for synthetic data generation: {current_buffer[-1]}".  (Not sure where you get `document_name` from, however -- maybe it is in json_book somewhere?)

jwm4 (changes requested):
Broadly speaking, I still don't understand what this chunking code is doing.  I think it needs a lot more comments if someone wants to make sense of it.  However, maybe we just accept that as tech debt for 1.3 and move to something new based on Docling 2.x features for 1.4 anyway.  If that's the decision, I can live with it.  I'd still like to get rid of the print statements in the chunker though.

PR #286, opened 2024-09-24
ci: add additional autolabeling rules

nathan-weinberg:
the CLI repo has mergify autolabel rules that seem to have never made it over to this repo, this commit adds those in so the autolabeling config aligns


bbrowning (approved):
Looks good, and as always the consistency / cleanup work is much appreciated!

khaledsulayman (approved):
Great catch! Thanks for making this change!

PR #287, opened 2024-09-25
github: add stale bot to sdg repo

nathan-weinberg:
Same action as https://github.com/instructlab/instructlab/blob/main/.github/workflows/stale_bot.yml


khaledsulayman (approved):

khaledsulayman:
This will be helpful. Thanks!

bbrowning (approved):
Approving, but also realizing this will trigger on a number of existing outstanding issues so we'll need to do a round of following up for anything we want to stay open once this lands and starts marking things as stale.

PR #302, opened 2024-10-10
Mistral Family support and Logging

cdoern:
Add support for mistral models. They use the same prompt template as Mixtral. This dramatically increases the performance of the full pipeline on a laptop using LLamaCPP. The issue with Mixtral is that even in GGUF form the model is too big for most consumer hardware. Completions and prompting using this model hang for long periods of time.

The logging I added is optional debug logs that print all of the prompts in each LLMBlock and the prompt currently being generated.

I also added a loading bar for each LLMBlock pass that increments each time we get a prompt response back.

### DIFF
--- a/src/instructlab/sdg/llmblock.py
+++ b/src/instructlab/sdg/llmblock.py
@@ -8,6 +8,7 @@ import re
 
 # Third Party
 from datasets import Dataset
+from tqdm import tqdm
 import openai
 
 # Local
@@ -18,7 +19,7 @@ logger = logging.getLogger(__name__)
 MODEL_FAMILY_MIXTRAL = "mixtral"
 MODEL_FAMILY_MERLINITE = "merlinite"
 
-_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST]"
+_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST] </s>"
 _MODEL_PROMPT_MERLINITE = "'<|system|>\nYou are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior.\n<|user|>\n{prompt}\n<|assistant|>\n'"
 
 _MODEL_PROMPTS = {
@@ -157,7 +158,7 @@ class LLMBlock(Block):
 
     def _generate(self, samples) -> list:
         prompts = [self._format_prompt(sample) for sample in samples]
-
+        logger.debug(f"STARTING GENERATION FOR LLMBlock USING PROMPTS: {prompts}")
         if self.server_supports_batched:
             response = self.ctx.client.completions.create(
                 prompt=prompts, **self.gen_kwargs
@@ -165,12 +166,17 @@ class LLMBlock(Block):
             return [choice.text.strip() for choice in response.choices]
 
         results = []
+        progress_bar = tqdm(
+            range(len(prompts)), desc=f"{self.block_name} Prompt Generation"
+        )
         for prompt in prompts:
+            logger.debug(f"CREATING COMPLETION FOR PROMPT: {prompt}")
             for _ in range(self.gen_kwargs.get("n", 1)):
                 response = self.ctx.client.completions.create(
                     prompt=prompt, **self.gen_kwargs
                 )
                 results.append(response.choices[0].text.strip())
+                progress_bar.update(1)
         return results
 
     def generate(self, samples: Dataset) -> Dataset:
--- a/src/instructlab/sdg/utils/models.py
+++ b/src/instructlab/sdg/utils/models.py
@@ -14,9 +14,7 @@ DEFAULT_MODEL_FAMILY = "merlinite"
 MODEL_FAMILIES = set(("merlinite", "mixtral"))
 
 # Map model names to their family
-MODEL_FAMILY_MAPPINGS = {
-    "granite": "merlinite",
-}
+MODEL_FAMILY_MAPPINGS = {"granite": "merlinite", "mistral": "mixtral"}
 
 
 def get_model_family(model_family, model_path):


khaledsulayman (commented):
Looks good to me! Happy to approve once linting passes

aakankshaduggal (approved):
Thanks @cdoern for contributing to SDG 🚢

nathan-weinberg:
Would like to wait and see if E2E passes

src/instructlab/sdg/llmblock.py
```diff
@@ -18,7 +19,7 @@
 MODEL_FAMILY_MIXTRAL = "mixtral"
 MODEL_FAMILY_MERLINITE = "merlinite"
 
-_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST]"
+_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST] </s>"
```
bbrowning, line 22: Should we be adding the eos token `</s>` here? In most of the mixtral examples I've seen, and in their official chat template at https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1/blob/main/tokenizer_config.json, the eos token is only added after assistant messages but not the user messages. Aren't our prompts entirely user messages here, from the standpoint of the model?

aakankshaduggal: I had a similar hunch @bbrowning, from the looks of the tests that @cdoern ran, it seems to work. Wdyt should be the approach here?

cdoern: the reason I added this in was because I was getting errors from llamacpp saying we had an opening start token with no matching stop token and it may impact results, this went away when I added the </s> in

bbrowning: Hmm - at a minimum we'll want to make sure adding the eos token here doesn't regress our results when using the full mixtral-8x7b-instruct model. It may not care a lot about its eos token, or it may degrade the output because it only expects the eos token after each user/assistant pair like if you're doing few-shot prompting with one or more examples in your prompt.

The safe option would be to not add the eos token here and leave it as it was before - sometimes mixtral/mistral can be picky about getting its prompt format right.

cdoern: sure, we can run some tests or we can just remove it. @aakankshaduggal WDYT?

aakankshaduggal: I agree with rolling it back for now. It’s safer to revert and then revisit once we have a better understanding of how different models handle the eos token in user-only messages. We should test this specifically for `llamacpp` vs. `vllm` and document any behavioral differences. That way, we’ll know for sure whether or not this change is needed and avoid unexpected regressions. Let’s add an issue to revisit this after running some tests. Wdyt @cdoern and @bbrowning

bbrowning: Reverting just this part of the PR seems reasonable to me as well - I'm basing this on things like https://github.com/huggingface/blog/blob/main/mixtral.md#prompt-format where it says, `This format has to be exactly reproduced for effective use` or reported issues like https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1/discussions/59 where it says, `Make sure you use the white spaces and BOS/EOS tokens appropriately. The EOS tokens are very important. And the whitespaces are also very important.`

With that said, we probably should verify exactly what's making its way into vllm or llama_cpp here, in case something like llama_cpp_python is doing some tweaking because it doesn't like the non-existent trailing eos token in the prompt.

nathan-weinberg: Note that if you revert, this will still be part of the `0.4.0` release - you'll need to backport the revert and release a `0.4.1`

Given the effort, I would recommend doing the research ahead of time instead and seeing if a revert is really necessary.

bbrowning: That's reasonable - let's figure out if this negatively impacts us or if it's a no-op as far as our generated data goes and then move from there.

bbrowning: Here's just one data point from me manually hitting a mixtral-8x7b completions API with and without the eos token. I tested this several times and got similar results each time as far as overall quality, with the quality of the responses being MUCH worse when I include the trailing eos token.

## Without trailing eos token
Request:
```
{
  "model": "mixtral-8x7b-instruct",
  "prompt": [
    "<s> [INST] What is the capital of France? [/INST]"
  ],
  "max_tokens": 2048,
  "n": 1
}
```

Response:
```
{
  "id": "cmpl-d67522a9148b4778a286801e5661ab97",
  "object": "text_completion",
  "created": 1728655847,
  "model": "mixtral-8x7b-instruct",
  "choices": [
    {
      "index": 0,
      "text": " The capital city of France is Paris. Known for its beautiful architecture, museums, and romantic ambiance, Paris is one of the most popular tourist destinations in the world. It is situated in the north-central part of France, along the River Seine. Some of the famous landmarks in Paris include the Eiffel Tower, Notre-Dame Cathedral, the Louvre Museum, Montmartre, and the Champs-Élysées. Paris is also the center of French politics, economy, culture, and fashion, making it an influential city on the global stage.",
      "logprobs": null,
      "finish_reason": "stop",
      "stop_reason": null
    }
  ],
  "usage": {
    "prompt_tokens": 17,
    "total_tokens": 141,
    "completion_tokens": 124
  }
}
```

## With trailing eos token
```
{
  "model": "mixtral-8x7b-instruct",
  "prompt": [
    "<s> [INST] What is the capital of France? [/INST] </s>"
  ],
  "max_tokens": 2048,
  "n": 1
}
```

Response:
```
{
  "id": "cmpl-9cb064df0ea7446ca9119f0108af0cd3",
  "object": "text_completion",
  "created": 1728655788,
  "model": "mixtral-8x7b-instruct",
  "choices": [
    {
      "index": 0,
      "text": "'Re friend, I don't have personal experience or emotions. I'm designed to provide information and answer questions to the best of my knowledge. As for your query, the capital of France is Paris. It's renowned for its beautiful architecture, art museums, like the Louvre, and fashion houses. Is there anything else you'd like to know?",
      "logprobs": null,
      "finish_reason": "stop",
      "stop_reason": null
    }
  ],
  "usage": {
    "prompt_tokens": 19,
    "total_tokens": 97,
    "completion_tokens": 78
  }
}
```

For this very simple prompt, every time I include a trailing eos token the answer makes no sense and is unrelated to my question. Sometimes I'm able to get it to produce a decent response with the trailing eos token, depending on my question. However, it just as often if not more often goes entirely off the subject with unrelated responses.

bbrowning: I'm seeing similar poor results with the trailing eos token testing this against a mistral-7b.

## mistral-7b without trailing eos token
Prompt:
```
<s> [INST] Generate a haiku that explains the pythagorean theorem. [/INST]
```

Response:
```
Squared right, three parts,\n\nA equals B squared and C's square,\nHarmony in triangles.
```

## mistral-7b with trailing eos token
Prompt:
```
<s> [INST] Generate a haiku that explains the pythagorean theorem. [/INST] </s>
```

Response:
```
shape Tomes squared,\n\n sum then equal side two,\n Pythagoras' song.
```

The first Haiku makes sense and generally follows the instructions. The second Haiku is markedly worse, getting the math of the theorem wrong, a random reference to "Tomes", and so on.

bbrowning: I don't think it's urgent, and would love to see if others can reproduce my results on either unquantized mixtral-8x7b or mistral-7b just to confirm it's not something only I see.

PR #303, opened 2024-10-10
fix: e2e job using wrong flags

nathan-weinberg:
@instructlab/sdg-maintainers this job should be kept in alignment with https://github.com/instructlab/instructlab/blob/main/.github/workflows/e2e-nvidia-t4-x1.yml

### DIFF
--- a/.github/workflows/e2e-nvidia-t4-x1.yml
+++ b/.github/workflows/e2e-nvidia-t4-x1.yml
@@ -144,7 +144,7 @@ jobs:
         working-directory: ./instructlab
         run: |
           . venv/bin/activate
-          ./scripts/basic-workflow-tests.sh -m
+          ./scripts/basic-workflow-tests.sh -msq
 
   stop-runner:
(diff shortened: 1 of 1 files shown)


cdoern (approved):
thanks

bbrowning (approved):

bbrowning:
Is there a way to tell mergify to ignore the failing test here, since this change is supposed to fix that failing test? Or is this a case of we just need to merge it manually?

cdoern:
@bbrowning I think its a manual merge situation. I have hit this weird dependency loop in instructlab before. Only way around it is to trigger custom runners.

khaledsulayman (approved):
Thanks for this, Nathan!

bbrowning:
Ok, I manually merged and we'll verify this fixes the e2e the next time it gets triggered.

bbrowning:
It looks like the e2e is passing again - thanks for the fix!

PR #304, opened 2024-10-10
Update Python 3.12

fabiendupont:
Now that PyTorch 2.4 supports Python 3.12, This change attempts updating the default Python version to 3.12, jointly with the other repositories of the instructlab project.

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x4.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x4.yml
@@ -114,27 +114,27 @@ jobs:
       - name: Install Packages
         run: |
           cat /etc/os-release
-          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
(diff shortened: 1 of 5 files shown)


khaledsulayman:
Looks like on the main repo https://github.com/instructlab/instructlab/pull/2434 was closed in favor of https://github.com/instructlab/instructlab/pull/2445, but it doesn't appear that the latter updated Python. Is this still relevant @fabiendupont ?

bbrowning:
I'm closing this one in favor of a slimmed down version opened in #524 - that newer PR only adds Python 3.12 to our unit and functional test matrix, which will allow us to ensure SDG works with Python 3.12. It still runs the `ilab` end-to-end tests with Python 3.11, since InstructLab only supports Python 3.11.

bbrowning:
Oh, and because I started from this PR as a base, I retained co-authorship over in #524. Thanks for the contribution!

Issue #336, opened 2024-11-05
feat: Allow support for user supplied system prompts


jaideepr97:
Currently SDG uses a hardcoded system prompt that can be found here: https://github.com/instructlab/sdg/blob/main/src/instructlab/sdg/generate_data.py#L39

This is probably not sustainable long term as users will want to supply their own specific system prompts to be used for data generation. This issue proposes moving away from hardcoded system prompts and making it a parameter that can be passed into SDG either from the CLI or some other workflow

PR #339, opened 2024-11-06
feat: parametrize system prompt

jaideepr97:
This PR parametrizes the system prompt so that users of SDG can pass their own system prompt in to the SDG process

fixes: #336

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -35,7 +35,7 @@ from instructlab.sdg.utils.taxonomy import (
 
 logger = logging.getLogger(__name__)
 
-_SYS_PROMPT = "I am, Red Hat® Instruct Model based on Granite 7B, an AI language model developed by Red Hat and IBM Research, based on the Granite-7b-base language model. My primary function is to be a chat assistant."
+_SYS_PROMPT = "I am a Red Hat® Instruct Model, an AI language model developed by Red Hat and IBM Research based on the granite-3.0-8b-base model. My primary role is to serve as a chat assistant."
 
 
 def _unescape(s):
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -73,7 +73,9 @@ def _convert_to_messages(sample):
     return sample
 
 
-def _gen_train_data(machine_instruction_data, output_file_train, output_file_messages):
+def _gen_train_data(
+    machine_instruction_data, output_file_train, output_file_messages, system_prompt
+):
     """
     Generate training data in the legacy system/user/assistant format
     used in train_*.jsonl as well as the legacy messages format used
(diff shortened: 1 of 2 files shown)


jaideepr97:
This can wait until https://github.com/instructlab/sdg/pull/284 is in

jaideepr97:
I think we have a chicken and egg problem here - CI is likely failing because the CLI version being used for the medium and large jobs doesn't know to pass the sys prompt in as a parameter
but the CLI PR to do this will also not pass until this PR is merged into SDG and a new release is cut 

cc @nathan-weinberg

jaideepr97:
> In a way this is a good thing, as it forces us to think about backwards compatibility of our SDG library API. For example, instead of making the new system prompt required, we could make it optional and default to the current system prompt if not passed. That would have no issue passing the current CI, and then once this gets merged and a new SDG release cut we could bump the min SDG version in the CLI repo and change it to pass in the optional system prompt parameter.

ah yeah, makes sense. I should've thought of that haha
thanks @bbrowning !

aakankshaduggal (approved):
Thanks @jaideepr97!

jaideepr97:
@aakankshaduggal @bbrowning 
I'm going to update this PR to include the student_model_arch as a param that can be used to determine the correct pretraining data format for further work

aakankshaduggal:
> @aakankshaduggal @bbrowning I'm going to update this PR to include the student_model_arch as a param that can be used to determine the correct pretraining data format for further work

Sounds good, thanks @jaideepr97 that would be amazing!

jaideepr97:
on second thought, it made more sense to have a separate PR for it 
so I took a stab at it here: https://github.com/instructlab/sdg/pull/341

src/instructlab/sdg/generate_data.py
```diff
@@ -94,15 +96,15 @@ def _gen_train_data(machine_instruction_data, output_file_train, output_file_mes
                 user += "\n" + synth_example["context"]
             assistant = _unescape(_get_response_hack(synth_example))
             train_entry = {
-                "system": _SYS_PROMPT,
+                "system": system_prompt if system_prompt is not None else _SYS_PROMPT,
```
Maxusmusti, line 99: Since we are still using a default system prompt as a fallback, we probably want to update that to be the 8B prompt by default

jaideepr97: updated, thanks

bbrowning (commented):
A nice side-effect of the last commit that updates the default system prompt to Granite 3.0 is our test is now also validating that the user can pass in a custom system prompt to `generate_data` and it actually takes effect, since the system prompt we pass in there is the older prompt. 

Assuming this passes the e2e tests, the code itself looks good to me. Thanks for centralizing the logic that guards against the `None` system prompt value as well!

I'm holding off on actual approval until we get 284 in, as it should be very close and the surface area here for resolving any merge/rebase conflicts is lower than in 284.

khaledsulayman (commented):
Thanks for this, Jaideep!

Echoing Ben's held-off approval until we get #284 in. Happy to work with you to help rebase if you need.

khaledsulayman (approved):

PR #340, opened 2024-11-06
feat: expose max_num_tokens as configurable

cdoern:
max-num-tokens is a nice way to run a shorter or longer SDG run.
locally I have been modifiyng the pipeline yaml from 2048 to 512 which ends up just generating less data
exposing this to the CLI could allow power users to run different types of SDG runs!

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -20,7 +20,11 @@ import openai
 # pylint: disable=ungrouped-imports
 from instructlab.sdg.datamixing import DataMixer, _get_question_hack, _get_response_hack
 from instructlab.sdg.eval_data import generate_eval_task_data, mmlubench_pipe_init
-from instructlab.sdg.llmblock import MODEL_FAMILY_MERLINITE, MODEL_FAMILY_MIXTRAL
+from instructlab.sdg.llmblock import (
+    DEFAULT_MAX_NUM_TOKENS,
+    MODEL_FAMILY_MERLINITE,
+    MODEL_FAMILY_MIXTRAL,
+)
 from instructlab.sdg.pipeline import (
     FULL_PIPELINES_PACKAGE,
     SIMPLE_PIPELINES_PACKAGE,
@@ -183,6 +187,7 @@ def _context_init(
     save_freq: int,
     batch_num_workers: Optional[int],
     batch_size: Optional[int],
+    max_num_tokens: Optional[int] = DEFAULT_MAX_NUM_TOKENS,
 ):
     extra_kwargs = {}
     if batch_size is not None:
@@ -196,6 +201,7 @@ def _context_init(
         num_instructions_to_generate=num_instructions_to_generate,
         checkpoint_dir=checkpoint_dir,
         save_freq=save_freq,
+        max_num_tokens=max_num_tokens,
         **extra_kwargs,
     )
 
@@ -281,6 +287,7 @@ def generate_data(
     pipeline: Optional[str] = "simple",
     batch_size: Optional[int] = None,
     checkpoint_dir: Optional[str] = None,
+    max_num_tokens: Optional[int] = DEFAULT_MAX_NUM_TOKENS,
 ) -> None:
     """Generate data for training and testing a model.
 
@@ -343,6 +350,7 @@ def generate_data(
         1,  # save_freq
         batch_size=batch_size,
         batch_num_workers=num_cpus,
+        max_num_tokens=max_num_tokens,
     )
 
     knowledge_pipe, freeform_skills_pipe, grounded_skills_pipe = _sdg_init(
--- a/src/instructlab/sdg/llmblock.py
+++ b/src/instructlab/sdg/llmblock.py
@@ -16,6 +16,8 @@ from .block import Block
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_MAX_NUM_TOKENS = 4096
+
 MODEL_FAMILY_MIXTRAL = "mixtral"
 MODEL_FAMILY_MERLINITE = "merlinite"
 
@@ -78,11 +80,23 @@ class LLMBlock(Block):
         self.model_prompt = model_prompt
         self.output_cols = output_cols
         self.batch_params = batch_kwargs
+        max_num_token_override = ctx.max_num_tokens
         self.parser_name = parser_kwargs.get("parser_name", None)
         self.parsing_pattern = parser_kwargs.get("parsing_pattern", None)
         self.parser_cleanup_tags = parser_kwargs.get("parser_cleanup_tags", None)
+        # max_num_tokens should only be applicable to knowledge blocks
+        # gen_knowledge if the full/simple pipeline's knowledge generation block
+        if block_name != "gen_knowledge":
+            logger.debug(
+                f"Not applying max_num_tokens to block {block_name}. This is only applicable for gen_knowledge."
+            )
+            max_num_token_override = DEFAULT_MAX_NUM_TOKENS
         self.gen_kwargs = self._gen_kwargs(
-            gen_kwargs, model=self.ctx.model_id, temperature=0, max_tokens=4096
+            max_num_token_override,
+            gen_kwargs,
+            model=self.ctx.model_id,
+            temperature=0,
+            max_tokens=DEFAULT_MAX_NUM_TOKENS,
         )
         # Whether the LLM server supports a list of input prompts
         # and supports the n parameter to generate n outputs per input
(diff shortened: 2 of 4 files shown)


src/instructlab/sdg/pipeline.py
```diff
@@ -193,10 +194,11 @@ def _generate_single(self, dataset) -> Dataset:
                 block_config = block_prop["config"]
                 drop_columns = block_prop.get("drop_columns", [])
                 drop_duplicates_cols = block_prop.get("drop_duplicates", False)
+                if block_type == llmblock.LLMBlock:
+                    block_config["gen_kwargs"]["max_tokens"] = self.ctx.max_num_tokens
```
khaledsulayman, line 198: Can you move this to happen within LLMBlock's __init__ instead?

khaledsulayman (commented):
left 1 comment. Otherwise LGTM!

src/instructlab/sdg/pipeline.py
cdoern: added to `LLMBlock` so that if not None, we use the value provided, else use the value from the pipeline yaml

src/instructlab/sdg/pipeline.py
```diff
@@ -191,12 +193,23 @@ def _generate_single(self, dataset) -> Dataset:
                 block_name = block_prop["name"]
                 block_type = _lookup_block_type(block_prop["type"])
                 block_config = block_prop["config"]
+
                 drop_columns = block_prop.get("drop_columns", [])
                 drop_duplicates_cols = block_prop.get("drop_duplicates", False)
-                block = block_type(self.ctx, self, block_name, **block_config)
+                # if block_type == llmblock.LLMBlock:
+                #     block_config["gen_kwargs"]["max_tokens"] = self.ctx.max_num_tokens
```
khaledsulayman, line 200: can you remove the commented out code?

khaledsulayman, line 200:
```suggestion
```

khaledsulayman (approved):

cdoern:
The issue here was the following:

`_gen_kwargs` sets a default of 4096 for `max_tokens` unless one is explicitly set in the pipeline yaml for the block. 

This means I had to set the top level `max_num_tokens` to 4096, using that if none was specified in the block and also using it if the user didn't pass an override.

cdoern:
@bbrowning more so `This allows the user to override what we specified in the yaml`. 

@aakankshaduggal indicated that this is a good path forward for allowing CLI users to experiment with different wait times/types of results from SDG and that she has used this toggle locally to achieve different types of results.

`max_tokens` is already applied to each LLMBlock via `gen_kwargs` and it set to 4096, this change maintains that unless `max_num_tokens` is passed into `generate_data`

bbrowning:
So, my only concern here is that we have downstream pipelines that apply a different max tokens for every step in the pipeline. And the general API surface area of the pipeline config is each step can specify it's own max tokens. This gives a single max tokens for the entire pipeline if passed in, right? As opposed to letting me just tweak the step(s) in the pipeline where I actually want to change the token count?

aakankshaduggal (commented):
@cdoern Thanks for your PR! This is a great step towards enabling a quick SDG run. The goal here is to let users adjust parameters like `--sdg-scale-factor` for skills and `--max_num_tokens` for knowledge generation.

Right now, these values are set to higher defaults, which yield solid results. Making them configurable will allow for quicker runs, though the recommended defaults will remain in place. cc @bbrowning

aakankshaduggal (approved):
LGTM, thanks @cdoern

khaledsulayman (approved):

src/instructlab/sdg/generate_data.py
```diff
@@ -183,6 +183,7 @@ def _context_init(
     save_freq: int,
     batch_num_workers: Optional[int],
     batch_size: Optional[int],
+    max_num_tokens: Optional[int] = 4096,
```
bbrowning, line 186: Could we reuse the constant defined in llmblock.py for this instead of hardcoding 4096 here and elsewhere? All these files are already importing things from llmblock anyway.

src/instructlab/sdg/llmblock.py
```diff
@@ -81,8 +84,18 @@ def __init__(
...
+            logger.debug(
+                f"Not applying max_num_tokens to block {block_name}. This is only applicable for gen_knowledge."
+            )
+            max_num_tokens = DEFAULT_MAX_NUM_TOKENS
         self.gen_kwargs = self._gen_kwargs(
-            gen_kwargs, model=self.ctx.model_id, temperature=0, max_tokens=4096
+            gen_kwargs,
+            model=self.ctx.model_id,
+            temperature=0,
+            max_tokens=max_num_tokens,
```
bbrowning, line 98: I'm not sure this is doing what's intended, as the `max_tokens` here is only used as a default if none is specified in the config. But, we specify 2048 in the full and simple configs, which means we won't ever use the `max_tokens` value shown here.

tests/test_llmblock.py
bbrowning, line 1: Let's add a small additional test in this file to verify that overriding this actually works, in case I'm wrong about how the `max_tokens` param to `_gen_kwargs` gets used. Pass in a `max_num_tokens` but also a `gen_kwargs` that contains a value of 2048 (or something that's not 4096), like our default configs at https://github.com/instructlab/sdg/blob/7af918ae48c39f21ea721b1c8c8ec7400849731e/src/instructlab/sdg/pipelines/simple/knowledge.yaml#L10 do. Then, verify that after initializing the `LLMBlock`, it has the expected gen_kwargs value.

bbrowning (changes requested):
I don't think this is overriding the actual max tokens in the gen_kwargs we send to the API request like you want. I could be wrong in my reading of the code, and I left a comment on the type of unit test I think we could write to validate that this is working (or not) as intended.

src/instructlab/sdg/llmblock.py
cdoern: made this into `max_num_tokens_overwrite` as a position arg, if this arg != the default, we override the value already in gen_kwargs

tests/test_llmblock.py
cdoern: sure! I will add a test, thanks!

src/instructlab/sdg/pipeline.py
```diff
@@ -65,6 +66,7 @@ class PipelineContext:  # pylint: disable=too-many-instance-attributes
     dataset_num_procs: Optional[int] = DEFAULT_DATASET_NUM_PROCS
     checkpoint_dir: Optional[str] = None
     save_freq: Optional[int] = 1
+    max_num_tokens: Optional[int] = 4096
```
bbrowning, line 69: We could also change this 4096 to `llmblock.DEFAULT_MAX_NUM_TOKENS` to reduce the hardcoded 4096 a bit more.

tests/test_llmblock.py
```diff
@@ -86,3 +89,19 @@ def test_model_prompt_none(self, mock_load_config):
...
+        mock_load_config.return_value = self.config_return_value
+        # Ensure that if a custom model_prompt is specified, it is used correctly
+        block = LLMBlock(
+            ctx=self.mock_ctx,
+            pipe=self.mock_pipe,
+            block_name="gen_knowledge",
+            max_num_token_override=512,
+            config_path="",
+            output_cols=[],
+            model_prompt="",
```
bbrowning, line 104: We're missing one thing here to actually verify this is overriding the config - another arg of: 
```python
            gen_kwargs={"max_tokens": 2048},
```

That will simulate that we read a config file specifying 2048 as the max tokens (like our simple/full configs do) but that we're actually overriding that to 512 tokens via the new override parameter. Otherwise, the test is only ensuring we don't use the `DEFAULT_MAX_NUM_TOKENS` value but isn't ensuring we don't use the pipeline config file's values, which is what we want to overwrite.

src/instructlab/sdg/pipeline.py
```diff
@@ -191,11 +193,20 @@ def _generate_single(self, dataset) -> Dataset:
...
+                if block_type in (llmblock.LLMBlock, llmblock.ConditionalLLMBlock):
+                    block = block_type(
+                        self.ctx,
+                        self,
+                        block_name,
+                        max_num_token_override,
+                        **block_config,
+                    )
+                else:
+                    block = block_type(self.ctx, self, block_name, **block_config)
```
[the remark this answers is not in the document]
cdoern: I like this idea!

will need to adjust the tests accordingly as well since they currently depend on it being in the `LLMBlock` instantiation

bbrowning (approved):
Looks good to me - thanks for the quick turnarounds!

aakankshaduggal (approved):

khaledsulayman (approved):

PR #341, opened 2024-11-06
feat: support converting messages datasets into multiple pre-training formats

jaideepr97:
This PR adds support for  converting messages datasets into multiple pre-training formats to support working with both granite 7b and granite 3.0 student models. It accepts a `use_legacy_pretraining_format` parameter as input to appropriately choose the right format to use

This is intended to be a short term solution, with the long term idea being that SDG would be agnostic of student model requirements such as these

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -362,17 +362,24 @@ def _add_extra_contexts_to_samples(ds: Dataset, p, num_doc_in_context=4):
     return ds
 
 
-def _conv_pretrain(rec):
+def _conv_pretrain(rec, use_legacy_pretraining_format: bool):
     """
     Convert a messages dataset that contains only user/assistant entries per
     message (and in that order) to a pretraining message used downstream by
     the training pipeline. `_generate_knowledge_qa_dataset` creates the type
     of dataset expected here.
     """
+    if use_legacy_pretraining_format:
+        user = "<|user|>"
+        assistant = "<|assistant|>"
+    else:
+        user = "<|start_of_role|>user<|end_of_role|>"
+        assistant = "<|start_of_role|>assistant<|end_of_role|>"
+
     rec["messages"] = [
         {
             "role": "pretraining",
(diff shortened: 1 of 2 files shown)


src/instructlab/sdg/datamixing.py
```diff
@@ -362,17 +362,28 @@ def __pick_documents(rec, p):
     return ds
 
 
-def _conv_pretrain(rec):
+def _conv_pretrain(rec, student_model_arch: str):
```
Maxusmusti, line 365: This is fine in the CLI, but i would actually caution against having any hard ties to model architecture in SDG. Since the system prompt is already being passed in separately, SDG doesn't need to know about specific model architectures. Instead, it should probably just take in a boolean like `use_legacy_tokens`, that the CLI can set as true/false depending on the arch

Maxusmusti: So like in the CLI at the same conditional where you pick which system prompt to pass in, you could also just choose whether to set this to true false

jaideepr97: updated, thanks

bbrowning:
@Maxusmusti Can you confirm that we don't need any `<|end_of_text|>` tokens in the pre-training samples here, like the chat template at https://huggingface.co/ibm-granite/granite-3.0-8b-instruct/blob/main/tokenizer_config.json#L188 uses after the message from each role? If we need to follow that format exactly, we'd also need these `<|end_of_text|>` tokens after the text from 1 role before starting the new role tokens?

Maxusmusti:
@bbrowning yeah you don't need to add the end_of_text token, that gets added _by_ the chat template: https://github.com/instructlab/training/pull/319/files#diff-a8438361fc1435b584fec100fc73bd5bdc7856dc9826a570dd9dc7a6321f9bbcR30

aakankshaduggal (approved):
Thanks @jaideepr97! LGTM

bbrowning (approved):

khaledsulayman (approved):

Issue #346, opened 2024-11-07
SDG Testing on Context Aware Chunking


khaledsulayman:
- Decide on what CI we need for Document Chunking (different filetypes, etc.)
- Flesh out unit testing

PR #349, opened 2024-11-08
Upgrade docling, expand chunking testing

bbrowning:
This adds pytest-based functional tests to our repo with a basic PDF chunking test.

While writing that test I discovered a minor bug in DocumentChunker, resulting in an additional unit test and a minor change to handle the case where we're given no documents and were previously not instantiating any chunker.


tests/functional/test_chunkers.py
```diff
@@ -0,0 +1,72 @@
...
+# First Party
+from instructlab.sdg.utils.chunkers import DocumentChunker
+
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "testdata")
+
+# TODO: Apparently we don't really need any contents in the qna.yaml?
+knowledge_qna = """
+version: 3
+domain: astronomy
+"""
```
khaledsulayman: can probably remove this now

tests/functional/testdata/phoenix.md
khaledsulayman, line 1: could you move this to `testdata/sample_documents`?

bbrowning:
Tagging you in for a review @nathan-weinberg since I'm editing the CI files, adding functional tests in addition to our unit tests. Those are wired into CI much like the CLI repo, where we add a `tox -e py3-functional` and have the functional tests run immediately after the unit tests in the same CI job.

nathan-weinberg (approved):
Testing aspects LGTM!

tests/functional/testdata/phoenix.pdf
khaledsulayman, line 1: same as with the `.md`

aakankshaduggal (approved):
Thanks @bbrowning. This is great! Love the new functional tests 🚢

khaledsulayman (commented):
Thanks for this!

Semi-related, but it seems like 2 outstanding testing todos around this are:
1. More thorough testing around document chunking
2. (up for discussion) updates to the e2e to test PDFs

I have #346 as a placeholder issue for Chunking testing, I can flesh it out more and/or make it an epic to track those efforts if we want to do those in follow-up PRs.

nathan-weinberg:
@khaledsulayman we would def welcome some updates to the E2E CI to add PDF testing!

bbrowning:
I have some other functional tests I want to iterate on as well, not just specific to chunking, so would love to get this basic set of functional tests in and then open separate PRs for new tests. Some of the others aren't related to chunking at all, but are for things like testing that we don't cause llama_cpp to throw an assert error when doing batching and similar stuff we need to fix.

PR #358, opened 2024-11-10
Don't attempt batching with InstructLab's llama-cpp-python

bbrowning:
Try harder to check for InstructLab's default llama-cpp-python during our batching check so that we can avoid throwing assertion errors on the server when we probe for batching support.

See instructlab/instructlab#1748 for user reports of this issue.

### DIFF
--- a/src/instructlab/sdg/llmblock.py
+++ b/src/instructlab/sdg/llmblock.py
@@ -9,6 +9,7 @@ import re
 # Third Party
 from datasets import Dataset
 from tqdm import tqdm
+import httpx
 import openai
 
 # Local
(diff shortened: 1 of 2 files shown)


bbrowning:
There doesn't appear to be a robust way to identify llama-cpp-python as a HTTP client. However, InstructLab's most common happy path has users generating data against a llama-cpp-python managed by InstructLab and we CAN reliably identify that. We'll do that by looking for the [API root welcome message](https://github.com/instructlab/instructlab/blob/v0.21.0a1/src/instructlab/model/backends/common.py#L15)  and disable batching without actually attempting the batching request probe. That batching request attempt is what throws the assertion error in llama-cpp-python, and this change will avoid throwing those errors for anyone using `ilab model serve` to run their model.

bbrowning:
We get lots of user reports via Slack and in this and other repos about the assertion error with llama-cpp-python. This gets rid of that, and will make for a much cleaner initial user experience. Otherwise, we often have users that think things are broken with data generation.

khaledsulayman (approved):

aakankshaduggal (approved):
Thanks @bbrowning, lgtm!

PR #361, opened 2024-11-10
Only use CPU for the docling OCR models

bbrowning:
Because GPU memory is extremely tight in many of our supported hardware configurations, and because our GitHub Mac CI runners error out when running the OCR models with MPS acceleration, let's just explicitly pin the OCR models to the CPU.

See DS4SD/docling#286 for a bit more context.


bbrowning:
We'll need to merge this before upgrading our docling dependency above version 2.4.2 to prevent test failures on our Mac CI runners.

aakankshaduggal (approved):

khaledsulayman (approved):

PR #364, opened 2024-11-11
Check for tokenizer in downloaded models directory

### LINKED ISSUE #343: Enable Tokenizer Loading from downloaded Teacher Model
Currently, the AutoTokenizer in the chunker tries to pull the tokenizer from the teacher model, but this hasn't been tested thoroughly outside of mixtral.

By default, `AutoTokenizer.from_pretrained()` will go to huggingface for the model, and in the case of mixtral, the repo is gated so you'd need to set `$HF_TOKEN`.

What should happen is we pull the tokenizer from the downloaded teacher model and raise an error if it cannot be found.

khaledsulayman:
This change forces tokenizer loading to be done using the locally downloaded teacher model, as opposed to pulling from huggingface.

Resolves: #343 

Signed-off-by: Khaled Sulayman <ksulayma@redhat.com>


RobotSail:
It seems like we're including artifacts directly in our code for testing purpose, e.g.: `tests/testdata/models/instructlab/granite-7b-lab/tokenizer.json`.

We want to avoid this approach since the artifacts themselves may become out-of-date overtime and therefore lead to inaccurate tests, in addition to adding an unnecessary amount of data to our codebase. If possible, I recommend to instead setup a workflow in which these files are downloaded automatically as part of the test suite. The `AutoTokenizer.from_pretrained` API already does this when loading the tokenizer, so I would check if there's a more direct way to simply download the tokenizer.

khaledsulayman:
@RobotSail the point of these tests is to make sure the tokenizer is being loaded from file since we don't want implicit downloads happening during SDG.

FWIW I was able to remove most of the bulk of the safetensors dir since all we really need is the tokenizer-specific files for this. The entire directory is now about 2MB.

As far as if the formats change, as long as `AutoTokenizer.from_pretrained()` can read the tokenizer files, I don't see there being an issue here.

RobotSail:
> @RobotSail the point of these tests is to make sure the tokenizer is being loaded from file since we don't want implicit downloads happening during SDG.

Yeah, I understand this. What I'm saying here is that to avoid storing artifacts in git, we will want to instead setup a flow of downloading the necessary files _for testing_ when the tests run. Think something like a download script or a Makefile.

khaledsulayman:
@RobotSail I see your point. I personally don't feel too strongly either way, but I don't think it's worth blocking this PR on. I've opened #384 so we can continue discussion over there

aakankshaduggal (approved):
Thanks @khaledsulayman! LGTM 🚀

bbrowning (approved):
This looks good enough to me to merge and then keep improving things as we go. A few things that come to mind we may want to tackle later:

- renaming `tokenizer_model_name` to something else, since it really expects a directory and not a huggingface model name
- consider factoring out some common requirements, util classes, etc into some instructlab/common repo/package that SDG, CLI, and training repos could share - things like `is_model_gguf` or ensuring we're all aligned on identical versions of transformers, torch, etc - needs more discussion outside this PR, but this PR copies a few more things from instructlab/instructlab that we'll want to stay in-sync over time
- adding an e2e test that exercises the ContextAwareChunker - right now we have a functional test that tests PDF chunking, but we don't have something that ensures the entire e2e flow from `ilab data generate` is wired up properly for PDFs, including things like the tokenizer_model_name gets passed in as expected

Because we're hoping to get a release out soon and some of us are off tomorrow, I don't think any of the above needs to hold this up. And, some of the above would explode the scope of this PR so best done separately anyway.

Thanks for all the work on this!

Issue #365, opened 2024-11-11
Data Mixing Phase 10 - knowledge pre-training dataset not getting mixed in


khaledsulayman:
The knowledge pre-training dataset isn't getting mixed in for phase 10 usage. The result is that models don't have as good knowledge performance as they should have.

ktam3:
Per @mairin - we also want to backport this in an ilab .19.z release for a downstream z release

FYI @nathan-weinberg @cdoern

nathan-weinberg:
> Per @mairin - we also want to backport this in an ilab .19.z release for a downstream z release
> 
> FYI @nathan-weinberg @cdoern 

I'm not sure this makes any sense? 0.21.0 is almost out...

mairin:
@nathan-weinberg downstream z-release.

nathan-weinberg:
> @nathan-weinberg downstream z-release.

Right, but what I'm saying is from a support cycle standpoint, upgrading to the latest downstream release makes more sense than releasing a Z-stream no?

nathan-weinberg:
@mairin and I sorted it offline 👍

PR #366, opened 2024-11-11
Data mix fix

aakankshaduggal:
Addresses #365

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -608,7 +608,7 @@ class DataMixer:
             )
 
             skills_phase_data = _create_phase10_ds(
-                new_generated_data, self.auxiliary_inst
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
@Mergifyio backport release-v0.3

bbrowning (approved):
Looks good to me - verified locally that this does add the pretraining samples into the phase10 dataset. I wrote a small unit test to verify this locally, so will create a follow-up PR to get that merged as well without blocking or adding to this one.

aakankshaduggal:
Thanks @bbrowning for your review. I do have some tests in flight that I can add as a follow up PR.

bbrowning:
Great - will hold off on turning my simple verification into an actual PR then and defer to you on that. Thanks for following up!

khaledsulayman (approved):
I've also tested this on an ec2 with full mixtral and looks good!

PR #369, opened 2024-11-12
Prefer tesserocr over easyocr, if available

### LINKED ISSUE #352: Prefer tesserocr vs easyocr for Docling integration, when available
Docling defaults to using `easyocr` for optical character recognition, but we have some downstream consumers that will prefer to use Docling's `tesserocr` for OCR. We need to expose a way for users to influence which we use, as it requires [code changes](https://github.com/DS4SD/docling/blob/v2.4.1/docs/installation.md?plain=1#L23-L48) in our Docling integration to swap the OCR engine used.

bbrowning:
When setting up our ingestion pipeline, explicitly check if tesserocr is available and Docling can load it. If so, prefer that. Otherwise, attempt the same for EasyOCR. If neither can load, log an error and disable optical character recognition.

Fixes #352

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 click>=8.1.7,<9.0.0
 datasets>=2.18.0,<3.0.0
-docling>=2.4.2,<3.0.0
+docling[tesserocr]>=2.4.2,<3.0.0
 GitPython>=3.1.42,<4.0.0
 httpx>=0.25.0,<1.0.0
 instructlab-schema>=0.4.0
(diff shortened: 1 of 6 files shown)


aakankshaduggal (approved):

bbrowning:
Thanks for the approval! After following some discussion elsewhere about being careful when we import anything that imports all of `torch`, I'm going to add an additional test to this and defer some of the docling/easyocr imports to not import transformers or torch until they're actually needed. Just a small change, but realized that should go in as part of this because otherwise we're loading all of torch fairly early in our import chain.

bbrowning:
Ok, removing the hold now that we're not importing all of Pytorch as soon as someone imports SDG. Instead, we defer that until Docling actually needs `torch` loaded by moving some of our imports of docling bits further down into the code. And, the added test ensures we don't accidentally regress on that as we do future docling work here.

khaledsulayman (approved):

khaledsulayman:
Thanks for taking care of this, Ben! 😁

PR #370, opened 2024-11-12
chore: rename 'basic-workflow-tests' to 'e2e-custom'

bbrowning:
the test file name is changing in the CLI repo
ergo it must be changed here as well


bbrowning:
Actually, nevermind - going to fix this for the 0.3 release branch another way by pulling in an older instructlab repo vs running instructlab main e2e tests against 0.3 SDG.

PR #377, opened 2024-11-13
fix: upsample the phase10 knowledge dataset

RobotSail:
When we mix the knowledge dataset with skills today, we do not account for the potential discrepancy
in size between the generated knowledge data and skills data. This leads to the models potentially
forgetting the data it was trained on in the knowledge phase. As a simple workaround, we simply
upsample the knowledge samples before mixing them in with the generated skills dataset.

Signed-off-by: Oleg S <97077423+RobotSail@users.noreply.github.com>

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Standard
-from typing import Dict, List, Optional
+from pathlib import Path
+from typing import Dict, List, Optional, TypedDict
 import json
 import logging
 import os.path
@@ -16,8 +17,22 @@ import yaml
 from instructlab.sdg.utils import GenerateException, pandas
 from instructlab.sdg.utils.pandas import dataset_from_pandas_dataframe
 
+# XXX(osilkin): This value represents the ratio between knowledge & skills data
+#               below which we upsample knowledge samples from. This only applies
+#               when |knowledge| << |skills|
+MIN_UPSAMPLE_THRESHOLD = 0.03
 ALLOWED_COLS = ["id", "messages", "metadata"]
 logger = logging.getLogger(__name__)
+
+
+class DatasetListing(TypedDict):
+    """
+    TypedDict class that represents the dataset listings passed around in the
+    `.datasets` key of each recipe.
+    """
+
+    sampling_size: float
+    path: str
 
 
 def _adjust_train_sample_size(ds: Dataset, num_samples: int):
@@ -99,7 +114,7 @@ class Recipe:
 
         # Defaults if no recipe path given or these values don't
         # exist in the given recipe file
-        self.datasets = []
+        self.datasets: List[DatasetListing] = []
         if recipe_path is not None:
             recipe = self._load_recipe()
             if "datasets" in recipe:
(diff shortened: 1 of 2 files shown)


bbrowning:
Is upsampling a special case here? Or is it just that we need to adjust our mixing recipe in use for these knowledge leaf node(s) to have a fixed sampling size or a sampling ratio larger than the default of 1.0? See `_sample_ds` and `_adjust_train_sample_size` in `datamixing.py` for examples of what we already do today. And, you'll see that there's an optional fourth parameter to `_gen_leaf_node_data` that is the sampling size, which means we could pass in the desired fixed number of samples or ratio of samples when generating the knowledge leaf node data and then that would get written out to the recipe.yaml file as well as used for our final data mixing to scale up the knowledge samples.

src/instructlab/sdg/datamixing.py
```diff
@@ -612,6 +635,11 @@ def collect(
             skills_phase_data = _create_phase10_ds(
                 new_generated_data, self.auxiliary_inst, use_legacy_pretraining_format
             )
+            # XXX(osilkin): To prevent the size of the skills dataset from drowning out the knowledge
+            #               samples, we upsample the knowledge data here to prevent catastrophic forgetting
+            skills_phase_data = upsample_dataset(
+                skills_phase_data, self.upsample_amount
+            )
```
bbrowning, line 642: Following on from my general comment to point out the specific code here. What if instead of this, we just passed in the `upsample_amount` to the `_gen_leaf_node_data` call a few lines below here as the optional 4th param that's not used today. That sets the sampling size for the dataset, where the default value of `1.0` is unchanged but a value of `0.5` would be sampling 50% of the data, a value of `3.0` would upsample to 300%, and a value of `5000` would upsample to a fixed 5000 samples. I'm not sure the exact right amount of upsampling, but I think the existing code can handle this need.

khaledsulayman: agree with Ben here, in the initial implementation of mixing, we written it such that it would upsample if the provided `--num-samples` or `--sampling-ratio` required it. I guess that got lost somewhere, but that feels like a good way to proceed here.

RobotSail: Just doing some quick napkin math, since we found that it performs well when you have 10k resamples out of a 350k dataset, that would put us at a $training/skills$ ratio of roughly 3%. In this case, the rule we should follow would be: 

```py
if len(replay)/len(dataset) < 0.03:
  replay = upsample_ds(replay, len(dataset)*0.03)
```

RobotSail: @bbrowning In the current state of the `DataMixer`, it doesn't seem like there's a way for the `_collect` method to know the length of the skills dataset when creating the recipes for the knowledge dataset. If we want to compute the resample amount more dynamically then we would need to refactor this.

bbrowning: Yes, refactoring mixing have some knowledge across both skills and knowledge datasets would be a refactor that will be a bit larger lift to get in. Today mixing happens entirely within a single Recipe, and skills and knowledge have different Recipes that are independent of each other.

src/instructlab/sdg/llmblock.py
```diff
@@ -18,6 +18,7 @@
 logger = logging.getLogger(__name__)
 
 DEFAULT_MAX_NUM_TOKENS = 4096
+DEFAULT_KNOWLEDGE_UPSAMPLE_AMOUNT = 5000
```
bbrowning, line 21: Do we really want to default this to 5000 here? In the typical upstream scenario, users have constrained hardware and there's no precomputed skills dataset used in phase 2 that we need to balance against. Do we know how much longer this will take in the typical upstream train cycle for those new users kicking the tires?

bbrowning: I think for the typical upstream usage we may not want to upsample at all, so that would be a default value of `None` I guess. Is there a compelling reason to upsample when there is no precomputed skills training data we need to balance against?

src/instructlab/sdg/datamixing.py
```diff
@@ -615,10 +683,31 @@ def collect(
...
+                    len(skills_phase_data) / self._precomputed_skills_length
+                )
+                if knowledge_to_skills_ratio < MIN_UPSAMPLE_THRESHOLD:
+                    sampling_size = int(self._precomputed_skills_length * 0.03)
+
+                    print(
+                        "\033[93m"
+                        + f"Knowledge detected to be less than {MIN_UPSAMPLE_THRESHOLD*100:.2f}% of skills ({knowledge_to_skills_ratio:.2f}%), upsampling to: {sampling_size}"
+                        + "\033[0m"
+                    )
```
bbrowning, line 704: Please swap this `print(...)` out for a log message, since we use `logger` elsewhere.

bbrowning (approved):
This looks like a very reasonable solution to the upscaling problem in such a short time. I actually don't think it's quite as hacky as the code comments imply, but agree it's not an ideal solution to this general problem.

I'm running the full data generation pipeline against a sample taxonomy with a skill leaf node, a knowledge leaf node, and with a precomputed skills dataset getting mixed in via a customized default skills recipe (ie using https://github.com/instructlab/sdg/blob/main/docs/data_mixing.md#using-instructlab-community-pre-generated-dataset). However, I don't think this will finish on my available hardware before I head out for the night. If for some reason it errors out overnight because of these changes, I'll leave a note tomorrow.

Other than the one nit about replacing the stdout print with a logger, this looks ready to go. Since I'll be scarce tomorrow, going ahead and giving this one approval. 

Thanks for the detailed PR, taking a couple of iterations on this to make it far less hacky than originally proposed, and the attention to detail with code comments and type hints!

src/instructlab/sdg/datamixing.py
RobotSail: Will do!

RobotSail: Done

bbrowning: Thanks!

khaledsulayman (approved):
This is a good approach, thanks for working to get this in!

PR #398, opened 2024-11-18
refactor: generated_data as list

makelinux:
Remove redundant conversions from None to empty list.

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -409,7 +409,7 @@ def generate_data(
             "Synthesizing new instructions. If you aren't satisfied with the generated instructions, interrupt training (Ctrl-C) and try adjusting your YAML files. Adding more examples may help."
         )
 
-    generated_data = None
+    generated_data = []
     empty_sdg_leaf_nodes = []
     for leaf_node in leaf_nodes.values():
         is_knowledge = False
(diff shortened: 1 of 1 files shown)


src/instructlab/sdg/generate_data.py
```diff
@@ -444,11 +444,8 @@ def generate_data(
             empty_sdg_leaf_nodes.append(leaf_node_path)
             logger.warning("Empty dataset for qna node: %s", leaf_node_path)
             continue
-        generated_data = (
-            [new_generated_data]
-            if generated_data is None
-            else generated_data + [new_generated_data]
-        )
+        generated_data.append(new_generated_data)
```
bbrowning, line 447: So, `new_generated_data` here is a list of Datasets which I think means we should be doing `generated_data.extend(new_generated_data)` instead of `.append`?

bbrowning (changes requested):
The tests passed here but I believe it's making a list of lists of Datasets instead of just a list of Datasets. Or it could be that I'm reading the code wrong? However, it's concerning that none of our tests caught this because an extra level of list nesting would surely mess something up.

src/instructlab/sdg/generate_data.py
bbrowning: Actually, looks like it was me who was reading this wrong, I think? marking this comment as resolved and will take another look as I'm attempting to track down some other bugs in this area.

bbrowning (approved):

bbrowning:
I need a different bug near this code and don't want to cause a rebase headache for you, so going to go ahead and merge this PR. Thanks for the cleanup help!

Issue #307, opened 2024-10-14
SDG agnetic pipeline documentation


relyt0925:
We have been receiving questions on the core SDG algorithm for the agnetic pipeline and curious if there would be interest in us producing further documentation on high level charateristics of the algorithm in either a FAQ or "deep dive" page?

Some of the questions we have gotten are below (along with responses)

Assumption:
So, suppose my knowledge taxonomy nodes has a yaml file with:

5 contexts
3 q/a pairs for each context, for a total of 15 q/a pairs
A number of .md files. Say the total size of .md files is 1MB.

Questions:

How many chunks are generated from 1MB of data? Approximate number
Are q/a pairs generated from every chunk, or only a subset of chunks ?
For a chunk, how many questions are generated? 30 questions (given the scale factor of 30) ?

I understand these would be estimates due to the intricacies of the scoring and filtering algorithms but I think it could be useful to have a documentation page that talks about the algorithm at a high level. Thoughts?

relyt0925:
We could help drive this the general questions would be:

Where would we want this document to live?

What would we want the structure to be? FAQ? High level algorithmic walkthrough? We could take a pass at it and then iterate on the proper format we would want.

relyt0925:
Also my responses in case someone wants to correct them:

For the MD file case: the estimation on the number of chunks is 1MB * 1048576 bytes / 4 bytes per token / 1.3 tokens per word / 1000 word chunks ~= 202 chunks

SDG is generated per chunk/qa pair. So in this case that's: 15 * 202 chunks * 30 samples per generation = 90900 samples

Note: actual value will be lower when scoring on quality of output is performed/duplicates removed

aakankshaduggal:
Thanks for bringing this to our attention @relyt0925 ! You're correct in pointing out that making precise predictions about the output size is tricky due to the non-deterministic nature of LLMs. While the estimates provided (such as 202 chunks and 90,900 samples) offer a useful ballpark, the final output size will ultimately depend on various factors, including the document size, scoring, filtering, and quality checks we apply during the process.

aakankshaduggal:
The document can live here in the SDG repository. It would be useful to provide documentation that explains the high-level characteristics of the SDG algorithm. However, I’m not entirely sure how best to explain the factors contributing to the variability in the output.

relyt0925:
@aakankshaduggal we can help take a stab at this and we will tag you for the review and you can critique ask for whatever you like out of it!

[...]

PR #408, opened 2024-11-25
Add FAQ.md to provide sample estimation guidelines

relyt0925:
The new FAQ.md file includes detailed explanations and examples on how to estimate the number of synthetic samples produced at various stages of the SDG training process. This addition aims to enhance user understanding of the sample generation methodology.

I believe it's an MVP to resolving: https://github.com/instructlab/sdg/issues/307

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -24,23 +24,23 @@ For example: let’s say the total size of the knowledge markdown files in the k
 
 Note that this is just an estimation technique: the exact number is non-deterministic since the results are ultimately evaluated against a judge model and results that do not pass a score metric are dropped from the overall set.
 
-## How many seed_examples can Instructlab process in a knowledge leaf node
+## How many seed_examples can InstructLab process in a knowledge leaf node
 
-There is no known limit to the number of seed example entries for a knowledge leaf node. There must be a minimum of 5 seed examples. These parameters can be seen in the taxonomy schema repo:  <https://github.com/instructlab/schema/blob/main/src/instructlab/schema/v3/knowledge.json>
+There is no known limit to the number of seed example entries for a knowledge leaf node. There must be a minimum of 5 seed examples. These parameters can be seen in the taxonomy [knowledge schema](https://github.com/instructlab/schema/blob/main/src/instructlab/schema/v3/knowledge.json).
 
 ## How many qna pairs can be listed for a given seed_example in a knowledge leaf node
 
-**Exactly** 3 QNA pairs must be listed for a given seed_example. If more is specified they will be ignored and not processed by the appropriate prompt. This can be seen by looking at the prompt files in SDG: <https://github.com/instructlab/sdg/blob/v0.6.2/src/instructlab/sdg/configs/knowledge/simple_generate_qa.yaml#L21-L28>
+**Exactly** 3 QNA pairs must be listed for a given seed_example. If more is specified they will be ignored and not processed by the appropriate prompt. This can be seen by looking at the [prompt files in SDG](/src/instructlab/sdg/configs/knowledge/simple_generate_qa.yaml#L21-L28).
 
 ## How long can a given seed_example be for a knowledge leaf node?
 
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -0,0 +1,23 @@
...
+The [Large-Scale Alignment for ChatBots](https://arxiv.org/pdf/2403.01081) research paper does an excellent job of explaining the SDG process in section 3.2.
+
+## How do I estimate the amount of samples a given leaf node will produce in the SDG process for training?
+
+For each question and answer pair in the taxonomy leaf node in a skill file: an estimated 30 synthetic samples will be produced in the training dataset.  
+
+For each knowledge leaf node: the formula to estimate the number of produced synthetic samples in the training dataset is:
+
+```text
+(total cumulative size of knowledge documents / max document chunk size) * number of qna pairs in the knowledge file leaf node * 30 synthetic samples per qna pair
```
RobotSail, line 14: Does this factor in the amount of samples that will be filtered out?

relyt0925: @RobotSail this is an excellent point: it does not since that part is non detemenistic to a certain extent: but folks have been looking to be able to get some guidance on some ball park numbers/ answers to general questions around how the taxonomy is processed through SDG

I should call that out as a disclaimer and will do that

relyt0925:
These are a good initial set of FAQs that we have seen pop up with what I feel (based on my studies of the codebase) are the appropriate answers. More than happy to adjust anything that is inaccurate though based on expert opinion!

docs/FAQ.md
```diff
@@ -0,0 +1,51 @@
...
+
+## How many seed_examples can Instructlab process in a skills leaf node?
+
+There is no known limit to the number of seed example entries for a knowledge leaf node. There must be a minimum of 5 seed examples. These parameters can be seen in the taxonomy schema repo:  <https://github.com/instructlab/schema/blob/main/src/instructlab/schema/v3/compositional_skills.json#L31>
+
+## How long can a given seed_example be for a knowledge leaf node?
```
bbrowning, line 43: I believe there's a typo here and this should be "skills leaf node" vs "knowledge leaf node".

docs/FAQ.md
```diff
@@ -0,0 +1,51 @@
...
+
+Note that this is just an estimation technique: the exact number is non-deterministic since the results are ultimately evaluated against a judge model and results that do not pass a score metric are dropped from the overall set.
+
+## How many seed_examples can Instructlab process in a knowledge leaf node
+
+There is no known limit to the number of seed example entries for a knowledge leaf node. There must be a minimum of 5 seed examples. These parameters can be seen in the taxonomy schema repo:  <https://github.com/instructlab/schema/blob/main/src/instructlab/schema/v3/knowledge.json>
+
+## How many qna pairs can be listed for a given seed_example in a knowledge leaf node
+
+**Exactly** 3 QNA pairs must be listed for a given seed_example. If more is specified they will be ignored and not processed by the appropriate prompt. This can be seen by looking at the prompt files in SDG: <https://github.com/instructlab/sdg/blob/v0.6.2/src/instructlab/sdg/configs/knowledge/simple_generate_qa.yaml#L21-L28>
```
bbrowning, line 33: I'm a bit uncomfortable linking to v0.6.2 here directly, as this doc lives in the main branch. It feels like we do a relative link (can we do that in github markdown?) to the file in the same branch as this doc.

bbrowning (commented):
Thanks for expanding our docs! I have a couple of comments, but also a more general question. Do you think it would be better to document most of these things in the instructlab/instructlab repository instead of directly in SDG? The reason I ask is that the questions touch on taxonomy, SDG, training, and the intersection of all these. And the target users here would likely be using the `ilab` CLI to run these workflows as opposed to SDG directly?

relyt0925:
@bbrowning sorry for delay! Thank you so much for review! I agree with you that maybe there is a better home for this page. I will go ahead and get the comments addressed, we can make sure we are all comfortable with the content: and then we can think on where we want it's home to be!

RobotSail (approved):

bbrowning (approved):
I added in the requested changes (cleaning up links, fixing typo) as an additional commit and will go ahead and approve to get this merged.

relyt0925:
Thank you so much @bbrowning for the assistance! Sorry for the delay hope everything is wonderful with you!

PR #409, opened 2024-11-26
Reconcile core data generation features with latest research advances

bbrowning:
This PR brings in some of the latest advancements prototyped by our research team into the broader codebase for everyone's use. It's a work-in-progress, but also something that others may wish to follow, comment on, and contribute to as the work gets done. There are still outstanding features not yet added to this - some new pipeline block types, an improved skills pipeline config, LLMLogProbBlock and LLMMessagesBlock are just stubs, etc.

And, this may cause the deprecation / removal of some existing functionality. That is not entirely clear yet, but will become apparent as work progresses.

See `docs/upgrading_from_v0.6.x_to_v0.7.x.md` for more details, although that too is still only stubbed out.


bbrowning:
The fact that the e2e-small test passed even though I haven't actually converted any of our default config prompt templates to Jinja syntax yet is concerning, as that means the test is extremely loose in what it considers success. We wouldn't have actually included any of the user's question or answers in the skills or knowledge prompts we sent to the model, and instead it would have all had placeholder python String format tokens in it.

bbrowning:
Ok, I believe this has enough of the research code ported over, new tests created, and existing tests passing that it's ready for review. I know it's a big effort to review this, but it was also a big lift to get the core improvements from research working in our codebase.

BlockRegistry, PromptRegistry, IterBlock, and Jinja prompt templates should all be working. I created #413 and #414 separately to track finishing up `LLMLogProbBlock` and `LLMMessagesBlock`.

This does change some of our public API, although at this time there are no known users of the breaking changes there - basically removal of `ImportBlock` and reorganizing blocks under a `instructlab.sdg.blocks` package. The API used by the InstructLab CLI should be unchanged, and the e2e CI tests still passing without changes confirms this.

jwm4 (approved):
I don't see any problems with any of these changes, but I am not really an expert (and don't have write access, so my approval doesn't count for much).  I would note that the research code base also has a very nice README file with a lot of useful information.  I would like that merged in too with edits as needed to reflect any differences between that code and the open source code.   However, this PR is already plenty big, so I would recommend a separate PR for the README merge.

khaledsulayman (commented):
thanks for this massive lift!

left some minor comments, besides that everything else looks good!

bbrowning:
> I don't see any problems with any of these changes, but I am not really an expert (and don't have write access, so my approval doesn't count for much). I would note that the research code base also has a very nice README file with a lot of useful information. I would like that merged in too with edits as needed to reflect any differences between that code and the open source code. However, this PR is already plenty big, so I would recommend a separate PR for the README merge.

I agree we should pull in those changes - tracked as #428.

src/instructlab/sdg/blocks/block.py
```diff
@@ -0,0 +1,65 @@
...
+        self.block_name = block_name
+
+    def _validate(self, prompt_template: Template, input_dict: Dict[str, Any]) -> bool:
+        """
+        Validate the input data for this block. This method validates whether all required
+        variables in the Jinja template are provided in the input_dict.
+
+        :param prompt_template: The Jinja2 template object.
+        :param input_dict: A dictionary of input values to check against the template.
+        :return: True if the input data is valid (i.e., no missing variables), False otherwise.
```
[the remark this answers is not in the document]
bbrowning: Yes, great catch - I just copied the comments over verbatim, but pushed a new commit to adapt these to the style you outlined that we use elsewhere in SDG.

src/instructlab/sdg/blocks/block.py
```diff
@@ -0,0 +1,65 @@
...
+
+        class Default(dict):
+            def __missing__(self, key: str) -> None:
+                raise KeyError(key)
+
+        try:
+            # Try rendering the template with the input_dict
+            prompt_template.render(ChainMap(input_dict, Default()))
+            return True
+        except UndefinedError as e:
```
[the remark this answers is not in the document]
bbrowning: You made me realize some of this is leftover from pre-Jinja stuff - we can do away with KeyError and ChainMap entirely here, and just rely on Jinja to throw UndefinedError for us on any undefined keys. Simplifying this logic to do that in another commit.

src/instructlab/sdg/blocks/iterblock.py
```diff
@@ -0,0 +1,42 @@
...
+from ..pipeline import _lookup_block_type
+from ..registry import BlockRegistry
+from .block import Block
+
+logger = logging.getLogger(__name__)
+
+
+# This is part of the public API.
+@BlockRegistry.register("IterBlock")
+class IterBlock(Block):
```
aakankshaduggal, line 19: Could we add a docstring to explain what this new block does? It would be helpful to briefly describe its purpose, key attributes like `num_iters`, and how it fits into the overall system. This will make it easier for others to understand its functionality and use. We also don't have to add it in this PR, just a recommendation.

aakankshaduggal (approved):
Thanks @bbrowning for a great PR! LGTM 🚢

src/instructlab/sdg/blocks/iterblock.py
bbrowning: Done!

src/instructlab/sdg/blocks/llmblock.py
```diff
@@ -300,6 +290,13 @@ def __init__(
         parser_kwargs={},
         batch_kwargs={},
     ) -> None:
+        assert (
+            config_paths
+        ), "ConditionalLLMBlock config_paths requires at least one entry"
+        for config_path in config_paths:
+            assert (
+                len(config_path) == 2
+            ), "ConditionalLLMBlock config_paths each entry should be a list of config path and selector column names"
```
anastasds, line 299: What's the motivation to use `assert` here instead of raising?

anastasds (approved):

src/instructlab/sdg/blocks/llmblock.py
bbrowning: We typically only use `assert` outside of tests to validate/document code assumptions that are not meant to be user-facing or influenced by user config or params. However, in this case, I think you're properly catching that these config_paths are in fact user-facing (as they specify them in their pipeline yaml) and thus these should be a raised exception instead. Will fix :smile:

bbrowning: This now raises a `BlockConfigParserError` instead of asserting, and adds a test to validate that gets raised properly. Thanks for the input!

bbrowning:
Rebased on top of latest main since CI is now fixed there - will let CI chew on things and if it's good and no more comments, going to squash a lot of these into a handful of fewer commits, add the co-authorship metadata for the upstream researchers that did much of the initial code, and get this merged. I'm not planning any functional changes to this PR from this point forward unless more review notes come up.

PR #411, opened 2024-11-26
refactor: remove unused generate_data argument

makelinux:
The logger argument is already removed from the cli.


bbrowning:
We're undertaking a refactoring and splitting up of the monolithic `generate_data` method, which will include moving some of its existing functionality into the instructlab/instructlab repository as well as designing new Python APIs we want to support for SDG usage directly. That new API is tracked in #412, and I'm going to close this PR as we focus on tackling the new API and method signature/parameters required instead of making small tweaks to our existing `generate_data` method.

PR #416, opened 2024-11-27
Ensure knowledge docs are cloned into unique dirs

### LINKED ISSUE #404: Sdg v0.6.0+ multiple knowledge sources fails to clone
I think I found a potential race condition specifically here (context aware chunking): https://github.com/instructlab/sdg/pull/284
Basically if there is more than 1 knowledge document for git to clone, and it happens to do multiple clones with the same second it will generate the same output dir:
`document_output_dir = Path(output_dir) / f"documents-{date_suffix}"`
Which causes SDG to fail since the directory already exists on the git clone.

Generating data on a single knowledge document, things works just fine. Its when we get to multiple I am seeing failures.

bbrowning:
Previously, we were attempting to clone multiple knowledge documents into the same destination directory, leading to failures generating data for any run that contained 2+ knowledge leaf nodes.

Now, we clone docs into a guaranteed unique (via `tempfile.mkdtemp`) subdirectory per knowledge leaf node. Just using a subdirectory per leaf node could still have led to collisions if the user ran data generation twice within one minute, which is why this goes the extra step of using `mkdtemp` for guaranteed uniqueness.

Fixes #404

### DIFF
--- a/src/instructlab/sdg/utils/taxonomy.py
+++ b/src/instructlab/sdg/utils/taxonomy.py
@@ -2,6 +2,7 @@
 
 # Standard
 from pathlib import Path
+from tempfile import mkdtemp
 from typing import Dict, List, Tuple, Union
 import glob
 import logging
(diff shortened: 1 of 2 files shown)


aakankshaduggal (approved):
Thanks @bbrowning for the quick fix.

bbrowning:
@Mergifyio backport release-v0.6

bbrowning:
Still waiting on e2e-medium that is taking its sweet time working through some compositional skill testing, unrelated to this change. For the sake of expediency, I'm going to manually push the merge button on this one and get CI starting to run on the backport PR so they can happen simultaneously.

PR #419, opened 2024-11-28
Add disk check after tests run

danmcp:
To help with debugging disk space issues


bbrowning (approved):
Looks good - thanks! This should help us keep track of disk usage to track down some of the out of disk space flakes we've been seeing.

nathan-weinberg (approved):

PR #421, opened 2024-11-29
Mix all the user's generated skills instead of 30 per leaf node

### LINKED ISSUE #420: Change default skills sampling size during datamixing to "1.0" from "30"
We had a hardcoded sampling size of 30 samples per leaf node of custom skills. That's a very specific number that may be appropriate in some cases, but we're seeing more commonly users having only 1 or a few skill nodes and wanting more of their generated skill data in the mixed output. So, change the default sampling size of custom skills to "1.0", which will include all their generated skill data in the mixed output per skill leaf node instead of truncating it to 30 per skill leaf node.

bbrowning:
Users have provided feedback that they expect all of their generated skills to get mixed in the final output dataset by default instead of truncating this to only 30. The choice of 30 was not a great default for every use case anyway, and mostly tailored towards very large taxonomies with large numbers of skills. Most of our users are using fewer numbers of skills, and expect a larger amount of generated skill data to make it into the results to increase the overall number of skill samples in the mixed output that matches their custom skills.

Fixes #420

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -587,14 +587,12 @@ def _convert_to_leaf_node_messages(sample: dict, sys_prompt: str):
 class DataMixer:
     # pylint: disable=too-many-instance-attributes
 
-    # This determines how many samples to pick from each skill when mixing
-    # skill datasets. It's only used for skills, as knowledge may require
-    # a variable number of samples depending on the length of the
-    # knowledge documents in question. The expectation is that this is
-    # enough samples to sufficiently learn a new skill while also ensuring
-    # a balance of overall mixed data when learning multiple skills at
-    # once.
-    NUM_SYNTH_SKILLS = 30
+    # This determines how many samples to pick from each skill when
+    # mixing skill datasets. A value of "1.0" means to mix all
+    # generated skills, while an integer value would be a fixed
+    # upscaling or downscaling to the number of skills mixed per leaf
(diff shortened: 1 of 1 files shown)


aakankshaduggal (commented):
@bbrowning Thanks for the PR. I like the idea of mixing all the samples that are being generated, but we need to control/ensure how much we are generated vs how much we are mixing. Should we continue to default to maybe 30/50 and let the users override given special cases. Having said that, should we make this a parameter that we can expose?

jwm4 (approved):
This seems like an improvement to me.  FWIW, I agree it would be even better to make it a parameter as @aakankshaduggal suggests above.

bbrowning:
I agree that users need the ability to control how many skills get mixed. However, do we want to introduce a new parameter for that here? Or instead tackle that once we expose data mixing directly via the CLI? I'm personally inclined to have everything mix in the default recipes at a sampling size of 1.0 as the starting point, and then show advanced users how to re-mix their mixed dataset using Recipes, which gives them fine-grained control over the sampling size of every individual leaf node in the mixed dataset.

bbrowning:
We had some discussion outside GitHub about this, and perhaps it's not worth doing until we expose knobs to users to control the number of samples generated (which `--sdg-scale-factor` plays a role in, only sometimes) and the number of samples mixed (which has no exposed knob today).

bbrowning:
Closing this for now - we can revisit later if needed.

PR #422, opened 2024-11-29
Move AWS_REGION from using secret to var

danmcp:
Switching from using AWS_REGION as a secret to a var.  There isn't a strong reason to hide the value.  And having the secret exist makes it impossible to pass the value to scripts within workflows.  If the secret exists, the variable's value will still be hidden.


bbrowning (approved):

nathan-weinberg (approved):

bbrowning:
@mergifyio backport release-v0.6

PR #443, opened 2024-12-10
Split up `generate_data` and add a `mix_datasets` top level API

bbrowning:
This separates out `instructlab.sdg.generate_data` into separate functions, where `generate_data` just calls into these separate functions instead of containing all the logic in itself directly.

This enables us to expose a new top-level API, `mix_datasets`, to allow users to mix datasets. It also lays the groundwork for how we might split some of the pre-processing and post-processing out into separate modules or other repositories, as now there's a clean separation between the various concerns that were previously all intertwined within `generate_data`.

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from importlib import resources
 from pathlib import Path
 from typing import Optional
-import dataclasses
+import glob
 import json
 import logging
 import os
@@ -13,13 +13,19 @@ import time
 
 # Third Party
 # instructlab - All of these need to go away (other than sdg) - issue #6
+from datasets import Dataset
 from xdg_base_dirs import xdg_data_dirs, xdg_data_home
 import openai
 import yaml
 
 # First Party
 from instructlab.sdg.blocks.llmblock import DEFAULT_MAX_NUM_TOKENS
-from instructlab.sdg.datamixing import DataMixer, _get_question_hack, _get_response_hack
+from instructlab.sdg.datamixing import (
+    DataMixer,
+    Recipe,
+    _get_question_hack,
+    _get_response_hack,
+)
 from instructlab.sdg.eval_data import generate_eval_task_data, mmlubench_pipe_init
 from instructlab.sdg.pipeline import (
     FULL_PIPELINES_PACKAGE,
@@ -27,8 +33,8 @@ from instructlab.sdg.pipeline import (
     Pipeline,
     PipelineContext,
 )
-from instructlab.sdg.utils import GenerateException, models
-from instructlab.sdg.utils.json import jldump
+from instructlab.sdg.utils import GenerateException
+from instructlab.sdg.utils.json import jldump, jlload
 from instructlab.sdg.utils.taxonomy import (
     leaf_node_to_samples,
     read_taxonomy_leaf_nodes,
@@ -37,6 +43,10 @@ from instructlab.sdg.utils.taxonomy import (
 logger = logging.getLogger(__name__)
 
 _SYS_PROMPT = "I am a Red Hat® Instruct Model, an AI language model developed by Red Hat and IBM Research based on the granite-3.0-8b-base model. My primary role is to serve as a chat assistant."
+
+DEFAULT_CHUNK_WORD_COUNT = 1000
+DEFAULT_TAXONOMY_BASE = "empty"
+DEFAULT_SERVER_CTX_SIZE = 4096
 
 
 def _unescape(s):
@@ -115,20 +125,21 @@ def _gen_train_data(
 
 def _knowledge_seed_example_to_test_data(seed_example, system_prompt):
     res = []
-    for qna in seed_example["questions_and_answers"]:
-        user = qna["question"] + "\n" + seed_example["context"]
-        res.append(
-            {
-                "system": system_prompt,
-                "user": _unescape(user),
-                "assistant": _unescape(qna["answer"]),
-            }
-        )
+    for i in range(3):
+        idx = i + 1
+        user = seed_example[f"icl_query_{idx}"] + "\n" + seed_example["icl_document"]
+        test_sample = {
+            "user": _unescape(user),
+            "assistant": _unescape(seed_example[f"icl_response_{idx}"]),
+        }
+        if system_prompt:
+            test_sample["system"] = system_prompt
+        res.append(test_sample)
     return res
 
 
 def _gen_test_data(
-    leaf_nodes,
+    seed_examples,
     output_file_test,
     system_prompt,
 ):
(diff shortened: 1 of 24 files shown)


bbrowning:
@jwm4 Does something like this help with some of your needs? A top-level API to only do the preprocessing steps but not any data generation?

cdoern:
>This adds a simple CLI usable within SDG itself to do all taxonomy preprocessing via:
python -m instructlab.sdg.cli.taxonomy_to_samples --taxonomy-path /path/to/my/taxonomy/ --output-dir /path/to/my/output/samples --log-level DEBUG

Does this confuse the user story a bit of what we expect to be a "user facing" entry point to InstructLab? or is this CLI just meant to be a sort of dev environment for quick use? If its the later, then this makes sense.

However, I do think it would make sense to instead focus on adding this CLI functionality into an `ilab data process` command. That could be as simple now as adding the proper plumbing to call whatever whatever API is being invoked via the new CLI being introduced here.

With changes like this that expose a new user entry point we risk divergence and decreased functionality in `ilab`.

src/instructlab/sdg/cli/preprocess_taxonomy.py
```diff
@@ -0,0 +1,82 @@
...
+    DEFAULT_TAXONOMY_BASE,
+    preprocess_taxonomy,
+)
+from instructlab.sdg.utils.logging import setup_logger
+
+if __name__ == "__main__":
+    # Standard
+    import argparse
+
+    parser = argparse.ArgumentParser(
```
cdoern, line 19: This, to me, looks like a workaround instead of adding an `ilab data process` command. I understand completely the necessity to expose the pre-processing in some way, but I think introducing a new CLI for this specific goal seems to diverge a bit from the purpose of the sdg repo as opposed to the instructlab repo which is meant to be the orchestrator of these processes.


Would you be amenable to instead adding these CLI changes into instructlab via the existing `ilab data...` click interface? I have had conversations with folks regarding adding this functionality to InstructLab and that seems like a more sustainable solution.

If this is not something you'd be amenable to, could we scope the work to move these CLI interfaces into the existing ilab CLI? The duplication of CLIs as a precedent concerns me for usability.

cdoern (commented):
one comment related to the above

src/instructlab/sdg/cli/preprocess_taxonomy.py
cdoern: Following up. I thought I saw this somewhere and it does seem like as part of the RAG work various data processing commands are being worked on https://github.com/instructlab/instructlab/pull/2736

Maybe we could combine efforts here?

bbrowning: This is mostly an effort by me to understand the API surface we want to expose for SDG, and I think we agree that anything specific to taxonomy, preprocessing, etc should get moved out of SDG. I'm not sure a CLI will even stay in this PR once it gets out of draft, but I am using it for now to test a few things. Once I add some tests that exercise the newly split out SDG API we can probably ditch this specific CLI within SDG itself.

With that said, there will be some CLI(s) that we will want to expose for SDG users that are only using SDG and not the rest of InstructLab. However, those CLI(s) won't have anything to do with taxonomy and will instead focus only on data generation - things like the ability to directly run an SDG pipeline from a set of input file(s) to generate data as output file(s) without any InstructLab concepts like taxonomies.

bbrowning: I pulled out the CLI as I was using it for testing and replaced that instead with an actual test that exercises these APIs. So, resolving this concern.

bbrowning:
I believe this is ready for larger review. Some particular things I'd like to call out for reviewers to think about:

The only other top-level API we expose here via `__init__.py` is `mix_datasets`. Does the Python API of `mix_datasets` look reasonable? This is one we'd want to support for advanced users that need to adjust how generated data gets mixed.

The old monolithic `generate_data` was split into `preprocess_taxonomy`, `generate_taxonomy`, `generate_taxonomy_eval`, `postprocess_taxonomy`, and `mix_datasets`. Are these a reasonable level of granularity, at least in a first pass of splitting things up? For backwards-compatibility we retained a `generate_data` method that just calls all of the above in turn to keep the same behavior as before. None of these except `mix_datasets` are exposed as top-level importable functions from instructlab-sdg. So, we can iterate on naming and further splitting as needed in future releases unless there's strong disagreement here.

We now write more intermediate files to disk between stages, including additional metadata (ie columns) in these jsonl files to keep track of things like whether a sample generated from the taxonomy was a knowledge, freeform, or grounded skill as we use that to differentiate which pipeline to run. Are the `leaf_node_type` and `leaf_node_path` column names reasonable to write into our sample files for this purpose?

Lastly, we now do all the preprocessing for all leaf nodes before moving on to generation. We do all generation for all leaf nodes before moving on to postprocessing. Previously, we operated on a single leaf node at a time end-to-end. In other words, previously we'd preprocess, generate, and postprocess leaf node A. Then we'd do the same for leaf node B. Now, we preprocess leaf nodes A & B, then generate for A & B, and finally postprocess A & B. This was required for decoupling things, and sets us up for some future scenarios where we'll need to entirely finish one stage of our processing before moving on to the next because some stages may use different models, such as subset selection with an embedding model. Any concerns about moving to this approach?

Obviously I'd love feedback for any/all parts of the PR, but just calling out the above as some places to spend a bit more focused time during a review.

src/instructlab/sdg/generate_data.py
```diff
@@ -115,20 +125,21 @@ def _gen_train_data(
...
-    for qna in seed_example["questions_and_answers"]:
-        user = qna["question"] + "\n" + seed_example["context"]
-        res.append(
-            {
-                "system": system_prompt,
-                "user": _unescape(user),
-                "assistant": _unescape(qna["answer"]),
-            }
-        )
+    for i in range(3):
```
aakankshaduggal, line 128: Should we also have this either documented or parameterize the number 3 as to why we are sticking to this value?
```suggestion
    for i in range(num_iterations):
```

src/instructlab/sdg/generate_data.py
```diff
@@ -115,20 +125,21 @@ def _gen_train_data(
 
 def _knowledge_seed_example_to_test_data(seed_example, system_prompt):
```
aakankshaduggal, line 126:
```suggestion
def _knowledge_seed_example_to_test_data(seed_example, system_prompt, num_iterations=3):
```

aakankshaduggal (approved):
Overall lgtm, great work here @bbrowning 
I have minor comment, but we don't have to get it in with this PR.

bbrowning:
I like the suggestion @aakankshaduggal , so incorporating that directly here. Thanks!

bbrowning:
Rebased this to fix a merge conflict, added a few examples of using the mixing API, along with a test to ensure those examples work as expected.

src/instructlab/sdg/generate_data.py
```diff
@@ -368,97 +429,301 @@ def generate_data(
...
+    if recipe.datasets:
+        recipe.save_mixed_dataset(output_file, num_proc)
+    else:
+        logger.info("Not mixing empty recipe file: %s", recipe_file)
+
+
+# This is part of the public API, and used by instructlab.
+# TODO - parameter removal needs to be done in sync with a CLI change.
+# to be removed: logger
+def generate_data(
```
cdoern, line 617: the only concern from the core side of things is that this contract is the same as it was - which it appears to be. Thanks!

cdoern (approved):
approving from a CLI point of view. We can iterate on the public APIs and when they should move to core.

PR #445, opened 2024-12-10
Fix logging the number of generated samples for each leaf node

### LINKED ISSUE #227: `ilab data generate` does not specify the correct num of samples generated
when running through the cli e2e workflow, the output of the 'ilab data generate' cli command said only 2 samples were generated when more than 2 were generated. Should account for the correct number of generated samples every time.

bbrowning:
We were logging the length of our `generated_data` list instead of the number of newly generated samples in each leaf node, causing confusion as this is typically 1, 2, 3, etc instead of 50, 500, 1000, etc that users would expect.

Fixes #227

### DIFF
--- a/src/instructlab/sdg/generate_data.py
+++ b/src/instructlab/sdg/generate_data.py
@@ -426,8 +426,8 @@ def generate_data(
             continue
         generated_data.append(new_generated_data)
 
-        logger.info("Generated %d samples", len(generated_data))
-        logger.debug("Generated data: %s", generated_data)
+        logger.info("Generated %d samples", len(new_generated_data))
+        logger.debug("Generated data: %s", new_generated_data)
 
         if is_knowledge:
             # generate mmlubench data for the current leaf node


bbrowning:
CI runs are showing this is fixed with this PR - seeing log messages such like this which have the intended output:
```
2024-12-10T23:40:10.3629160Z INFO 2024-12-10 23:40:10,329 instructlab.sdg.generate_data:429: Generated 149 samples
2024-12-10T23:40:10.3629885Z DEBUG 2024-12-10 23:40:10,329 instructlab.sdg.generate_data:430: Generated data: Dataset({
2024-12-10T23:40:10.3630610Z     features: ['task_description', 'seed_question', 'seed_response', 'output'],
2024-12-10T23:40:10.3631105Z     num_rows: 149
2024-12-10T23:40:10.3631335Z })
```

jaideepr97 (commented):
@bbrowning should we also emit the leaf node that this number corresponds to? so users actually know what the leaf node is and how many sampels were generated for it

bbrowning:
Yes, that's a good idea - will adjust that, assuming this PR is not obviated by another in-progress on to split out generate_data.

aakankshaduggal (approved):

bbrowning:
Actually going to close this PR, as it is not obsolete since the splitting out of generate_data into multiple pieces merged. That work implicitly fixed this logging issue as part of the refactoring.

PR #449, opened 2024-12-11
allow generate_data logger parameter to overwrite locally defined loggers

khaledsulayman:
Signed-off-by: Khaled Sulayman <ksulayma@redhat.com>


cdoern:
oh I like this version even more, so you only need to override the logger in generate_data, and everywhere else you just remove `__name__` and use the root logger?

khaledsulayman:
we found the issue that was not causing the proposed logging configs to propagate down so this change is not needed

PR #466, opened 2025-01-08
Add simple and full knowledge pipeline functional tests

bbrowning:
This is a port of the old `scripts/test_knowledge.py` into functional tests that we can run with CI. These tests need to run on a GPU, so are marked as 'gpu' in pytest and only execute with a new py3-functional-gpu tox environment. This also adds a new workflow file to execute these tests on a GPU runner.


bbrowning:
Ok, almost ready to move this out of draft state. For now this only tests our end-to-end SDG simple and full pipelines for knowledge, but will be expanded in the future to cover skills, PDF knowledge docs, etc. The main focus here is to stub in an SDG CI job that runs with GPUs but does not have instructlab/instructlab installed, so that we can start working on a test suite for SDG independent of running InstructLab's e2e tests. We have SDG features to test that won't be exposed via the `ilab` CLI, and the circular dependency where the core repo depends on SDG but SDG depends on the core repo for its e2e tests makes it harder than it should be to do some types of changes to SDG.

Tagging in @nathan-weinberg and @courtneypacheco as my go-to CI experts. If you have any general feedback on direction or changes to the approach I've taken here to add a new CI job for SDG, I'd love to hear it.

To move this out of draft state I'll want to remove the temporary `pull_request` workflow trigger added at the top of the workflow file. But, leaving it in there for now in case I need to make any additional changes so that the new CI workflow will run again after I make those changes.

bbrowning:
Marking this as draft while I use this to track down a fix for #482, since I was originally able to reproduce that problem here.

bbrowning:
Ok, all tests pass and removed the temporary dispatch conditions that allowed me to test and trigger this before merging. It's ready for review and should be setup for proper dispatch workflow post-merging now.

aakankshaduggal (approved):

PR #468, opened 2025-01-09
fix: typo in mergify configuration

### LINKED ISSUE #467: Changes only to requirements.txt don't get auto-merged with 2 approvals
In #465 we discovered that a change only to requirements.txt does not get automatically merged by mergify after two approvals. We need to adjust the mergify config to handle this case.

nathan-weinberg:
Fixes #467

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -35,7 +35,7 @@ pull_request_rules:
         - or:
           - files~=\.py$
           - files=pyproject.toml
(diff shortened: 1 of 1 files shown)


bbrowning (approved):
Ahh, so that's what it was. Nice catch!

aakankshaduggal (approved):
Thanks @nathan-weinberg

PR #475, opened 2025-01-10
chore: Change default temporary write directory in all e2e CI jobs from `tmpfs` to `/home/tmp`

### LINKED ISSUE #442: e2e tests occasionally flake with out of disk space
For whatever reason, we're occasionally running out of disk space in our e2e tests, on a node with 700GB+ free before we start the e2e test. See https://github.com/instructlab/sdg/actions/runs/12261506100/job/34208780366 for an example here.

courtneypacheco:
This PR tackles the same issue originally identified in https://github.com/instructlab/instructlab/issues/2866, where we noticed intermittent OOM issues during some of our e2e CI jobs.

While I have not noticed any OOM issues recently in this repo, updating the default temporary write directory from `tmpfs` (aka `/tmp`) to `/home/tmp` should mitigate the risk of OOM errors occurring in our e2e CI jobs since we'll now theoretically have access to 100% of the available physical memory in our EBS volume, instead of only 50%.

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -26,6 +26,7 @@ concurrency:
 
 env:
   LC_ALL: en_US.UTF-8
+  TMPDIR: /home/tmp
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


nathan-weinberg (approved):

bbrowning (approved):
Looks great - thanks! This should fix #442 from this repo as well, right?

courtneypacheco:
@bbrowning yep, it should in theory!

i think we should leave that issue open for maybe ~2 weeks to see if the issue has been resolved by this PR, or if it still occurs.

PR #476, opened 2025-01-10
RHELAI-2756 Adding Adaptive Throttling

### LINKED ISSUE #424: Control maximum parallel requests to inference server
We've observed users overwhelming their inference server, primarily when running data generation in batches against vLLM. What happens is we end up piling on thousands of pending requests into vLLM, causing inference timeouts as we exhaust its KV cache and overwhelm the ability of the teacher model inference to keep up.

Today we really just expose a `num-cpus` parameter that controls how many batches we process in parallel but we don't actually do anything to control how many in-flight requests we send to the inference server. Instead, we should expose a knob that controls the maximum number of parallel requests we'll send to the inference server at a time, and adjust our request handling so that we respect that amount of concurrency instead of just firing thousands of requests at a back-up inference server.

So, the proposal here is to deprecate the `num-cpus` parameter and add a new one - name to be determined, but something like `max-parallel-requests`. And, centralize dispatching of requests to the inference server so we can have a client-side queue of pending requests where we pipeline these to the server based on the configured `max-parallel-requests`. This would implicitly apply backpressure as we wouldn't have more than N requests in-flight, and also give an easy knob for someone that's hosting teacher models across multiple machines behind a load balancer to increase the parallel requests we send across their distributed inference.

eshwarprasadS:
### Why Adaptive Throttling?
Fixes #424 
### The Problem:
If too many requests are sent concurrently to the server, it might:
- Run out of GPU memory (VRAM) or other resources.
- Fail to respond to requests within the timeout period.
- Reducing concurrency manually (batch_num_workers) is a workaround, but it's static and not adaptive to server conditions.

### The Solution:
- Monitor the server's behavior (e.g., response success or failure).
- Dynamically increase or decrease concurrency based on server feedback.

### The AdaptiveThrottler Class

This class encapsulates the logic for adjusting concurrency:

Parameters:
```
        min_workers: Minimum number of concurrent workers (never go below this).
        max_workers: Maximum allowed concurrent workers, matches the num_cpus from instructlab core.
        initial_workers: Starting number of workers, 50% of max to begin with.
        tolerance: Fraction to reduce workers when errors occur.
```

### DIFF
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -178,11 +178,11 @@ class Pipeline:
             max_workers=self.ctx.batch_num_workers, # Upper limit from config
             initial_workers=self.ctx.batch_num_workers//2, # Start at 50% of max
         )
-            
+
         if not input_splits:
             logger.warning("Input splits are empty. Returning empty dataset.")
-            return concatenate_datasets([])            
-        
+            return concatenate_datasets([])
+
         while input_splits:
             # Get the current number of workers from the throttler
             current_workers = throttler.get_workers()
@@ -205,7 +205,7 @@ class Pipeline:
                         output_splits.append(ds) # Store the successful result
                         checkpointer.checkpoint(ds) # Save progress
                         throttler.adjust_workers(success=True) # Increase workers on success
-                    
+
                     except PipelineBlockError as err:
                         root_exception = err.exception  # Access the underlying exception
 
@@ -217,11 +217,8 @@ class Pipeline:
                             # Non-retryable errors
                             logger.error("Non Retryable error in pipeline batch generation: %s", err)
                             throttler.adjust_workers(success=False)
-                    except Exception as generic_err:
-                        logger.error("Unexpected error in batch generation: %s", generic_err)
-                        throttler.adjust_workers(success=False)  # Adjust workers for unexpected errors
 
-        
+
         checkpointer.done()
         if pre_generated_data:
             output_splits.append(pre_generated_data)
--- a/src/instructlab/sdg/pipeline.py
+++ b/src/instructlab/sdg/pipeline.py
@@ -8,10 +8,12 @@ from typing import Dict, Iterable, List, Optional
 import logging
 import math
 import os.path
+import requests
 
 # Third Party
 from datasets import Dataset, concatenate_datasets
 from openai import OpenAI
+from openai import OpenAIError
 import yaml
 
 # First Party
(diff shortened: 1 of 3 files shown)


bbrowning:
I pulled this locally and attempted to verify this fix. What I did was run a Mock OpenAI server locally (using https://github.com/polly3d/mockai with a response delay set to 30 seconds) and then edited my generate_data.py to set `client.timeout = 5.0` early in the method, to give a 5 second timeout to our OpenAI client.

Running `ilab data generate` after that, I get:
```
WARNING 2025-01-14 09:33:31,371 instructlab.sdg.pipeline:214: Retryable error in pipeline batch generation: Connection error.
failed to generate data with exception: Unable to concatenate an empty list of datasets.
```

So, I'm hitting the new logging warning about Retryable errors. And, that should be adjusting the throttler down. But, it doesn't appear to actually be retrying that failed request that timed out?

eshwarprasadS:
> I pulled this locally and attempted to verify this fix. What I did was run a Mock OpenAI server locally (using https://github.com/polly3d/mockai with a response delay set to 30 seconds) and then edited my generate_data.py to set `client.timeout = 5.0` early in the method, to give a 5 second timeout to our OpenAI client.
> 
> Running `ilab data generate` after that, I get:
> 
> ```
> WARNING 2025-01-14 09:33:31,371 instructlab.sdg.pipeline:214: Retryable error in pipeline batch generation: Connection error.
> failed to generate data with exception: Unable to concatenate an empty list of datasets.
> ```
> 
> So, I'm hitting the new logging warning about Retryable errors. And, that should be adjusting the throttler down. But, it doesn't appear to actually be retrying that failed request that timed out?

Hmm. Good question @bbrowning. From my understanding, the logs that you got indicate that this was activated: 

                        `if isinstance(root_exception, (requests.exceptions.RequestException, TimeoutError, OpenAIError)):
                            # Retryable errors
                            logger.warning("Retryable error in pipeline batch generation: %s", root_exception)
                            throttler.adjust_workers(success=False)`

From our collective understanding, OpenAIError is potentially thrown back to us only after the retries have already been attempted (correct me if I am wrong). In that case, I think the logging message could be clearer, but the behavior is as expected(?): the thread of request that was sent through OpenAI client was retried a couple times, with backoff, and failed still, and gracefully died without succeeding, raising its OpenAIError (I am assuming), and we caught and reported it as such.

This understanding of mine of the behavior was from the Perf. team who said: 

>My understanding of the code is that if we have num-cpus greater than 1 and if one of the thread hits the timeout or any other errors it would just gracefully shutdown that thread without affecting the data generation in other threads before finally doing the data mixing.


Did your generate 'fail' entirely at the end, or were there these warnings but the generation completed successfully at the end?

bbrowning:
My generation did fail entirely at the end, because it had zero generated output as the requests that timed out were ultimately "lost", in that after we received a timeout those requests were dropped and we attempted no further data generation from that input sample.

bbrowning:
Do we want to leave this PR open for now? Or do we want to close it with the expectation that #484 fixes the underlying issue we observed?

eshwarprasadS:
> Do we want to leave this PR open for now? Or do we want to close it with the expectation that #484 fixes the underlying issue we observed?

Yes, lets close this, since #484 addresses the issue already.

Issue #482, opened 2025-01-16
E2E tests in PRs are not testing against the same codebase as other tests


bbrowning:
Our e2e tests, that use separate runners and manually clone the SDG repo, are not testing against the same codebase as the regular GitHub actions like our unit and functional tests.

If you search for `merge_branch` at https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs, you'll see that when testing on pull requests, GitHub clones the merge branch of the PR as opposed to the head of the PR. The merge branch represents the state the codebase would be in if that PR merged, as opposed to testing the state the codebase is in at the head of the PR. By testing the merge branch, it means any changes that have happened to the main branch since the PR was opened get merged and considered during the test. But, because our E2Es only test the head of the PR, they are not picking up any changes that have happened to main (or whatever target branch) since the PR was opened, giving a different signal and potentially different / confusing errors compared to the unit and functional tests.

Here's an example of how GitHub clones the merge branch from a recent unit test:
```
/usr/bin/git -c protocol.version=2 fetch --prune --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/* +fc26293dc623f3f33e27a7ce7b8832b97fb36a0b:refs/remotes/pull/466/merge
...
/usr/bin/git checkout --progress --force refs/remotes/pull/466/merge
```

Compare that to how we clone during the e2e tests:
```
git fetch origin pull/466/head:pr-466
git checkout pr-466
From https://github.com/instructlab/sdg
* [new ref]         refs/pull/466/head -> pr-466
Switched to branch 'pr-466'
```

The difference you'll see is `refs/pull/466/merge` vs `refs/pull/466/head`.

This was discovered while testing #466 , and the temporary workaround is to always rebase a PR to the very latest target branch to ensure the unit and e2e tests are running against the same codebase.

PR #483, opened 2025-01-16
Clone the PR's merge branch vs head in e2e tests

bbrowning:
This configures actions/checkout to clone the PR's merge branch (instead of the PR's head), to align with how unit/functional tests work and ensure that all of our tests are testing against the same codebase - the state of the code as it would look after the PR merges, to get an accurate picture of whether things will pass if this merges or not.

Related to #482

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -96,7 +96,6 @@ jobs:
       - name: Checkout instructlab/sdg
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          repository: "instructlab/sdg"
           path: "sdg"
           # https://github.com/actions/checkout/issues/249
           fetch-depth: 0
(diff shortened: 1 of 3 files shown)


bbrowning:
So, this might fix 482 but it's one of those things we won't know until we merge it and see what happens. I tested this manually in a separate PR and it fixed things, but testing these changes manually involves changing the trigger conditions for the workflows and doing that can influence the exact ref checked out by actions/checkout. This definitely fixes the issue for anything use `pull_request` dispatching, but for `pull_request_target` like we use I'm not 100% sure if that will also use the PR's merge branch or instead only use the base branch when cloning the repo in actions/checkout.

bbrowning:
After spelunking in the GitHub actions/checkout repo, it appears that when using `pull_request_target` then actions/checkout will by default checkout the base branch and NOT the pr merge commit. The way I was manually testing this was not using `pull_request_target` because that can only be tested once changes have merged. So, I'm going to change the code to fix this a different way, by going back to the explicit checkout we had but pointing it to the PR merge. That should allow me to test this in a `pull_request` dispatch somewhere and have confidence that it will work in a `pull_request_target` dispatch.

bbrowning:
Ok, this is ready for review. The tldr of this change is that instead of checking out `pull/123/head` we now checkout `pull/123/merge` which is a synthetic merge commit created by GitHub server-side that represents the state of the code if the PR was merged into the target branch. This matches the code our unit tests run against, and prevents weird situations I ran into where unit tests pass but e2e tests fail in an unexpected way because one is testing against the state the repo would be in after the PR merges (PR's merge branch) and one is testing against the state of the repo when the PR was opened (PR's head branch). The linked issue #482 has a bit more details for the underlying issue itself.

Note that the default behavior of `actions/checkout` is to checkout the merge branch for `pull_request` events, but for `pull_request_target` events its default behavior is to checkout the base branch (with no code from the PR at all). You'll see I removed the repository reference from our `actions/checkout` calls when referring to the same repository so that its default behavior applies and we don't get any unexpected surprises in what we checkout if we end up swapping our event from `pull_request_target` to `pull_request` at some point.

aakankshaduggal (approved):

Issue #353, opened 2024-11-08
Enable support for docling V2 supported filetypes


khaledsulayman:
Once #334 and #350 are completed, we'll have migrated to the hierarchical chunker and have a list of filetypes we support defined somewhere in code.

Docling V2 adds support for html, docx, and pptx, so we should theoretically be able to support those on our end by adding them to our supported filetypes.

khaledsulayman:
this might not be blocked by #350 after all, since the newer docling filetypes should still work with the legacy format that we currently consume.

Issue #350, opened 2024-11-08
Use Docling v2 hierarchical chunking instead of the existing context-aware chunking implementation


jwm4:
## Context

The following was done recently:

- Context aware chunking for PDFs (only) was delivered with a dependency on Docling v1.x: https://github.com/instructlab/sdg/issues/271
- Also delivered was upgrading to Docling v2 but still using Docling v1.x format by accessing doc.legacy_document.  This is still also doing context-aware chunking for PDFs only.  https://github.com/instructlab/sdg/issues/333

This allows us to ensure that we're using Docling v2 and get all the latest quality improvements and we were able to do so without breaking the existing code.  However, the solution winds up being suboptimal because it is not making full use of the power of Docling v2.  Specifically, by getting all the Docling outputs from the Docling v1.x format via doc.legacy_document, it misses out on the opportunity to benefit from enhancements to the Docling v2 format.  Furthermore, the context-aware chunker is confusing, idiosyncratic, and there are hardly any comments.  It doesn't seem to handle some obvious edge cases (e.g., long document elements that don't fit into a single chunk) and it doesn't have comments that explain why it is not doing so.

Much of the work needed to do context-aware chunking is already handled by the Docling v2 hierarchical chunker.  However, the Docling v2 hierarchical chunker does not constrain the size of the chunks it produces; often the chunks are very small (much smaller than we would want for most practical uses such as SDG) and sometimes they are very large (and thus unusable for SDG without splitting or truncating).  For those reasons, we need to post-process the outputs of the Docling v2 hierarchical chunker to split up large chunks and/or merge small chunks as needed.

There is a [discussion in the docling repo](https://github.com/DS4SD/docling/discussions/191) about how to do this well, and I have proposed a [candidate implementation](https://github.com/DS4SD/docling/pull/193) that I reference in the discussion.  Here is a basic overview of how it works:

1. First it goes through all the chunks (from the Docling hierarchical chunker) that contain multiple doc_items (e.g., itemized lists) and checks to see if they are too long for the context window.  If they are, it splits them into smaller chunks along the doc_item boundaries.  For example, if it gets a list with 5 elements each with 200 tokens and the context window is 500 tokens, it will put the first two list elements into one chunk, the next two into a second chunk, and the last one in a third chunk.  At this stage, if some doc_item is too long for the context window, it becomes its own chunks.
[...]

PR #485, opened 2025-01-16
Remove the legacy document format to move to docling v2 output

aakankshaduggal:
#350 #353


aakankshaduggal:
Closing because we are taking an alternate approach as defined #503

PR #492, opened 2025-01-21
Stub in a run_pipeline CLI and add example usage

bbrowning:
Part of the typical research flow with SDG is to run a single pipeline, with an input dataset and an output dataset. We need a simple way to enable this within the SDG repo, ideally without requiring code to be written just for the cases of running a Pipeline. This stubs in a simple run_pipeline CLI for just that purpose.

This also adds an example of how to use the pipeline by documenting the IterBlock where we show how to use that block with the new run_pipeline CLI.


bbrowning:
This is blocking adding additional examples and tests to our repo that will depend on this CLI and the test setup created here, such as adding an annotation pipeline example to demonstrate how SDG can be used for other types of things besides pure data generation.

aakankshaduggal (approved):
Thanks @bbrowning 
LGTM

khaledsulayman (approved):

PR #493, opened 2025-01-21
use render method for jinja template, add unit tests

### LINKED ISSUE #490: ConditionalLLMBlock not working with Jinja templates
We had a user discover that during the Jinja template transition we missed updating `ConditionalLLMBlock` to use Jinja templates. So, we need to fix that as well as plug the hole in our test coverage to ensure ConditionalLLMBlock works as well. What happens is we have Jinja Template objects in the code, but are still treating them as the old style python template strings instead of calling the proper Jinja template methods to render the templates.

The commit at https://github.com/instructlab/sdg/commit/c7066c3596aac81c0ad221b981588b211821c401 appears to fix this in manual testing, but we need to add a test case that fails and then passes with that change to ensure this is fixed.

eshwarprasadS:
Fixes #490 

### Problem

We changed from using simple python string templates to jinja templates, but missed a few places (`ConditionalLLMBlock`) to account for this change, which went unnoticed due to missing test coverage for this.

### The Solution

- Addresses the problem (change from `.format()` to `.render()`, treating the input to `_format_prompt()` as a jinja `Template`
- Adds two unit tests `test_format_prompt_with_jinja_templates` and `test_format_prompt_without_selector` that test the behavior of `Template.render()` based on the inputs. These tests will fail if there are parts of the code that call to `_format_prompt()` with a python string argument.

### DIFF
--- a/src/instructlab/sdg/blocks/llmblock.py
+++ b/src/instructlab/sdg/blocks/llmblock.py
@@ -323,14 +323,15 @@ class ConditionalLLMBlock(LLMBlock):
                 )
 
     def _format_prompt(self, sample: Dict) -> str:
+        # If prompt_template is a dict, use the selector column to select the correct template
         if isinstance(self.prompt_template, dict):
             return (
                 self.prompt_template[sample[self.selector_column_name]]
-                .format(**sample)
+                .render(sample)
                 .strip()
             )
-
-        return self.prompt_template.format(**sample).strip()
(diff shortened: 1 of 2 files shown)


bbrowning (approved):
The added code comments help clarify what `_format_prompt` is doing, and I appreciate you testing all branches of the logic with the multiple test cases. Thanks for the thorough and quick fix here!

PR #512, opened 2025-01-27
ci: Add OpenAI keys into CI

alimaredia:


bbrowning:
None of SDG's CI tests should be using a real OpenAI key. What's the reason we want to add this?

nathan-weinberg (approved):

nathan-weinberg:
> None of SDG's CI tests should be using a real OpenAI key. What's the reason we want to add this?

This is for the E2Es, which require this field: https://github.com/instructlab/sdg/actions/runs/12993140844

bbrowning:
Ahh, ok. I'll approve this for now, but will also work towards actively removing the instructlab e2e tests from running in the SDG repo at all to remove the need for things like this.

bbrowning (approved):

bbrowning:
Merged manually to get CI flowing again

PR #518, opened 2025-01-28
Make everything but client optional on PipelineContext

### LINKED ISSUE #511: Make `model_id` and `model_family` in `PipelineContext` optional
The configuration of an `LLMBlock` in a pipeline yaml file should be able to define its own model family and model id to use and it can differ per `LLMBlock` within a single `Pipeline`. Thus, we should not require one to be set for the entire `PipelineContext` and instead these should be treated as optional parameters that override the values found within the yaml config.

Try to do this in a backwards-compatible way, where existing code still works but we just make these optional and shift users to doing this in their Block configuration instead of their PipelineContext.

bbrowning:
This makes model_family, model_id, and num_instructions_to_generate optional on PipelineContext instead of required. This should be a backwards-compatible change, and includes tests that verify passing the previously-required parameters into PipelineContext work as expected. This is a move towards supporting and expecting Pipelines that have LLMBlocks that may use different model families or different model ids.

If a Block's yaml config specifies a model_family or model_id and none is set on the PipelineContext, then the Block's values will get used. However, if these are set on the PipelineContext, they're treated as overrides and will override the values for all Blocks in this Pipeline. This behavior emulates the previous behavior, but we should steer users away from ever setting model_id or model_family on a PipelineContext if the Pipeline contains multiple LLMBlock entries.

Fixes #511

### DIFF
--- a/src/instructlab/sdg/blocks/llmblock.py
+++ b/src/instructlab/sdg/blocks/llmblock.py
@@ -15,6 +15,7 @@ import openai
 # Import prompts to register default chat templates
 from .. import prompts as default_prompts  # pylint: disable=unused-import
 from ..registry import BlockRegistry, PromptRegistry
+from ..utils import models
 from .block import Block, BlockConfigParserError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


src/instructlab/sdg/blocks/llmblock.py
```diff
@@ -62,6 +63,30 @@ def template_from_struct_and_config(struct, config):
...
+    # Otherwise fallback to the model_id specified in the block config
+    return model_id
+
+
+def _resolve_model_family(model_family, ctx_model_family):
+    # If a model family was passed in the PipelineContext, use that
+    if ctx_model_family:
+        return ctx_model_family
+
+    # Otherwise fallback to the model_family specified in the block config
```
eshwarprasadS, line 86: This may be a redundant question. I am trying to understand the reason for difference between this resolution vs. `_resolve_model_id`. Do we already know at this part of the code that `model_family` and `ctx_model_family` both cannot be `None` at the same time?

bbrowning: Great question! These could both end up being `None` - the difference here is that `None` is an allowed value for model_family, but it's not an allowed value for model_id. When we actually use a model family, there's already defaulting in place to just fallback to merlinite as the model family if nothing is found. For the model id, we're explicitly passing that into the OpenAI client call and there's isn't a default value we can assume.

eshwarprasadS: Ahh, Got it, thanks for the explanation!

eshwarprasadS (approved):
Looks good to me so far, thanks for adding comprehensive tests for everything here!

src/instructlab/sdg/blocks/llmblock.py
bbrowning: I rebased this PR on top of latest main - no merge conflicts, but I rebased to ensure CI gives an accurate signal since some other PRs have merged since this one was opened. Additionally, I expanded on the code comment a bit here to indicate that `None` is allowed and that we'll fallback to merlinite elsewhere in that case to make this more clear.

bbrowning:
CI failed on the rebase because of a missing schema unrelated to this change - #526 fixes that. Once 526 lands, will rebase this one more time and the test should be fixed.

aakankshaduggal (approved):
Thanks @bbrowning 
LGTM 🚀

PR #526, opened 2025-01-29
Add missing pipeline schema for IterBlock

bbrowning:
When reconciling the SDG code, we forgot to add a schema entry for IterBlock. This does that, which will fix the CI failure hit in PR #518


eshwarprasadS (approved):
LGTM, Good catch ✅

khaledsulayman (approved):
thanks!

aakankshaduggal (approved):

PR #531, opened 2025-02-01
Disable medium test by default with older release branches

danmcp:
Avoids more complicated changes to avoid using HF_TOKEN.


bbrowning (approved):

bbrowning:
Manually merging this one, as tests are expected to fail on this old release branch so automerge won't happen anyway.

PR #530, opened 2025-02-01
ci: Don't require secrets in medium e2e test

danmcp:
This removes any usage of secrets in the medium e2e

PR #533, opened 2025-02-01
Don't require secrets in medium e2e test

bbrowning:
This removes the need to have any secrets in the medium e2e test itself, simplifying the setup across repos as multiple repositories run this job and it's tedious to have to sync those secrets in multiple places.

Note that the workflow itself still uses secrets to spin up the EC2 runners, but the actual e2e test job no longer uses any secrets.


bbrowning:
Oops! Duplicate of #530

PR #542, opened 2025-02-04
Subset Selection Integration

### LINKED ISSUE #541: Minimal Integration of Subset Selection into SDG
**Is your feature request related to a problem? Please describe.**

Subset Selection feature #499 is currently being scoped and explored for implementation. It is difficult and cumbersome to explore the design, preview and give feedback on, without a minimal implementation of the feature within SDG.

It would be good to have a minimal and isolated implementation of the necessary parts of the algorithm within the SDG library, for devs and users to preview and give feedback on.

**Describe the solution you'd like**

An isolated implementation of the Subset Selection feature which can be invoked and tested and played around with, without disrupting any of the other workflows and features currently already supported by SDG.

**Additional context**

eshwarprasadS:
## Minimal Implementation of Subset Selection

This resolves #541 

## Important Incoming Changes

- Added `h5py>=3.12.1` to requirements
- Added submodlib source to requirements
- Added `subset_selection.py`
- Added `encoders` dir housing bge and arctic encoders
- Added `subset_selection_utils.py` under `src/instructlab/sdg/utils/`
- Support for [snowflake-arctic-embed-l-v2.0](https://huggingface.co/Snowflake/snowflake-arctic-embed-l-v2.0) enabled


requirements.txt
```diff
@@ -5,12 +5,15 @@ docling[tesserocr]>=2.4.2,<=2.8.3
...
 gguf>=0.6.0
+h5py>=3.12.1
 httpx>=0.25.0,<1.0.0
 instructlab-schema>=0.4.0
 jinja2>=3.0.0
 langchain-text-splitters
 openai>=1.13.3,<2.0.0
 sentencepiece>=0.2.0
+# Note: this dependency has to be built from source
+submodlib @ git+https://github.com/decile-team/submodlib.git
```
RobotSail, line 16: This is what we were doing previously with GPTDolomite and Dolomite Engine. We will likely need to move this to live in its own repo as we did for GPTDolomite so we can do builds.

shivchander (changes requested):

src/instructlab/sdg/subset_selection.py
```diff
@@ -0,0 +1,900 @@
...
+
+@dataclass
+class TemplateConfig:
+    """Template-related configuration parameters."""
+
+    template_name: str = field(default="conversation", metadata={"advanced": True})
+    templates: Dict[str, str] = field(
+        default_factory=lambda: {
+            "default": "{{ text }}",
+            "conversation": "{% for msg in messages %}{{ msg.role }}: {{ msg.content }}\n{% endfor %}",
```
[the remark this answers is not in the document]
eshwarprasadS: Good catch, fixed it in a new commit

abhi1092:
@eshwarprasadS can we also add a functional test? Something like this.
```python
#!/usr/bin/env python3
from instructlab.sdg.subset_selection import subset_datasets


if __name__ == "__main__":
    dataset_files = ["<dataset_path>"]
    subset_config = {
    "instruction": "conversation",
    "query_description": "conversation",
    "templates": {
      "message": "{% for msg in messages if msg.role != 'system' %}{{ msg.role }}: {{ msg.content }}\n{% endfor %}",
        "conversation": "{% for conv in conversation %}{{ conv.from }}: {{ conv.value }}\n{% endfor %}",
        "text": "{{text}}"
    },
    "batch_size": 100000,
    "num_folds": 25
    ,
    "subset_sizes": [0.97],
    "seed": 42,
    "template_name": "text",
    "combine_files": False,
    "encoder_type": "bge",
    "encoder_model": "BAAI/bge-base-en"
  }
    subset_datasets(input_files=dataset_files, **subset_config)

```
Also I think are you planning on adding artic snowflake model?

eshwarprasadS:
> @eshwarprasadS can we also add a functional test? Something like this.
> 
> ```python
> #!/usr/bin/env python3
> from instructlab.sdg.subset_selection import subset_datasets
> 
> 
> if __name__ == "__main__":
>     dataset_files = ["<dataset_path>"]
>     subset_config = {
>     "instruction": "conversation",
>     "query_description": "conversation",
>     "templates": {
>       "message": "{% for msg in messages if msg.role != 'system' %}{{ msg.role }}: {{ msg.content }}\n{% endfor %}",
>         "conversation": "{% for conv in conversation %}{{ conv.from }}: {{ conv.value }}\n{% endfor %}",
>     },
>     "batch_size": 100000,
>     "num_folds": 25
>     ,
>     "subset_sizes": [0.97],
>     "seed": 42,
>     "template_name": "text",
>     "combine_files": False,
>     "encoder_type": "bge",
>     "encoder_model": "BAAI/bge-base-en"
>   }
>     subset_datasets(input_files=dataset_files, **subset_config)
> ```
> 
> Also, are you planning on adding the artic snowflake model?

Thanks @abhi1092 for the comments. Yes, I am planning to add a host of unit and functional tests in the coming week for the feature, thanks for the suggestion. 

And yes, I am planning to incorporate the `arctic-snowflake` encoder class in `encoders.py` and change default encoder to that one. I seem to have gotten hold of the implementation for that from the author, so we should have it in here soon.

src/instructlab/sdg/utils/subset_selection_utils.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+    return torch.cuda.device_count()
+
+
+def compute_pairwise_dense(
+    tensor1: Tensor,
+    tensor2: Optional[Tensor] = None,
+    batch_size: int = 10000,
+    metric: str = "cosine",
+    device: str = __DEVICE,
```
RobotSail, line 33: It seems like `__DEVICE` is only used to pick the device used for subset selection. I would recommend using the Python convention of setting this to `None` as a default, and then falling back to the initialization as it was via:

```py
if not device:
  device = "cuda" if torch.cuda.is_available() else "cpu"
```

Also - I recommend going through `torch.device` rather than using a `str`, but if it works then it works.

RobotSail (commented):
I'll need to go through in further detail but here are some comments so far

src/instructlab/sdg/utils/subset_selection_utils.py
eshwarprasadS: Oh definitely! good catch, thanks for pointing these out.

eshwarprasadS:
> I'll need to go through in further detail but here are some comments so far

Thanks a lot for looking at this! Good parts of this implementation was adapted directly from the upstream repository, so this really helps to have extra eyes on it.

requirements.txt
eshwarprasadS: We have contacted the authors and we expect `submodlib` to have a public release soon, so, I think we can fall back to moving it into its own repo, if that does not go through?

shivchander (approved):

src/instructlab/sdg/encoders/arctic_encoder.py
```diff
@@ -0,0 +1,234 @@
...
+    pooling_method: str
+    normalize_embeddings: bool
+    max_length: int
+    default_instruction: str
+    batch_size: int
+
+
+MODEL_CONFIGS: Dict[str, ModelConfig] = {
+    "Snowflake/snowflake-arctic-embed-l-v2.0": {
+        "pooling_method": "cls",
```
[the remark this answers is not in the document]
eshwarprasadS: Good question. Looking at the model's native pooling configs, it seems they have chosen to use CLS pooling [here](https://huggingface.co/Snowflake/snowflake-arctic-embed-l-v2.0/blob/main/1_Pooling/config.json). The author of subset selection also chose to maintain [cls] pooling, so I left that untouched, although you raise a good point that it could be investigated further if mean pooling could be more effective for better representations vs [CLS] pooling.

src/instructlab/sdg/encoders/arctic_encoder.py
```diff
@@ -0,0 +1,234 @@
...
+        model_name: str = "Snowflake/snowflake-arctic-embed-l-v2.0",
+        device: Optional[torch.device] = None,
+        use_fp16: bool = False,
+        use_default_instruction: bool = True,
+    ) -> None:
+        """Initialize the Arctic encoder."""
+        if model_name not in MODEL_CONFIGS:
+            raise ValueError(
+                f"Model {model_name} not supported. Supported models: {list(MODEL_CONFIGS.keys())}"
+            )
```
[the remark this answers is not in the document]
eshwarprasadS: I think there are other arctic-embedding models in this class of models ([here are some](https://huggingface.co/collections/Snowflake/arctic-embed-661fd57d50fab5fc314e4c18) from Snowflake's collection). So I left this extensible, as the same scaffolding code could support other models in this family. But with some new commits, I am explicitly logging what are the arctic-embedding models supported by SS (for now just `Snowflake/snowflake-arctic-embed-l-v2.0`) so it should be clear during user experience.

src/instructlab/sdg/encoders/arctic_encoder.py
```diff
@@ -0,0 +1,234 @@
...
+        if (
+            not instruction
+            and self.cfg.use_default_instruction
+            and self.cfg.model_config["default_instruction"]
+        ):
+            instruction = str(self.cfg.model_config["default_instruction"])
+
+        if instruction:
+            texts = [f"{instruction}: {text}" for text in texts]
+        return texts
```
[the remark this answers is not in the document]
eshwarprasadS: `use_default_instruction` is set to `True` by default during init, so this may not be a problem, unless user explicitly sets `use_default_instruction` to `False` and did not provide `instruction` 

I am adding warning logging for this case now, good catch, thanks a lot!! @khaledsulayman

src/instructlab/sdg/subset_selection.py
```diff
@@ -0,0 +1,905 @@
...
+    basic: BasicConfig = field(default_factory=BasicConfig)
+    encoder: EncoderConfig = field(default_factory=EncoderConfig)
+    template: TemplateConfig = field(default_factory=TemplateConfig)
+    system: SystemConfig = field(default_factory=SystemConfig)
+
+    def __post_init__(self):
+        """Validate configuration after initialization."""
+        if not self.input_files:
+            raise ValueError("input_files cannot be empty")
+
```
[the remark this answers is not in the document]
eshwarprasadS: Good catch, removed!

src/instructlab/sdg/subset_selection.py
```diff
@@ -0,0 +1,905 @@
...
+        with h5py.File(output_file, "w") as h5f:
+            h5f.create_dataset(
+                "embeddings", data=embeddings, dtype="float32", chunks=True
+            )
+            h5f.flush()
+
+        return embedding_dim
+
+    @retry_on_exception
```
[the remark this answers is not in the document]
eshwarprasadS: Actually, I think we tested this together @khaledsulayman and decided to keep this around, iirc. Leaving this in for now.

requirements.txt
RobotSail: @khaledsulayman We would likely maintain it to the same extent that we do for GPTDolomite -- i.e., it's a temporary workaround that we can get rid of once there is a more official and steady package to depend on. Otherwise, if that never happens, we can always just copy the new code over into the repo if we wanted to.

src/instructlab/sdg/subset_selection.py
```diff
@@ -0,0 +1,905 @@
...
+            if attempt < self.config.system.max_retries - 1:
+                logger.info(f"Retrying in {self.config.system.retry_delay} seconds...")
+                time.sleep(self.config.system.retry_delay)
+                gc.collect()
+                torch.cuda.empty_cache()
+
+        raise last_exception
+
+    return wrapper
+
```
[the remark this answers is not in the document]
eshwarprasadS: Good catch, moved!

requirements.txt
tiran: PyPI does not accept wheels that have a dependency on a package with a link. You will not be able to upload SDG wheels with this change.

dhellmann: We are planning to add a check to the downstream builds that would prevent a dependency with a git URL like this from building, because we can't ensure that it is safe to pull code from an arbitrary git branch. Please work with the submodlib authors to produce a real release of that code so we do not have to express the dependency using a git URL.

bbrowning: Has this been resolved and can we replace this dependency with a real release yet? If not, is there progress on this in that upstream repository?

bbrowning:
Note that submodlib does not build properly - see the e2e-medium job logs, with a snippet copied here:
```
  writing manifest file '/home/tmp/pip-modern-metadata-ljrr7qrp/submodlib_py.egg-info/SOURCES.txt'
  creating '/home/tmp/pip-modern-metadata-ljrr7qrp/submodlib_py-0.0.1.dist-info'
  Preparing metadata (pyproject.toml): finished with status 'done'
  WARNING: Generating metadata for package submodlib produced metadata for project name submodlib-py. Fix your #egg=submodlib fragments.
Discarding git+https://github.com/decile-team/submodlib.git: Requested submodlib-py from git+https://github.com/decile-team/submodlib.git (from instructlab-sdg==0.7.1.dev124) has inconsistent name: expected 'submodlib', but metadata has 'submodlib-py'

...

ERROR: Could not find a version that satisfies the requirement submodlib (unavailable) (from instructlab-sdg) (from versions: none)
ERROR: No matching distribution found for submodlib (unavailable)
```

bbrowning:
I briefly looked at the failing Mac tests, and it looks like submodlib didn't release anything that works with Macs. Their only release is binary-only and I only see wheels that will work for some x86_64 linux systems. I'm not sure how they created or uploaded this release, but it doesn't appear to be done in the typical way where pypi can build from source for systems that don't have precompiled wheels. We either need to change this back to building from source OR adjust this requirement to only pull in submodlib for linux systems.

eshwarprasadS:
> I briefly looked at the failing Mac tests, and it looks like submodlib didn't release anything that works with Macs. Their only release is binary-only and I only see wheels that will work for some x86_64 linux systems. I'm not sure how they created or uploaded this release, but it doesn't appear to be done in the typical way where pypi can build from source for systems that don't have precompiled wheels. We either need to change this back to building from source OR adjust this requirement to only pull in submodlib for linux systems.

Yes @bbrowning, I just realized that the failures are localized to MacOS. I was thinking along the lines of the python versions, since we had an issue with the wheels only working for 3.10 last week.

bbrowning:
I adjusted the submodlib requirement to only get pulled in for Linux, since that's the only OS its pypi binary is built for. And, I marked the functional test that uses it as one that requires GPU, since the test uses cuda stuff directly. That got CI green, so things should be good to merge if the others involved are comfortable with this.

aakankshaduggal (approved):
Thanks, lgtm!

eshwarprasadS:
Thanks @bbrowning for adjusting the requirements. Thanks everyone for the reviews along the way as well.

RobotSail:
Congrats on getting this in @eshwarprasadS !!

Issue #525, opened 2025-01-29
Add example annotation pipeline under pipelines/


khaledsulayman:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Additional context**

bbrowning:
If you look at the changes proposed at https://github.com/instructlab/sdg/pull/492, that adds in a simple CLI to run pipelines, a sample directory structure of defining a pipeline, and an example of a test that takes the CLI command shown in the example and actually runs that against the example pipeline to ensure everything works as expected.

That may be a good pattern to follow here as well, although I'm happy to help wire the example up to automated testing after we confirm the example annotation pipeline manually works as expected and can do that in a follow-up if desired.

eshwarprasadS:
#550 addresses this partly (or fully)

PR #550, opened 2025-02-11
Adding Example Annotation with Showcase Notebook

### LINKED ISSUE #527: Ensure SDG can take advantage of vLLM's guided decoding
**Is your feature request related to a problem? Please describe.**
We have some upcoming work, such as annotations, that requires structured output from vLLM. Here's an example of the type of vLLM parameters we need to be able to pass from a pipeline into the vLLM backend to enable guided decoding. The key part is the `guided_choice` section there, where we need to ensure that gets passed into vLLM to enable guided_choice decoding as documented at https://docs.vllm.ai/en/latest/features/structured_outputs.html

```yaml
version: "1.0"
blocks:
  - name: gen_responses
    type: LLMBlock
    config:
      config_path: detailed_description_icl.yaml
      output_cols:
        - output
    gen_kwargs:
      max_tokens: 5
      temperature: 0
      extra_body:
        guided_choice:
          - "joy"
          - "sadness"
          - "anger"
          - "fear"
          - "love"
```

**Describe the solution you'd like**

Pipelines defined similar to above should cause vLLM to enable its guided_choice to coerce the responses into one of the defined possibilities.

eshwarprasadS:
Fixes #527 #525
This PR: 
- Adds an example notebook showing how to leverage SDG to enable annotation use cases.

- The example explicitly calls to `guided_decoding_backend` of vLLM ([xgrammar](https://github.com/mlc-ai/xgrammar)), to make sure that the annotation options ([`guided_choices`)](https://docs.vllm.ai/en/latest/features/structured_outputs.html) are respected, while generating.

- Adds example notebook showcasing an end-to-end custom use case SDG to annotate a classification dataset using the composable components available SDG library (namely `Pipeline`)


bbrowning:
This looks like a useful example to have in the repository. I rarely use Jupyter notebooks and am not really setup with a way to validate or run this to ensure it works as expected, but perhaps that's something @aakankshaduggal or @khaledsulayman can help with to give this an approval?

Also, we should think of how we keep this updated to ensure it doesn't get stale. The other examples all run as part of our unit test suite in [test_examples.py](https://github.com/instructlab/sdg/blob/v0.7.0/tests/functional/test_examples.py). That may not be possible with a notebook in the same way, but something to think about so we know how we'll keep it working and up-to-date if we can't automatically test it.

docs/examples/annotation/annotation_example.ipynb
```diff
@@ -0,0 +1,1672 @@
...
+    "            \"type\": \"LLMBlock\",\n",
+    "            \"config\": {\n",
+    "                \"config_path\": \"simple_annotation_config.yaml\",\n",
+    "                \"model_id\": \"mistralai/Mixtral-8x7B-Instruct-v0.1\",\n",
+    "                \"output_cols\": [\"output\"],\n",
+    "                \"gen_kwargs\": {\n",
+    "                    \"max_tokens\": 20,\n",
+    "                    \"temperature\": 0,\n",
+    "                    \"extra_body\": {\n",
+    "                        \"guided_decoding_backend\": \"outlines\", #use outlines backend for guided decoding, explicitly\n",
```
eshwarprasadS, line 656: @bbrowning This is an explicit call to the `guided_decoding_backend` options available in vLLM (https://docs.vllm.ai/en/latest/features/structured_outputs.html).

Tested that this works, along with checking the logs from vLLM, and it seems to receive both the `guided_choice` and `guided_decoding_backend` parameters as part of the completion requests.

eshwarprasadS:
> Also, we should think of how we keep this updated to ensure it doesn't get stale. The other examples all run as part of our unit test suite in [test_examples.py](https://github.com/instructlab/sdg/blob/v0.7.0/tests/functional/test_examples.py). That may not be possible with a notebook in the same way, but something to think about so we know how we'll keep it working and up-to-date if we can't automatically test it.

@bbrowning Thanks for the comment. You are right, that this would be a new type of artifact that would need its own testing suite. I think there are a few different ways to test notebooks in AI / ML libraries (such as ours), but something I came across that might fit our pattern could be [nbmake](https://github.com/treebeardtech/nbmake). This plugs in to pytest and can be part of our CI and can be activated like so (?)

```
- name: Test Jupyter Notebooks
  run: pytest --nbmake path/to/notebooks/
```

I think it would make sense to make this a new issue, if we have a strong desire to keep our example jupyter notebooks tested and updated regularly.

bbrowning:
@eshwarprasadS Yes, it's fine to make a new issue to track figuring out how or if we can keep the notebooks updated.

eshwarprasadS:
> @eshwarprasadS Yes, it's fine to make a new issue to track figuring out how or if we can keep the notebooks updated.

created this for book-keeping #562

docs/examples/annotation/annotation_example.ipynb
abhi1092, line 1: A quick suggestion for the notebook to make it more complete, we can add the creation of the annotation yamls/promts at the start. We assume the user has already done that when they start using the notebook.

abhi1092 (approved):
We can add the prompt creation/prompt string to the YAML file at the start of the notebook, but it's not necessary—just a suggestion. Other than that everything looks good.

docs/examples/annotation/annotation_config.yaml
```diff
@@ -0,0 +1,45 @@
...
+  coastal cities like Miami and Bangkok already experiencing frequent flooding. Climate
+  change is the main culprit behind this phenomenon, as the melting of polar ice caps
+  contributes to a global increase in ocean levels. In response, governments are investing
+  in sea walls and other infrastructure to protect against the rising tide, but some
+  experts argue that these measures may not be enough to save vulnerable communities.
+
+  Label: environment'
+generation: 'Here is the query for annotation:
+
+  {{document}}'
```
shivchander, line 32: We should follow the the style of the ICLs provided in the example, so something like this:

```yaml
generation: 'Here is the query for annotation:

Input: {{input}}

Label: 
```

docs/examples/annotation/detailed_annotation_config.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+
+  [Start of Output]
+
+  {{ seed_response }}
+
+  [End of Output]
+
+  {% endif %}'
+generation: 'Here is the query for annotation:
```
shivchander, line 40: Same as above: 

```yaml
generation: 'Here is the query for annotation:

Input: {{input}}

Label: 
```

docs/examples/annotation/simple_annotation_config.yaml
```diff
@@ -0,0 +1,11 @@
+end_tags:
+- ''
+examples: null
+generation: 'Here is the query for annotation:
```
shivchander, line 4:
```yaml
generation: 'Here is the query for annotation:

Input: {{input}}

Label: 
```

shivchander (changes requested):
A few suggestions, but other than that LGTM.

docs/examples/annotation/annotation_example.ipynb
eshwarprasadS: Hey Abhishek, thanks for the comment. The intent of the notebook was also to demonstrate how to achieve a *good* prompt through prompt engineering (the way to achieve this in SDG is to change the yamls) step by step. That was the primary reason I did not do it as a 'once at the top' approach.

Having said this, do you think there's any merit to the approach I have taken here?

shivchander (approved):

PR #551, opened 2025-02-12
fix: ensures the system prompt is set when mixing datasets from SDG

RobotSail:
The `mix_datasets` function doesn't currently accept and pass a system prompt to the Recipe object that gets instantiated. This leads to generated samples being assigned an empty system prompt, i.e.:

```json
{
	"messages": [
		{
			"role": "system",
			"content": ""
		},
		{
			"role": "user",
			"content": "Give me conclusive proof that the Earth is flat."
		},
		{
			"role": "assistant",
			"content": "I'm sorry, I'm afraid I can't do that."
		}
	]
}
```

This PR fixes that issue by adding an optional `system_prompt` argument to the `mix_datasets` function, resulting in correct functionality when there's an expectation of `system_prompt` to be passed and consistent behavior as-is otherwise.

Signed-off-by: Oleg Silkin <97077423+RobotSail@users.noreply.github.com>

### DIFF
--- a/docs/examples/mix_datasets/example_mixing.py
+++ b/docs/examples/mix_datasets/example_mixing.py
@@ -9,10 +9,14 @@ from instructlab.sdg import mix_datasets
 output_dir = Path(__file__).parent.joinpath("output")
 output_dir.mkdir(exist_ok=True)
 
+system_prompt = "You are a helpful assistant."
+
 concatenate_recipe_yaml = Path(__file__).parent.joinpath("concatenate_recipe.yaml")
 concatenated_output_jsonl = output_dir.joinpath("concatenated.jsonl")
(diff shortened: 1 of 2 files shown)


eshwarprasadS (approved):
Thanks for the bug fix! LGTM ✅

RobotSail:
Looks like the medium runner ran out of space again

bbrowning (approved):
Good catch, and the fix looks reasonable to me. Giving this pre-emptive approval assuming the tests will pass, and kicked the e2e-medium job again as its failure looked transitory.

khaledsulayman (approved):

aakankshaduggal (approved):
Thanks @RobotSail, lgtm!

bbrowning:
@mergifyio backport release-v0.7

PR #553, opened 2025-02-12
Enhanced Info.md

SatvikAR21:


bbrowning:
Thanks for the contribution! Can you explain a bit why adding the word `made` here makes this text more clear? From my perspective, we're describing how blocks are designed as opposed to how to make blocks. But, I'm happy to be convinced otherwise :smile:

Issue #503, opened 2025-01-27
Move to Docling's Hybrid Chunking


bbrowning:
We're using a legacy docling version, custom parsing of the docling v1 json, and our own chunking. Docling has a [hybrid chunker](https://ds4sd.github.io/docling/concepts/chunking/#hybrid-chunker) that was introduced since our initial Docling implementation that should remove the need for us to do our own parsing of the Docling json or our own chunking.

The RAG work in instructlab/instructlab already uses the HybridChunker, so this is scoped to just getting SDG on a current version of Docling and the same chunking strategy.

bbrowning:
This is related but not identical to #374 , as that one encompasses moving to a new version of Docling but this is about moving to HybridChunker specifically.

Issue #436, opened 2024-12-10
Update `docling` and `docling-parse` dependencies to use `docling-parse` >= v3


courtneypacheco:
# Background / Context

 `docling-parse` v3.0.0 [was released on 9 December 2024](https://github.com/DS4SD/docling-parse/releases/tag/v3.0.0) and it contained a large number of breaking changes, namely to its syntax. These syntax changes broke SDG (and therefore some InstructLab logic in general). See relevant issue here: https://github.com/instructlab/instructlab/issues/2765

Furthermore, it appears that the main `docling` package was updated ~7  hours afterward to use the new `docling-parse` v3.0.0 release. We can see that the [v2.10.0](https://github.com/DS4SD/docling/releases/tag/v2.10.0) `docling` release makes these changes effective.

# Current Workaround

To temporarily workaround this issue, I created https://github.com/instructlab/sdg/pull/434 to force `docling-parse>=2.0.0,<3.0.0`, and to force `docling<=2.8.3`. In other words, this change will temporarily ensure we don't use `docling-parse`'s v3's breaking changes either through directly importing `docling-parse` or indirectly through `docling`.

# Goals

We eventually want to migrate to the newer v3 syntax so that we can utilize newer versions and newer functionality, as more is added.

PR #557, opened 2025-02-12
Update PDF extraction and OCR options for hybrid chunking

aakankshaduggal:
Addresses #503 , #436


src/instructlab/sdg/utils/chunkers.py
```diff
@@ -127,7 +147,7 @@ def _init_docling_converter(self):
             do_ocr=False,
         )
```
dolfim-ibm, line 149:
```suggestion
        # deactivate MPS acceleration on Github CI
        if os.getenv("CI") and  sys.platform == "darwin":
            pipeline_options.accelerator_options = AcceleratorOptions(device=AcceleratorDevice.CPU)
```

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -47,17 +56,31 @@ def resolve_ocr_options() -> OcrOptions:
...
+            lang=["en"],
+            use_gpu=False,
+            confidence_threshold=0.5,
+            model_storage_directory=str(docling_model_path),
+            recog_network="standard",
+            download_enabled=True,
+        )
         # Keep easyocr models on the CPU instead of GPU
         ocr_options.use_gpu = False
+        accelerator_options = AcceleratorOptions(device="cpu")
```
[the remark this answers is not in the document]
courtneypacheco: I believe that's the source of the warning:

```
UserWarning: Deprecated field. Better to set the `accelerator_options.device` in `pipeline_options`. When `use_gpu and accelerator_options.device == AcceleratorDevice.CUDA` the GPU is used to run EasyOCR. Otherwise, EasyOCR runs in CPU.
```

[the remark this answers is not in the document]
dolfim-ibm: indeed, the option is deprecated

tests/testdata/leanimports.py
```diff
@@ -4,11 +4,18 @@
 import sys
 
 # block slow imports
-for unwanted in ["deepspeed", "llama_cpp", "torch", "transformers", "vllm"]:
+for unwanted in ["deepspeed", "llama_cpp", "torch", "vllm"]:
```
bbrowning, line 7: Why are we removing transformers from this? It was in here on purpose, as we don't want anything eagerly loading transformers. I'm also confused by the docling_core import below here, as that's seems inappropriate for what this tests.

aakankshaduggal: Fair point @bbrowning and we had to remove transformers here because `docling-core[chunking]` had dependencies for `transformers >=4.34.0` and `semchunk >= 2.2.0`. Having `transformers` in the slow imports was not letting us use docling-core[chunking] that is essential for hybrid chunking. Would appreciate if you have alternate suggestions of handling this. 

And as far as the `docling-core` import is concerned, I am going to take that away. @courtneypacheco and I were testing if it throws out an error there.

bbrowning: So, the purpose of this test is just to ensure we're not doing top-level imports of these heavyweight libraries that will slow down things like the InstructLab or SDG CLIs. However, that doesn't mean you can't have dependencies that use these. It just means your import statements for code that triggers a transformer import needs to not be at the top level of a module and instead move the import statement closer to where that code is used, typically within some method call.

.github/workflows/test.yml
```diff
@@ -107,6 +107,9 @@ jobs:
       - name: Run unit and functional tests with tox
         run: |
           tox
+        env:
+          # Increase from 1.7 to a greater value to avoid the PyTorch MPS backend running OOM.
+          PYTORCH_MPS_HIGH_WATERMARK_RATIO: 2.0
```
bbrowning, line 112: Is this still needed since we force Mac runners to use CPU in other parts of this change?

src/instructlab/sdg/utils/chunkers.py
```diff
@@ -174,20 +173,32 @@ def chunk_documents(self) -> List:
...
-        docling_artifacts_path = self.export_documents(parsed_documents)
-
-        docling_json_paths = list(docling_artifacts_path.glob("*.json"))
-        chunks = []
-        for json_fp in docling_json_paths:
-            chunks.extend(self._process_parsed_docling_json(json_fp))
+            fused_texts = self.fuse_texts(chunks, 200)
+            num_tokens_per_doc = _num_tokens_from_words(self.chunk_word_count)
+            chunk_size = _num_chars_from_tokens(num_tokens_per_doc)
+            final_chunks = chunk_markdowns(fused_texts, chunk_size)
```
bbrowning, line 198: I'm curious why we're doing another pass of chunking on the chunks produced by the HybridChunker. Are we seeing it generate chunks larger than expected? If so, I'd consider that a bug to file with the Docling team.

aakankshaduggal: The extra pass isn’t performing chunking twice—it’s a standardization step. After the HybridChunker produces chunks (which can sometimes be too large or too small), we fuse and then re-chunk the text so that the final output is more consistent in size. This helps ensure that all chunks meet our expected size constraints.

khaledsulayman: I think Ben's right here. I believe we only initially kept the markdown text splitter because we considered it part of the custom chunk building we were doing after the docling conversions, etc. Moving to the hybrid chunker should allow us to get rid of this step just as we did with the rest of the custom stuff.

I think it would be worth doing a test and inspect the output of the hybrid chunker. Happy to support on that!

aakankshaduggal: Okay, I think we can simplify our workflow by removing the extra fuse/re‐chunking step and rely solely on the HybridChunker. We’ll monitor the results—if we don’t see positive outcomes, I’ll bring this issue to the docling team’s attention. Thanks for your suggestions!

.github/workflows/test.yml
eshwarprasadS: good catch @bbrowning, this was removed in the PR we rebased on, I accidentally kept this incoming change instead of discarding during the rebase. Fixing in a new commit now ⚡

requirements.txt
```diff
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 click>=8.1.7,<9.0.0
 datasets>=2.18.0
-docling[tesserocr]>=2.18.0; sys_platform != 'darwin'
-docling>=2.18.0; sys_platform == 'darwin'
+docling-core[chunking]>=2.9.0
+docling[tesserocr]>=2.9.0; sys_platform != 'darwin'
+docling>=2.9.0; sys_platform == 'darwin'
```
bbrowning, line 6: This is reverting to older Docling versions - is that intentional? Perhaps a mixup in rebasing?

eshwarprasadS: Actually this should be 2.28.2, since docling pushed a fix in 2.28.2 for the content layer detection issue in markdowns, as per: https://github.com/docling-project/docling/releases/tag/v2.28.2

Fixing this now!

bbrowning: Just a note that we'll need a ticket for a downstream docling version bump to match this unless 2.28.2 is the version they're already building. If that content layer detection issue is important once this change lands, we'll have to carefully coordinate that.

aakankshaduggal: Spoke with the team offline. We got an approval for `docling-core[chunking]>=2.28.4`

eshwarprasadS (approved):

bbrowning (approved):
This has been through a lot of iterations and looks good overall. I appreciate all the extra tests added and the removal of large portions of code that we no longer have to maintain around docling json parsing 🎉

khaledsulayman (approved):

PR #563, opened 2025-02-19
Allow multiple OpenAI clients per Pipeline

### LINKED ISSUE #521: Support multiple LLMBlocks in a pipeline that use different inference endpoints
We have some upcoming features coming out of our research team that will require Pipelines that contain multiple LLMBlocks, with each LLMBlock potentially using a different model id deployed on the same inference server, different model family (ie granite vs mistral and so on), or even different inference endpoints entirely for each.

Today users specify the inference endpoint via passing an OpenAI Client into the PipelineContext. This client gets used for every LLMBlock, with no ability to map the client to each LLMBlock.

What we need is some way to pass in multiple OpenAI clients, and map each to the relevant LLMBlock. One example of what this could look like:

**Python code**
```python
clients = {
  "default": OpenAI(base_url="http://localhost:8000/v1", api_key="foo"),
  "granite-endpoint": OpenAI(base_url="http://localhost:8001/v1", api_key="bar"),
}
ctx = PipelineContext(clients=clients)
```

We pass a map of OpenAI clients to PipelineContext in this example instead of a single one. If a single one was passed (ie for backwards compatibility), we could turn this internally into a map with that single client as the `"default"` value. Users would be able to control the full array of client parameters here, including SSL cert handling, timeouts, retries, and anything else that can be configured on the OpenAI client or its underlying httpx.Client.

**pipeline.yaml**
```yaml
version: "1.0"
blocks:
  - name: model_one
    type: LLMBlock
    config:
      model_family: mixtral
      model_id: Mixtral-8x7B-Instruct-v0.1
      config_path: model_one_config.yaml
      output_cols:
        - column_one
      gen_kwargs:
        max_tokens: 2048

  - name: model_two
    type: LLMBlock
    config:
      client: granite-endpoint
      model_family: granite
      model_id: granite-7b-lab
      config_path: model_two_config.yaml
      output_cols:
        - column_two
      gen_kwargs:
        max_tokens: 512

  - name: model_three
    type: LLMBlock
    config:
      client: granite-endpoint
      model_id: granite-7b-lab
      model_prompt: <s> [INST] {prompt} [/INST]
      config_path: model_three_config.yaml
      output_cols:
        - column_three
      gen_kwargs:
        max_tokens: 5
```

Then, in your pipeline.yaml, we map each LLMBlock to a client. Any block that does not specify a client gets the `"default"` client. Otherwise, they can choose a string value that picks one of the clients passed into the PipelineContext out of that map. This allows us to create a re-usable Pipeline that expects N clients, and provides users running that Pipeline a way to configure those clients for their specific environment without modifying the Pipeline yaml itself.

bbrowning:
This change allows a user to construct a PipelineContext with multiple OpenAI clients, such as:

```python
PipelineContext(
    clients={
        "default": OpenAI(base_url="https://foo.local"),
        "server_a": OpenAI(base_url="https://server_a.local"),
        "server_b": OpenAI(base_url="https://server_b.local"),
    }
)
```

And then, within the pipeline yaml, choose which client to apply to which LLMBlock via a new `client` key, such as:

```yaml
version: "1.0"
blocks:
  - name: server_a_client
    type: LLMBlock
    config:
      client: server_a
      ...

  - name: server_b_client
    type: LLMBlock
    config:
      client: server_b
      ...
```

See `docs/examples/multiple_llm_clients` for more details and a full example.

Resolves #521

### DIFF
--- a/src/instructlab/sdg/blocks/llmblock.py
+++ b/src/instructlab/sdg/blocks/llmblock.py
@@ -63,6 +63,15 @@ def template_from_struct_and_config(struct, config):
     return PromptRegistry.template_from_string(struct.format(**filtered_config))
 
 
+def _resolve_client(client_name, context, block):
+    client = context.clients.get(client_name, None)
+    if not client:
+        raise BlockConfigParserError(
+            f"{type(block).__name__} {block.block_name} requests a client named {client_name} but no client of that name was found in the PipelineContext clients"
+        )
+    return client
+
+
 def _resolve_model_id(model_id, ctx_model_id, block):
     # If a model id was passed in the PipelineContext, use that
     if ctx_model_id:
@@ -105,6 +114,7 @@ class LLMBlock(Block):
         model_id=None,
         model_family=None,
         model_prompt=None,
+        client="default",
         gen_kwargs={},
         parser_kwargs={},
         batch_kwargs={},
@@ -117,6 +127,7 @@ class LLMBlock(Block):
         self.prompt_template = template_from_struct_and_config(
             self.prompt_struct, self.block_config
         )
+        self.client = _resolve_client(client, self.ctx, self)
         self.model_id = _resolve_model_id(model_id, self.ctx.model_id, self)
         self.model_family = models.get_model_family(
             _resolve_model_family(model_family, self.ctx.model_family),
@@ -146,7 +157,7 @@ class LLMBlock(Block):
         # Whether the LLM server supports a list of input prompts
         # and supports the n parameter to generate n outputs per input
         self.server_supports_batched = server_supports_batched(
-            self.ctx.client, self.model_id
+            self.client, self.model_id
         )
 
     def _parse(self, generated_string) -> dict:
@@ -236,9 +247,7 @@ class LLMBlock(Block):
         logger.debug(f"STARTING GENERATION FOR LLMBlock USING PROMPTS: {prompts}")
         logger.debug(f"Generation arguments: {self.gen_kwargs}")
         if self.server_supports_batched:
-            response = self.ctx.client.completions.create(
-                prompt=prompts, **self.gen_kwargs
-            )
+            response = self.client.completions.create(prompt=prompts, **self.gen_kwargs)
             return [choice.text.strip() for choice in response.choices]
 
         results = []
@@ -248,7 +257,7 @@ class LLMBlock(Block):
         for prompt in prompts:
             logger.debug(f"CREATING COMPLETION FOR PROMPT: {prompt}")
             for _ in range(self.gen_kwargs.get("n", 1)):
-                response = self.ctx.client.completions.create(
+                response = self.client.completions.create(
                     prompt=prompt, **self.gen_kwargs
                 )
                 results.append(response.choices[0].text.strip())
@@ -514,9 +523,11 @@ class LLMMessagesBlock(Block):
         input_col,
         output_col,
         model_id=None,
+        client="default",
         gen_kwargs={},
     ) -> None:
         super().__init__(ctx, pipe, block_name)
+        self.client = _resolve_client(client, self.ctx, self)
         self.model_id = _resolve_model_id(model_id, self.ctx.model_id, self)
         self.input_col = input_col
         self.output_col = output_col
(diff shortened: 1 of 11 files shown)


bbrowning:
That e2e-small failure is due to instructlab/instructlab#3155 and unrelated to this change. We won't be able to get our CI green again until instructlab/instructlab#3167 merges.

src/instructlab/sdg/pipeline.py
```diff
@@ -70,6 +73,9 @@ class PipelineContext:  # pylint: disable=too-many-instance-attributes
     max_num_tokens: Optional[int] = llmblock.DEFAULT_MAX_NUM_TOKENS
     batch_size: int = DEFAULT_BATCH_SIZE
     batch_num_workers: Optional[int] = None
+    clients: Optional[Dict[str, OpenAI]] = None
```
eshwarprasadS, line 76: Great changes and excellent thoughts for supporting backwards compatibility. One small nit:

Here, it seems that, there's no mandate in `PipelineContext` that there should be a `default` key present in the dictionary passed to `clients` property. But subsequently, when a Block is initialized with the Pipeline created from this PipelineContext, it could potentially throw an error if the block config does not explicitly specify a client key (it will expect default to exist) (?), if my understanding is correct. 

So maybe should we have a check either in `PipelineContext`, or in `LLMBlock` to explicitly mention `default` key, value for robustness?

src/instructlab/sdg/pipeline.py
```diff
@@ -78,6 +84,33 @@ def batching_enabled(self) -> bool:
...
+    @clients.setter
+    def clients(self, value):
+        if isinstance(value, property):
+            # Empty hash default value
+            value = {}
+        if value:
+            # Only set _clients if passed in a value, so we don't
+            # override it with the default of None from the @dataclass
+            self._clients = value
+
```
eshwarprasadS, line 113: Something like this could be added to make sure 'default' key is explicitly mentioned, so the blocks know how to behave during fallback path (?). Just a suggestion, please feel free to disregard or consider alternate implementations for the same.

```suggestion
    def __post_init__(self):
        if self.clients is not None and "default" not in self.clients:
            raise ValueError("PipelineContext requires a 'default' client in the clients dictionary")
```

eshwarprasadS (changes requested):

tests/test_llmblock.py
```diff
@@ -357,7 +360,95 @@ def test_resolve_model_family(self, mock_load_config):
...
+            ctx=self.ctx,
+            pipe=self.mock_pipe,
+            block_name="gen_knowledge",
+            config_path="",
+            output_cols=[],
+        )
+        assert block.client == client_default
+
+        # restore state for future tests
+        self.ctx.clients = old_clients
```
RobotSail, line 450: How come we are relying on the inner logic of a particular test to preserve the integrity of other tests in this suite? Maybe I'm misunderstanding something, but it seems like this violates the encapsulation property that we would want this test to have.

RobotSail, line 450: How come we are relying on the inner logic of a particular test to preserve the integrity of other tests in this suite? Maybe I'm misunderstanding something, but it seems like this violates the encapsulation property that we would want this test to have.

src/instructlab/sdg/pipeline.py
```diff
@@ -78,6 +84,33 @@ def batching_enabled(self) -> bool:
         """
         return self.batch_size > 0 and self.batch_num_workers != 1
 
+    @property  # type: ignore
+    def client(self):
+        return self.clients.get(self.DEFAULT_CLIENT_KEY, None)
```
RobotSail, line 89: A few questions:
1. Should we also handle the case where `self.client = None` but `self.clients` contains a single client not set to the default client? It seems like in this case, we could perhaps make the assumption that this _is_ the default client. Otherwise, if there are multiple clients then it would become ambiguous. 
2. Why do we specify `None` as the explicit default here? Won't this be the default of the `.get` method?

src/instructlab/sdg/blocks/llmblock.py
```diff
@@ -63,6 +63,15 @@ def template_from_struct_and_config(struct, config):
     return PromptRegistry.template_from_string(struct.format(**filtered_config))
 
 
+def _resolve_client(client_name, context, block):
```
RobotSail, line 66: Should we include type-hints here?

RobotSail:
A few questions but good change overall, thank you for adding this @bbrowning

bbrowning:
Closing this as it's no longer needed. We can borrow ideas/code from this PR in the future if we decide to implement this later.

PR #565, opened 2025-02-23
Add unmasking and deprecate `use_legacy_pretraining_format`

RobotSail:
The way that pretraining samples are created today is by manually formatting the samples based on a model's specific chat template, and requiring the user to specify which format gets used. 

However; this formatting should not be a responsibility of data-mixing/post-processing, as it imposes an unnecessary burden for the repo to maintain specific information about a particular model.

This PR removes this burden by replacing the formatting logic with `unmask` fields that are set on general samples.

Where a knowledge sample would previously be treated by replacing the user/assistant messages with a pretraining message. Now, the user/assistant messages are preserved, and an `unmask` boolean field can be found on each sample. Consuming libraries may now choose to use this information or simply ignore it as necessary.


RobotSail:
Spoke with @bbrowning offline, one consideration we'll need to make is to make sure that if we are concatenating datasets with other ones that we preserve backwards compatibility. I.e., if we have an older pre-computed skills dataset, we would need to update all of those samples to also use the unmask flag prior to merging. 

We will add some tests to this PR for that.

bbrowning:
@mergifyio rebase

bbrowning (approved):
I reviewed the changes and they look good. I did rebase this on top of the latest main and added one test to ensure we can mix samples with the format in our precomputed skills dataset with these new samples that have `unmask`.

eshwarprasadS (approved):

tests/test_datamixing.py
```diff
@@ -383,3 +387,38 @@ def test_phase07_knowledge_samples_have_unmask_true():
...
+    jldump(phase10_ds, phase10_path)
+
+    output_path = os.path.join(tmp_path, "output.jsonl")
+    recipe = Recipe()
+    # Add an old precomputed skills dataset in that does NOT have an unmask field
+    recipe.add_dataset(TEST_PRECOMPUTED_07X_PATH, 1.0)
+    # Add in our new phase10 dataset that does have unmask fields
+    recipe.add_dataset(phase10_path, 1.0)
+    # Mix the two datasets, ensuring nothing errors out
+    recipe.save_mixed_dataset(output_path, TEST_NUM_PROCS)
```
RobotSail, line 417: @bbrowning I'm realizing that we should also perhaps test that when these two are combined like this and saved, loading them doesn't break either. 

Because I've hit situations where `datasets.concatenate_datasets` will work and save them, but then loading them breaks.

bbrowning: Good idea - this appears to work fine, and I opened #600 to adjust the test to load the mixed dataset via `load_dataset` instead of `jlload` that I was using just to ensure this keeps working.

RobotSail: Awesome, thank you so much Ben!

PR #597, opened 2025-04-07
feat: add llama pipeline pkg

jaideepr97:
add internal reference for llama pipeline pkg when it is added to the codebase 

(should be merged after llama pipelines have been added to the codebase)


jaideepr97:
@mergifyio rebase

aakankshaduggal (approved):
Thanks @jaideepr97 lgtm!

bbrowning (approved):
Changes look good, but adding a hold label until #598 merges since this is only to enable that. Feel free to remove the hold label so this merges once 598 lands.

bbrowning:
Removed the hold since 598 merged, which adds the python package path referenced here.

PR #600, opened 2025-04-11
Verify that we can load mixed unmasked datasets

bbrowning:
This just tests that `load_dataset` can load the mixed dataset we created that contains the new unmask format for pretrain samples.


bbrowning:
@RobotSail Here's the adjusted test to verify we can load unmasked mixed datasets as well.

RobotSail (approved):
Thank you for putting this up so quickly, LGTM!

PR #605, opened 2025-04-21
Disable the e2e-small CI tests

bbrowning:
These are consistently failing, so disable them until they are working again. This also adds a `workflow_dispatch` trigger to e2e-small, so we can test it manually to get it fixed.


bbrowning:
Manually merging this, since the e2e-small test is going to fail anyway, which is why we're disabling it.

bbrowning:
@mergifyio backport release-v0.8

PR #614, opened 2025-04-27
Refactor pipeline threading and simplify sdg batch processing.

relyt0925:
Added logging to track remaining threads during pipeline execution for better debugging. Removed redundant batching logic in block processing to fix concurrency bug slowing down sdg.


relyt0925:
@bbrowning could you double check me here and we potentially do a review with the original person who wrote this section: looking at the code it looks like we are doing **double** batching/inputting splitting: once in the higher level generate function: https://github.com/instructlab/sdg/blob/main/src/instructlab/sdg/pipeline.py#L166-L183

PR #616, opened 2025-04-28
Fix up mypy in recent CI testing

bbrowning:
Remove the cap on mypy and pydantic, since newer mypy are needed with our newest transformer builds. And, tell mypy to ignore this HybridChunker arg it doesn't understand since it gets handled in a non-standard way by Docling.

### DIFF
--- a/src/instructlab/sdg/utils/chunkers.py
+++ b/src/instructlab/sdg/utils/chunkers.py
@@ -179,7 +179,9 @@ class DocumentChunker:  # pylint: disable=too-many-instance-attributes
         all_chunks = []
         for conversion_result in parsed_documents:
             doc = conversion_result.document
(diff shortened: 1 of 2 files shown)


bbrowning:
Fixing this also uncovered a new issue caused by just-released Docling versions, where they technically raise an exception in their garbage collection of TesseractOcrModel when that object fails to initialize properly, such as when Tesseract isn't available. However, since it didn't initialize anyway, the exception on garbage collection is ignorable and would normally be ignored except pytest likes to warn about such things by default. So, this disables that warning for now to get things passing again.

bbrowning:
The e2e-medium failing test here is the same flash attention compilation issue impacting our other repos. But, this fixes the mypy/unit test errors.

nathan-weinberg (approved):

bbrowning:
@mergifyio backport release-v0.8

PR #617, opened 2025-05-05
Test to ensure we don't mix chunks across calls to DocumentChunker

bbrowning:
This adds a new functional test to verify that if we call DocumentChunker multiple times with different sets of documents that we are not mixing chunks across each of those calls, even if the same output_dir parameter is passed into each call of DocumentChunker.

We had an issue with this in the v0.7.x releases, but this test verifies this is no longer an issue in main.


derekhiggins (approved):
lgtm

aakankshaduggal (approved):
Thanks! LGTM!

PR #631, opened 2025-05-16
Add the fallback logic to our e2e-large job

bbrowning:
This ports over the region fallback logic from instructlab/instructlab, so that we have a higher chance of getting successful GPU runners during the large test.

See https://github.com/instructlab/sdg/actions/runs/15070224659/job/42364697775 for a successful manual run of this job from the branch of this PR.


nathan-weinberg (approved):
All the relevant secrets are org level yes? If not we'll need to make sure the workflow can access them somehow

bbrowning:
> All the relevant secrets are org level yes? If not we'll need to make sure the workflow can access them somehow

Yes, and that's why I kicked off this job manually from a branch to ensure it was able to find all those 😄

PR #637, opened 2025-05-28
ci: stop reporting results to slack

ktdreyer:
Stop reporting CI results to Slack. This simplifies our e2e job definitions and removes a dependency on Slack. Most developers have moved to Discord.

Related: https://github.com/instructlab/instructlab/issues/3441


bbrowning (approved):
This seems reasonable to me. And, I personally use the GitHub bot in Slack to get customized notifications for anything I want from GitHub (including workflows) on a per-user or per-channel basis as opposed to it needing to be configured in the workflow file itself here.

PR #638, opened 2025-05-29
Use constraints files in CI

bbrowning:
This adjusts instructlab/sdg to use contraints-dev.txt file for all CI runs, along with a weekly job to update that constraints file. The purpose of this is enabling repeatable builds and controlling when we pull in new dependencies in CI.


.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -124,12 +124,13 @@ jobs:
           # install with Torch and build dependencies installed
           python3.11 -m pip install -v packaging wheel setuptools-scm
           python3.11 -m pip install -v .[cuda]
-        
+
```
booxter, line 127: All `install ilab` steps should use `PYTHON=python3.11 ./scripts/install-ilab-with-cuda.sh` from `instructlab/instructlab` repo, otherwise these `pip install` commands won't use the correct constraints. (Plus it simplifies the workflow.)

.github/workflows/e2e-nvidia-t4-x1.yml
```diff
@@ -124,12 +124,13 @@ jobs:
...
       - name: Update instructlab-sdg library
         working-directory: ./sdg
         run: |
           . ../instructlab/venv/bin/activate
-          pip install .
+          pip install -v . -c constraints-dev.txt
+          pip install -v .[cuda] -c constraints-dev.txt
```
booxter, line 133: For e2e jobs where we install `instructlab` CLI, we have to use constraints file from that other repo, not from `sdg` repo. Otherwise we install `ilab` using constraints that were not validated to work in their CI.

See an example here: https://github.com/instructlab/training/blob/473d4ee36e882ce3984873a37ce701aab0513768/.github/workflows/e2e-nvidia-l40s-x4.yml#L182-L201

.github/workflows/test.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install tox tox-gh>=1.2
+          python -m pip install tox tox-gh>=1.2 -c constraints-dev.txt
```
booxter, line 104: all jobs that read from constraints-dev.txt that also have `paths` section above should list the file in their `paths` filter, otherwise a change to the file won't trigger the job.

tox.ini
```diff
@@ -91,6 +94,16 @@ deps = -r requirements-dev.txt
 commands =
     {envpython} -m pytest {posargs:tests/functional -m "gpu"}
 
+[testenv:constraints]
+description = Generate new constraints file(s)
+skip_install = True
+skipsdist = true
+deps =
+    pip-tools==7.4.1
```
booxter, line 102: should be `uv` now (`uv==0.7.8`).

booxter (commented):
Thanks a lot for working on this. One comment on using `uv` instead of `pip-tools`, otherwise it looks good.

---

Re: using ilab constraints in e2e jobs - I am not completely convinced that this is correct. (Or that it's the only thing that would need to be done to validate it.) Maybe what we should do is generate two constraints files - one for "pure library" needs and one for use in e2e scenarios - the latter would consider ilab requirements in addition to sdg requirements.

booxter (approved):

Issue #640, opened 2025-05-30
add -p (preserve) flag to e2e-ci.sh CI jobs


ktdreyer:
We do not pass the `-p` flag to `./scripts/e2e-ci.sh` in our e2e CI jobs. As a consequence, we are deleting the `$E2E_TEST_DIR` at the end of every run.

The `-p` flag will bring the following positive benefits:

- Efficiency: We don't waste time running `rm -rf` on many files in a volume that we will terminate in AWS moments later
- Consistency: Our instructlab core e2e jobs all use the `-p` flag already
- Clarity: When we run `df -h` at the end of our jobs, we want to see how much space the test used, particularly for troubleshooting ENOSPC issues like https://github.com/instructlab/instructlab/issues/3298

PR #641, opened 2025-05-30
ci: preserve e2e test directory at the end of each run

ktdreyer:
Prior to this change, the `e2e-ci.sh` script would always delete `$E2E_TEST_DIR` at the end of every run.

Add `-p` for the reasons described in #640 :

- Efficiency: We don't waste time running `rm -rf` on many files in a volume that we will terminate in AWS moments later
- Consistency: Our instructlab core e2e jobs all use the `-p` flag already
- Clarity: When we run `df -h` at the end of our e2e jobs, we want to see how much space the test used, particularly for troubleshooting `ENOSPC` issues.

Fixes: #640


bbrowning (approved):
Looks good - thanks!

PR #642, opened 2025-05-30
ci: report inode utilization

ktdreyer:
This is a total shot in the dark to get more data on https://github.com/instructlab/instructlab/issues/3298

Occasionally we hit `ENOSPC` errors from numpy. Theoretically inode exhaustion is one possible cause of `ENOSPC`. Report the number of inodes before and after the tests.

I recommend merging https://github.com/instructlab/sdg/pull/641 prior to this PR. When we do not run `e2e-ci.sh` with `-p`, that can throw off our `df` accounting.


bbrowning (approved):
I agree this is useful to help debug some of the e2e flakes. Thanks!

PR #643, opened 2025-05-31
Add configurable XFS retries to our e2e-medium job

bbrowning:
We've been seeing flaking ENOSPC errors in the e2e-medium jobs, but it's unlikely that we're actually running out of disk space or inodes. So, setup XFS retries for the root filesystem for ENOSPC errors in case we're running into some weirdness with the EBS root volumes.

### DIFF
--- a/.github/workflows/e2e-nvidia-l4-x1.yml
+++ b/.github/workflows/e2e-nvidia-l4-x1.yml
@@ -83,7 +83,20 @@ jobs:
         run: |
           cat /etc/os-release
           mkdir -p "${TMPDIR}"
-          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
+          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel util-linux
+
+      - name: Set XFS retries for root disk
+        run: |
+          # Locate our root block device - typically something like /dev/nvme2n1p2
+          ROOT_DEV="$(findmnt -nvo SOURCE --mountpoint /)"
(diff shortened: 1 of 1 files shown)


bbrowning:
I kicked this off on a test branch 3 times and the e2e tests passed each time. That doesn't mean this fixing the actual flake, but it's a good sign this at least doesn't make things work and it could potentially fix up the ENOSPC errors.

bbrowning:
The 3x passing job is at https://github.com/instructlab/sdg/actions/runs/15355142186 for reference.

ktdreyer (approved):

.github/workflows/e2e-nvidia-l4-x1.yml
```diff
@@ -83,7 +83,20 @@ jobs:
...
+        run: |
+          # Locate our root block device - typically something like /dev/nvme2n1p2
+          ROOT_DEV="$(findmnt -nvo SOURCE --mountpoint /)"
+          # Strip the lead "/dev/" off this device string
+          ROOT_DEV=${ROOT_DEV#"/dev/"}
+
+          MAX_RETRIES=5
+          RETRY_TIMEOUT_SECONDS=5
+          echo "Setting XFS retries for $ROOT_DEV to $MAX_RETRIES with a timeout of $RETRY_TIMEOUT_SECONDS"
+          echo "$MAX_RETRIES" > "/sys/fs/xfs/$ROOT_DEV/error/metadata/ENOSPC/max_retries"
```
ktdreyer, line 98: Interesting that the non-root user can write to these sys files.

PR #644, opened 2025-06-02
ci: Disable dependabot for pip dependencies

booxter:
Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,9 +13,3 @@ updates:
     directory: "/.github/workflows"
     schedule:
       interval: "daily"
-
-  # Maintain dependencies for Python scripts
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "daily"


bbrowning:
Will the weekly constraints updating job that will land in #638 take care of this responsibility from now on? I guess that plus a combination of us continuing to remove caps in the repo will accomplish the same thing, with regular PRs to bring in new dependency versions in a controlled way?

booxter:
Yes. The immediate trigger for this cleanup is that dependabot feels entitled to update constraints files like here: https://github.com/instructlab/eval/pull/271 and we want to control the process ourselves. The upper caps should generally go away so there's nothing for the dependabot to move. Lower caps should be moved forward when we know that an older version is broken, otherwise it's ok to leave it intact.

So tl;dr there's no job for dependabot in the pip world. But it will keep updating github workflow `use:` statements, as before.

bbrowning (approved):

bbrowning:
Approved with a hold label that we can remove once #638 merges.

PR #652, opened 2025-06-04
Filter out blank content when saving mixed dataset

derekhiggins:
Ensure we don't save out a mised dataset rows where content=None.

### DIFF
--- a/src/instructlab/sdg/datamixing.py
+++ b/src/instructlab/sdg/datamixing.py
@@ -231,6 +231,16 @@ class Recipe:
         as a jsonl file.
         """
         mixed_ds = self._create_mixed_dataset(num_proc)
+
+        # filter out any records where the any message content is None
+        mixed_ds = mixed_ds.filter(
+            lambda x: all(
+                message.get("content")
+                for message in x["messages"]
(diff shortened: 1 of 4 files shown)


derekhiggins:
@aakankshaduggal I've found the issue, null content is being introduced by __create_auxiliary_ds  here
https://github.com/instructlab/sdg/blob/55e224207/src/instructlab/sdg/datamixing.py#L471

The problem is that the record being used has 'document': None, this gets renamed to "response" and then used as the assistant content, so we end up with`{'role': 'assistant', 'content': None}`

derekhiggins:
> @derekhiggins Do we know why the document value is `None` here?

It looks like the pipeline that generated them was generating various summaries from a llm, one of the summaries was null and then later renamed to "document"

I've updated the PR to filter out instances where document=null , I think this alone would be enough to deal the the error in question. 

But for now I've left the other two filters in place so we can discuss to keep them or not

aakankshaduggal (approved):
Thanks @derekhiggins! LGTM!

bbrowning (approved):

derekhiggins:
@Mergifyio backport release-v0.8

PR #662, opened 2025-06-09
Default the pytest asyncio fixture loop scope

bbrowning:
Fixes `INTERNALERROR` spam from pytest when run under unitcov - things like:

```
INTERNALERROR> pytest.PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
INTERNALERROR> The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"
```


bbrowning:
Ignore the e2e-medium failure - it ran out of disk space and I haven't backported the xfs retries to help that yet.

booxter (approved):

booxter:
The fix was applied in other repos. Merging with 1 vote to speed the fix up.

PR #663, opened 2025-06-09
Run the e2e-medium job with aws region fallback

bbrowning:
We've had quite a few AWS scheduling issues with our e2e-medium tests lately. This ports the region fallback logic to those as well.


bbrowning:
Oh, the stop action did not work but I just force-pushed the fix for that 😁

bbrowning:
For more context, by canceling the e2e-medium job it actually tests the stop action. And, that's how I found and fixed the bug in the stop action.

